PR #39, opened 2014-09-18
[BUGFIX] Always set charset

andreasfernandez:
Currently, every driver needs to define their own SetCharSet method, because
the original method just returns FALSE. This patch sets the charset for every
driver, unless the method is overridden.


dregad:
@andreasfernandez was looking at this again, but noticed that your pull request results in no changes after your last merge. Please update it, or close it if this is no longer an issue.

tmotyl:
It was not fixed and it's still an issue. For TYPO3 core we have patched it internally. 
And about the merge, @dregad can please cherry pick these 2 commits:
https://github.com/ADOdb/ADOdb/commit/d1c216ead1f9c29ff66f3f7646228f07164f1f04
and 
https://github.com/ADOdb/ADOdb/commit/7c9d6db5203d0869841cd882076ddd905b759bf3
?

PR #62, opened 2014-11-25
PDO minor fixes

gohanman:
The inconsistent method parameters for MetaColumns and MetaTables just cause NOTICEs if error reporting is turned up high enough.

The gating around $this->_driver is probably to #56 but admittedly I don't know what the underlying problem is. I was simply trying to avoid the fatal error that occurs if the uninitialized method is treated as an object.

### DIFF
--- a/drivers/adodb-pdo.inc.php
+++ b/drivers/adodb-pdo.inc.php
@@ -438,12 +438,12 @@ class ADODB_pdo_base extends ADODB_pdo {
 		return $ret;
 	}
 
-	function MetaTables()
+	function MetaTables($ttype=false,$showSchema=false,$mask=false)
 	{
 		return false;
 	}
 
-	function MetaColumns()
+	function MetaColumns($table,$normalize=true)
 	{
 		return false;
 	}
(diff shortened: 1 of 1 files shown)


dregad:
> The inconsistent method parameters for MetaColumns and MetaTables just cause NOTICEs if error reporting is turned up high enough.

Fixed in 47d94059f44f0c52d19b22f533730c0d55ab638d.

> The gating around $this->_driver is probably to #56 but admittedly I don't know what the underlying problem is. I was simply trying to avoid the fatal error that occurs if the uninitialized method is treated as an object.

I'm not sure about this, can you provide some background or sample code for me to get a better understanding of the problem ?

gohanman:
I have to confess I don't remember the exact circumstance. It was a couple years ago that I was seeing some kind of problem with an uninitialized $this->_driver in the Begin/Commit/Rollback functions. I thought it was a fatal error but now that I'm trying to verify calling method_exists on a non-object is merely an E_NOTICE.

It probably only occurs through developer error, e.g.
```php
$ado = NewADOConnection('pdo');
$ado->Concat('one', 'two');
```

dregad:
> calling method_exists on a non-object is merely an E_NOTICE.

If not initialized, $_driver will be *null*. 

I just did a quick test, and calling `method_exists(null, 'blah');` just returns false with no errors or warnings (with error_reporting set to E_ALL), so I propose to close this without further change unless you can provide a code sample that reproduces the behavior..

gohanman:
Nope, closing it sounds fine to me. Sorry about the mix up.

PR #11, opened 2014-04-30
Avoid use of SERIAL in ALTER TABLE with PostgreSQL

asmecher:
Per email conversation with Damien -- see http://phplens.com/lens/lensforum/msgs.php?id=18150

PR #73, opened 2015-01-20
Handle partially SERIAL datatype.

jahto:
This is an alternative approach to pull request "Avoid use of SERIAL in ALTER TABLE with PostgreSQL #11". It has the benefit of not touching anything not PostgreSQL-specific and handles at least the adding of SERIAL columns. Not modifying them, unfortunately.


peterdd:
@dregad please add pgsql label to #73 and #11 

Haven't reviewed yet, but topic xmlschema+pgsql is important on my todo.

PR #83, opened 2015-02-27
URLs in documentation need updating

nirgal:
The following commit was originally submited at https://bugs.debian.org/471265 back in 2008.
Some URLs have changed again since, but this is better anyways.


dregad:
Fixed in 1a04b430e19d51f2aff03588233079c2ae615eb2

PR #86, opened 2015-03-12
Remove PGSQL legacy function calls, and switch to using proper ones where applicable.

ipso:
Minor fixes overall, mostly to better support HHVM v3.6. The biggest issue is pg_exec() is supposed to be for prepared statements, yet it was being called on mostly non-prepared statements. While this worked in stock PHP versions, it broke HHVM v3.6 and is "not correct(tm)" anyways, so its fixed now.

### DIFF
--- a/adodb-lib.inc.php
+++ b/adodb-lib.inc.php
@@ -16,7 +16,6 @@ $ADODB_INCLUDED_LIB = 1;
 */
 
 function adodb_strip_order_by($sql)
-{
 	$rez = preg_match('/(\sORDER\s+BY\s(?:[^)](?!LIMIT))*)/is', $sql, $arr);
 	if ($arr)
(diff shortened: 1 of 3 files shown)


dregad:
Many thanks for your contribution. 

I think I told you this before, but anyway here I go again. It would greatly facilitate the code review if you could *rebase* your changes on top of latest master branch, prior to submitting your pull request... I also strongly suggest you create one dedicated branch for each pull request instead of using your master branch.

Since you keep adding commits as they come, your pull request includes changes which are unrelated to the PR, as well as commits which were already merged in master.

Thanks for your understanding.

ipso:
I apologize, still relatively new to Github and GIT in general, I will close this PR and re-submit it as per your request.

PR #89, opened 2015-03-13
Prevent redeclaring of constants...

ipso:
This one I'm not 100% sure about, in our use case we would get fatal errors due to trying to redeclare constants, I think it has to do with the adodb-lib already being included. Anyways the below seemed to fix the issue... But what I don't quite understand is why only the below constants were causing problems and not other ones...

### DIFF
--- a/adodb.inc.php
+++ b/adodb.inc.php
@@ -86,17 +86,13 @@ if (!defined('_ADODB_LAYER')) {
 	// ********************************************************
 
 
-
-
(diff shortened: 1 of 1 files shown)


dregad:
I'd like to have more information about how you actually include and use the library to get constant redeclarations (sample code to reproduce the error would be best).

adodb.inc.php does not redefine constants if they have already been set (see `if ( !defined('_ADODB_LAYER')) {` at line 35, so as it is I don't think this PR is necessary.

Issue #41, opened 2014-09-22
mssqlnative / insert_ID() fails if server returns more than 1 row


gitjti:
okay, msssqlnative (5.19) driver's insert_id() mechanism has some issues if server returns
more than requested SELECT SCOPE_IDENTITY(). In my case, insert trigger outputs rows affected info.

Current code expects to get only one result -> error

	// retrieve the last insert ID (where applicable)
	sqlsrv_next_result($rez);
	sqlsrv_fetch($rez);
	$this->lastInsertID = sqlsrv_get_field($rez, 0);

dregad:
Can you please submit a pull request for this ?

gitjti:
Done.


-Jarkko

PR #102, opened 2015-05-18
Fix for bug #41 (mssqlnative / insert_ID() fails if server returns more than 1 row)

gitjti:
If inserting involves trigger action, SQL server probably  returns more
than result ”rows” and we are interested in last one only.


dregad:
Thanks for your contribution

PR #103, opened 2015-05-19
mssqlnative: sqlsrv_configure(warnings_return_as_errors should be War…

MorrisR2:
In /Volumes/devmoodle/lib/adodb/drivers/adodb-mssqlnative.inc.php we find these three lines:

```
30:    sqlsrv_configure("WarningsReturnAsErrors", $constant);
142:  sqlsrv_configure('warnings_return_as_errors', 0);
147:  sqlsrv_configure('warnings_return_as_errors', 0);
```
The last two throw "error code = -14, sqlsrv_configure: message = An invalid parameter was passed to sqlsrv_configure".  The documentation confirms WarningsReturnAsErrors is correct,  warnings_return_as_errors is not.
See http://php.net/manual/en/function.sqlsrv-configure.php


dregad:
Many thanks for your contribution.

PR #104, opened 2015-05-22
Added data dictionary support for PDO.

CreatrixTeam:
Added data dictionary support for PDO which should allow ADODB extensions using the data dictionary to run successfully when using PDO. However the support is unlikely to run on the old versions of PHP that ADODB was originally designed before.

### DIFF
--- a/adodb-active-record.inc.php
+++ b/adodb-active-record.inc.php
@@ -16,6 +16,12 @@
   	for info on Ruby on Rails Active Record implementation
 */
 
+/**
+* Maximum length allowed for object prefix. Matches the setting in adodb-xmlschema03.inc.php
+*/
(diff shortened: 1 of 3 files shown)


CreatrixTeam:
We are new to the Github process. We were wishing to commit new changes that affect the file "adodb-active-record.inc.php" which will add experimental table pre fixing support, but we were hoping to do so after this request gets through, but so far no response. If possible, please reply as to whether this request is under review, shall be under review, or rejected.

dregad:
Sorry guys, I do this in my spare time and I don't have much of that at the moment. I'll review your pull request when I can. 

Your contributions are of course welcome and appreciated, feel free to send more fixes and enhancements.

CreatrixTeam:
Thank you for your reply. That was all we needed to know.

PR #107, opened 2015-05-26
Query pagination optimization

ipso:
Optimize query pagination when row counting is disabled. Rather than issue 3+ separate queries, it simply uses just one in most cases, resulting in huge performance increases especially on large tables. 

Please review line #577 where we reduce the number of rows in the returned recordset (rather than re-ssue the query), is this the safe/proper way to do that?

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -5,4 +5,5 @@
 *.tmp
 .project
 .settings/
+/nbproject/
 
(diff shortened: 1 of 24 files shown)


dregad:
May I kindly ask you to rebase your feature branch on top of latest master, so that the pull request only contains relevant commits

PR #108, opened 2015-05-26
Load balancer/query read/write splitting...

ipso:
First version of this class that allows the user to do read/write splitting and load balancing across multiple connections. It can handle and load balance any number of master or slaves (multi-master or master/slave), including dealing with connection failures with minimal changes to their existing application and ADODB itself.

Example usage is in the file itself.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -5,4 +5,5 @@
 *.tmp
 .project
 .settings/
+/nbproject/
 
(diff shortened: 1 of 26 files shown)


dregad:
May I kindly ask you to rebase your feature branch on top of latest master, so that the pull request only contains relevant commits

Issue #105, opened 2015-05-24
ExecuteCursor produce error: oci_execute() expects parameter 1 to be resource, string given


apit:
```php
$rs = $db->ExecuteCursor('begin sp_validasi_todo(:fua, :query, :c); end;',  'c', 
						array('fua'=>$fua, 'query'=>$superviseeSql)); 
```

it seems `adodb_strip_count_keyword()` in adodb.inc.php#1162 (dev-master) is turning `resource(77, oci8 statement)` (vardumped) in `$sql[1]` and `$sql[4]` into string of 'Resource id `#resourcenumber'`. by commenting `adodb_strip_count_keyword()` line, i am able to get correct result for sp. Havent checkout for other queries tho.

Edit: 
PHP 5.3.29
OCI8 Version => 2.0.8
Oracle Run-time Client Library Version => 11.2.0.4.0
Latest adodb as of now via composer

ipso:
Pretty sure I'm responsible for that bug... I don't quite understand whats going on though, is the Oracle driver sending _Execute() a resource rather than a string SQL query? 

I don't have access to Oracle to test this, but what if you change the following code in adodb.inc.php:

        //Strips keyword used to help generate SELECT COUNT(*) queries from SQL if it exists.
        function adodb_strip_count_keyword( $sql ) {
                return ADODB_str_replace( '_ADODB_COUNT', '', $sql );
        }

To:

        //Strips keyword used to help generate SELECT COUNT(*) queries from SQL if it exists.
        function adodb_strip_count_keyword( $sql ) {
                if ( is_string( $sql ) == TRUE ) {
                          return ADODB_str_replace( '_ADODB_COUNT', '', $sql );
                }

                return $sql;
        }

Does that fix that issue for you?

apit:
in my case (ExecuteCursor), $sql is an array:
```php
array:5 [▼
  0 => "begin sp_validasi_todo(:fua, :query, :c); end;"
  1 => :oci8 statement {@91}
  2 => 0
  3 => 1
  4 => & :oci8 statement {@92}
]
```
so, yeah, thanks :).

dregad:
Issue #88 introduced this regression, and I don't think this should be closed without patching the code. 

@ipso if you could please review your code and make sure it also work in the case submitted by @apit, and propose a pull request to fix the problem. Thanks.

PR #109, opened 2015-05-27
Fix Issue #105 by only attempting str_replace() on strings rather tha…

ipso:
…n arrays.


dregad:
Fixed in 52d1725e95ee30f28d4c6a9b3ec72c8e4d4e025f

PR #111, opened 2015-05-27
Load balancer/query read/write splitting

Mike-Benoit:
First version of this class that allows the user to do read/write splitting and load balancing across multiple connections. It can handle and load balance any number of master or slaves (multi-master or master/slave), including dealing with connection failures with minimal changes to their existing application and ADODB itself. 

Of course once a transaction starts all queries "stick" to that connection until it ends, at which point load balancing/query splitting resumes again.

Example usage is in the file itself.


ipso:
Any plans on adding this? Its completely self contained and doesnt modify any other code.

dregad:
Plans, yes. Timeline, no.
I'm very busy atm, and don't have much time to spend on ADOdb unfortunately.

ipso:
Just wanted to put a friendly bump on this as it's approaching a year long wait... It's in been production for that long serving millions of users and has been solid from that standpoint.

Thanks.

mnewnham:
Are you able to provide documentation for this?
  - An overview with explanation of how it works, uses, features etc.
 - Reference for public functions  done in a manner similar to our standards, with examples.

If you do this in plain text, I am happy yo massage them into our dokuwiki format. In addition , as I have not used this type of functionality before, I can use this to build a test platform.

ipso:
Sure, I will see what I can put together.

 It is designed to be as simple as possible and be a drop in system to distribute your queries to master/slave servers. As you can see from the example usage at the top of the file, 3 lines gets you up and running, no other changes needed anywhere else in the codebase.

mnewnham:
I briefly looked at this, simply to ensure that a connection could be established under various databases (mysqli,sql server, oracle)

please note that the example usage line:
``$db_connection_obj = new ADOdbLoadBalancerConnection( 'master', 10, $dsn )``
should say
``$db_connection_obj = new ADOdbLoadBalancerConnection('driver','master', 10,$persistent, $dsn )``

mnewnham:
@ipso , have you had any time to draft some documentation for this?

ipso:
Unfortunately I haven't had time yet, it's still on my TODO list though and will get to it soon hopefully.

ipso:
Below is a first crack at some documentation:

ADODB Load Balancer is a load balancing or read/write query splitting middle man that assists with Master/Slave replication setups. Typically all writes or any queries within a transaction would be sent to the master database connection, and all reads would be load balanced between master and slave connections. Additionally if a connection to one database is unable to be established, it will automatically try a different connection for basic fail over functionality.

This class is designed to require the minimum necessary changes to the existing application code base, usually all that is required is adding a few lines for each connection/database. If we take an existing example code snippet of:

```
include('adodb.inc.php');    # load code common to ADOdb
$conn = &ADONewConnection('access');        # create a connection
$conn->PConnect('northwind');   # connect to MS-Access, northwind DSN
$recordSet = &$conn->Execute('select * from products');
```

It could be converted to use load balancing with the following:
```
include('adodb.inc.php');    # load code common to ADOdb
include('adodb-loadbalancer.inc.php'); 

$conn = new ADOdbLoadBalancer();

$db_connection_obj = new ADOdbLoadBalancerConnection( 'access', 'master', 10, FALSE, $dsn ); //Master with weight of 10
$conn->addConnection( $db_connection_obj );

$db_connection_obj = new ADOdbLoadBalancerConnection( 'access', 'slave', 100, FALSE, $dsn ); //Slave with weight of 100
$conn->addConnection( $db_connection_obj );

$recordSet = &$conn->Execute('select * from products'); //Run queries as normal against $conn
```


Further example Usage:
```
  $db = new ADOdbLoadBalancer();
  $db_connection_obj = new ADOdbLoadBalancerConnection( 'postgres8', 'master', 10, FALSE, $dsn ); //Master with weight of 10
  $db_connection_obj->getADODbObject()->SetFetchMode(ADODB_FETCH_ASSOC); //Pass specific settings to the ADOdb object itself.
  $db->addConnection( $db_connection_obj );

  $db_connection_obj = new ADOdbLoadBalancerConnection( 'postgres8', 'slave', 100, FALSE, $dsn ); //Slave with weight of 100
  $db_connection_obj->getADODbObject()->SetFetchMode(ADODB_FETCH_ASSOC); //Pass specific settings to the ADOdb object itself.
  $db->addConnection( $db_connection_obj );

  $db_connection_obj = new ADOdbLoadBalancerConnection( 'postgres8', 'slave', 100, FALSE, $dsn ); //Slave with weight of 100
  $db_connection_obj->getADODbObject()->SetFetchMode(ADODB_FETCH_ASSOC); //Pass specific settings to the ADOdb object itself.
  $db->addConnection( $db_connection_obj );

  //If necessary, define one or more SQL queries that are run once per connection, only when it is established.
  $db->setSessionInitSQL( 'SET TRANSACTION ISOLATION LEVEL READ COMMITTED' );
 
  //If necessary, define session variables that are executed immediately and upon each connection when its first established.
  $db->setSessionVariable( 'TIME ZONE', 'PST8PDT' );

  //Perform ADODB calls as normal..
  $db->Execute( 'SELECT * FROM MYTABLE1' ); //Likely to go to a slave.
  $db->Execute( 'SELECT * FROM MYTABLE2' ); //Likely to go to a slave.
  $db->Execute( 'INSERT INTO MYTABLE VALUES(...)' ); //Will go to the master.
```


In the above example, three separate connections are registered, though no actual connections will be made until a query is sent to that connection, one master and two slaves. The weight for the slaves is significantly higher than the master, so the read queries will be evenly split between the slaves and only about 10 in 210 read queries will go to the master, but all write queries or transactions will go to the master. Since the master database handles all the writes, typically it would be a significantly lower weight than the slave connections to offset the load.


ADOdbLoadBalancerConnection( $driver, $type = 'master', $weight = 1, $persistent_connection = FALSE, $argHostname = '', $argUsername = '', $argPassword = '', $argDatabaseName = '' ) - Defines the connection that will be load balanced when queries are run. The type can be 'master' or 'slave', and the higher the weight the more queries it will receive when load balancing.

setSessionInitSQL( $query ) - This can be called multiple times to add multiple queries that are run when the connection is first established. Primarily for things like setting transaction isolation modes or any other query that should be run at the start of a session.

setSessionVariable( $name, $value, $execute_immediately = TRUE ) - Defines 'SET SESSION' variables by name/value. Example usage may be: setSessionVariable( 'TIME ZONE', 'PST8PDT' ), this is executed on all connections immediately, or only when new connections are established.

mnewnham:
I'd like to offer some suggestions first on the 'session' functionality:

1. Using the terminology 'setSession' feels very ambiguous in this use environment. It's highly likely that users will have browser-based session management either through PHP or quite possibly through the ADOdb session manager, after all, it is exactly this type of scenario (cluster) that the ADOdb session manager is designed for. I do understand that this is a class method, but maybe you would consider renaming it to ``setLbSession`` or some-such.
2. The ``SET SESSION`` syntax seems to be supported by very few databases, so I'm not sure, from a portability perspective, the value of having a method for this specific syntax. Perhaps I don't quite understand the importance of this in the set up and running of the application. Maybe you could expand out your usage of this functionality to make it clearer.  
3. Is there a likelyhood that we would want to vary the initsql statements based on whether we would be running a master/slave configuration? For example, we might set a default transaction isolation method differently for a read-only slave.

**Possible feature enhancement**
Is there a way that I can force a query to run on a specific machine? Say that I have a very memory-intensive procedure, and I want to force it to run on the lowest priority slave. Is there a way that I can do this?

ipso:
Thanks for reviewing this in such detail, I will respond below:

> Using the terminology 'setSession' feels very ambiguous in this use environment. It's highly likely that users will have browser-based session management either through PHP or quite possibly through the ADOdb session manager, after all, it is exactly this type of scenario (cluster) that the ADOdb session manager is designed for. I do understand that this is a class method, but maybe you would consider renaming it to setLbSession or some-such.

I used setSession so it would match the database "SET SESSION" terminology. I personally prefer to stay away from abbreviations as much as possible in function names, since setLbSession() may not be obvious what "LB" is, but I'm happy to rename this if you require.

> The SET SESSION syntax seems to be supported by very few databases, so I'm not sure, from a portability perspective, the value of having a method for this specific syntax. Perhaps I don't quite understand the importance of this in the set up and running of the application. Maybe you could expand out your usage of this functionality to make it clearer.

In our case, its critical to be able to set database session variables. For MySQL its like the example for setting SQL_MODE and TRANSACTION ISOLATION, as well as for changing TIMEZONE. This is of course optional, but important functionality for many applications. 

I do realize that not all database use SET SESSION, and even the ones that do, some require a "=" in the syntax and some don't (ie: MySQL vs. PGSQL), however I didn't want to bloat the rest of ADODB with functionality to handle this in a more cross-database manner as I'm not even sure its possible due to the wide range of syntax and different names. Even for timezones, it might be "set session timezone = 'PST8PDT'" or "set session time zone 'PST8PDT'".

As for making it clearer, are you referring to the code, or documentation, or both?

> Is there a likelyhood that we would want to vary the initsql statements based on whether we would be running a master/slave configuration? For example, we might set a default transaction isolation method differently for a read-only slave.

That is definitely possible I suppose, I just can't think of a real-world use-case for that at the moment. I would prefer to keep things as simple as I possibly can at least initially, unless there is a known specific use-case that we must handle. Can you think of one?

>Possible feature enhancement
Is there a way that I can force a query to run on a specific machine? Say that I have a very memory-intensive procedure, and I want to force it to run on the lowest priority slave. Is there a way that I can do this?

There is not currently a easy way to do that, but you could just iterate through the connections and pick the one you wish to send a query to directly if you want. Having such functionality would be further complicated with the failover code if a connection fails and such too of course.

> isReadOnlyQuery method
Perhaps you could change this to use a case statement based on the dataProvider? I noticed that it can't, for example recognize the SQL Server UPDLOCK keyword, and I'm sure that the must be others out there, and that IF statement could get quite long.

I think it may be a little more complicated than replacing it with a SWITCH statement, since its mostly checking to make sure its a SELECT query first, then ensuring that the keywords are NOT contained in the query using AND. I'm drawing a blank as to how a SWITCH statement would improve that logic, do you happen to have any examples?

Similar to the _ADODB_COUNT keyword feature I submitted a year ago, we will never be able to handle every case where a query tries to write to the database, especially when you take into account stored procedures, so I definitely could see a need to add another keyword like _ADODB_LB_WRITE, or some other mechanism to tell it its a write query. Until we get it out in the wild and start to get some user feedback, its difficult to know what will work best though.

mnewnham:
What I'd like to work with you on here is making sure that before we make the product part of the suite,  we can ensure portability across  all the supported databases, and in particular, from the beginning, the important commercial offerings - oracle, mssql and db2

I accept your position of simplifying functionality, but from our perspective, we have to balance that against portability, and ensuring that we can achieve that from the start. So right now we have a couple of issues we need to resolve:   

That the ``SET SESSION`` functionality is limited effectively to the mysql driver, and that we must be able to cogently explain to users of other databases what alternative methods are available to achieve the same effect.

That we do need to be able to easily add in support for other drivers in the isReadOnlyQuery method. The reason I suggested a case switched based on the driver is that I think that the single if statement could easily dissolve into a mish-mash of 

````php
(else if $db->databasetype == 'x' && stripos ) || $db->databasetype == 'y' && !blah blah) || (!$db->databaseType == 'z'
````
type statements, making code maintenance more difficult. 

**More Feature Thoughts**
I'm wondering how well this product plays with 2 other parts of the ADOdb suite

1.  Query caching, in particular the Memcached option
2. The ADOdb session management system. 

I see these 2 items as extremely complementary to you feature.

**Comment by ipso** [comment]
I must confess that I have never used any database other than MySQL and PostgreSQL. This code has been tested and running on both for quite some time now. 


> That the SET SESSION functionality is limited effectively to the mysql driver, and that we must be able to cogently explain to users of other databases what alternative methods are available to achieve the same effect.

I'm not sure I follow, but its not specific to any one database type at all, its actually extremely generic, its just the example that is specific to MySQL. (We actually use/prefer PGSQL)

Essentially all the setSessionVariable function does it execute the query immediately *as well as* stores it in memory to be executed when a new connection to a database is established to keep session state consistent. The load balancer code already has setSessionInitSQL() which handles running queries when switching connections, but its missing the "execute now" part that is critical to things like switching time zones. Maybe we just add a flag to that function to execute the query immediately or only on new connections, then setSessionVariable() could go away completely. 

Do you think that would suffice?


>That we do need to be able to easily add in support for other drivers in the isReadOnlyQuery method. The reason I suggested a case switched based on the driver is that I think that the single if statement could easily dissolve into a mish-mash of
>(else if $db->databasetype == 'x' && stripos ) || $db->databasetype == 'y' && !blah blah) || (!$db->databaseType == 'z'
type statements, making code maintenance more difficult.

I think we need to add a isReadOnlyQuery to each driver then, then simply have the load balancer call back to that. The existing code would probably work as a "default", but then we can have driver specific code in each driver class itself. I wanted to try and keep the original code as self contained as possible to make merging it easier, but I'm fine with offloading more work into each driver. Outside of MySQL/PGSQL I may not be much help though.

> More Feature Thoughts
I'm wondering how well this product plays with 2 other parts of the ADOdb suite
>Query caching, in particular the Memcached option
The ADOdb session management system.
I see these 2 items as extremely complementary to you feature.

In theory it should all work, if I recall the query caching was tested somewhat, but we have never used the ADODB session management.

**Comment by mnewnham** [comment]
I don't think there's really a need to provide driver specific versions (unless you want to), It looks like the isReadOnlyQuery is the point at which we need to analyze the connection type.

**Working with cached record sets**
I think that we need to throw an error if they are using a cache function and not using a memcached server. The cache functions work on a first call + timeout expiration, so it's conceivable that different servers might have different cached result sets.

**Function SetBlackListFunctions**
Does this need to be finished? If you have decided against its usefulness, perhaps you could strip the code out.

**And Finally**
And I know that you're not going to like it, but we really need to get some function documentation into this code, as well as consistent function visibility. We've described how we want to move forward with coding standards on [this page](http://adodb.org/dokuwiki/doku.php?id=project:standards), and I see no reason why, given this is all new code, we cannot get close to conformance from the start. 

As I hope you can tell, we've invested a lot of time over the last year or two, tearing open code, identifying undocumented feature and trying to make sense of it. There's still a lot to do in that project, and I don't want to add this one to the pile.

I definitely see this feature as part of a suite, _**ADOdb Connection Managent**_ or some other contrived name, along with session management and caching, and if we can start with this, pulling the other 2 parts along (both of which are outdated) I think we will see the reward. (not financial obviously)

**Comment by ipso** [comment]
>I don't think there's really a need to provide driver specific versions (unless you want to), It looks like the isReadOnlyQuery is the point at which we need to analyze the connection type.

Well its how ADODB handles all other DB specific operations anyways, so may as well stay consistent? This is also a critical function performance wise, so we should make every effort to streamline it as much as possible, and making a large SWITCH may do the opposite. 

Should we put the "main" isReadOnly function in adodb-lib.inc.php or adodb.inc.php?

> Working with cached record sets
I think that we need to throw an error if they are using a cache function and not using a memcached server. The cache functions work on a first call + timeout expiration, so it's conceivable that different servers might have different cached result sets.

I disagree, there may be cases where caching is useful but memcached is not needed. We currently use ADODB's caching to just cache the table schema pretty much, and for data that rarely changes, it works great, even with the load balancer. 

>Function SetBlackListFunctions
Does this need to be finished? If you have decided against its usefulness, perhaps you could strip the code out.

I have removed it now.

>And Finally
And I know that you're not going to like it, but we really need to get some function documentation into this code, as well as consistent function visibility. We've described how we want to move forward with coding standards on this page, and I see no reason why, given this is all new code, we cannot get close to conformance from the start.
> As I hope you can tell, we've invested a lot of time over the last year or two, tearing open code, identifying undocumented feature and trying to make sense of it. There's still a lot to do in that project, and I don't want to add this one to the pile.
> I definitely see this feature as part of a suite, ADOdb Connection Management or some other contrived name, along with session management and caching, and if we can start with this, pulling the other 2 parts along (both of which are outdated) I think we will see the reward. (not financial obviously)

Not a problem, I have mostly documented all the code and formatted it to PSR-2 now. See most recent commit.




Additionally, after getting back into the code a little more, I realized setSessionVariables() is more complicated than I remembered. It will need to stay separate from setSessionInitSQL() because it needs to account for cases where session variables are changed multiple times.

For example if during a single connection, the timezone is changed 10x, then the queries get load balanced to a different connection, it needs to copy the session state to the new connection, so it just needs to know about the last timezone that was set, not the 10 before it. That is why it has to deal with the session variable name and value separately. 

Keep in mind the function is still generic to handle all databases and session variables, the only thing is that the user must define the actual "value", ie:

MySQL uses: SET SESSION foo_bar='foo'
So the function call would be: setSessionVariable( 'foo_bar', '=foo' );

PGSQL uses: SET SESSION foo_bar 'foo'
So the function call would be: setSessionVariable( 'foo_bar', 'foo' );

The session variable name 'foo_bar' is still the same in both cases, so the LB code can properly maintain its state across connections.

We could stick a session variable operator in each driver, (MySQL would be '=' and PGSQL would be ' ') that might make this easier to use across more databases, but the reality is that session variable names will also be different too, so I'm not sure how much we will gain from that. 

I just need some more direction on how you would like to proceed, as this seems like the biggest sticking point.

**Comment by mnewnham** [comment]
If we are going to keep the local caching in effect, we need to make sure that anyone that implements it in their setup understands that there may be an issue.  Whilst your configuration might work fine, (I'm guessing you have a long cache timeout), other setups with short caching times might be problematic. Perhaps we can add a warning that can be suppressed by defining a constant, e.g.

````php
if (!DEFINED('ADODB_SUPPRESS_CACHE_WARNING'))
    print 'Please read the warning about conflicts between caching and the load balancer on the documentation page';
````

If local caching is enabled and the load balancer is running

**Comment by ipso** [comment]
I believe you are trying to solve a problem in code that should simply be mentioned in documentation. 

I urge you to reconsider putting such a warning in ADODB, if our application was pushed to production servers, and a database administrator decided to ratchet the caching timeout down to the threshold that suddenly triggers that warning, it would completely break the entire application or if it wasn't a "print" line but instead some kind of trigger_error, it would fill the logs with millions of garbage messages, possibly causing other problems like running out of disk space or causing a high load situation.

A low cache timeout is no different than a high cache timeout, I don't see why some magic threshold we arbitrarily determine makes that any different and its creating an impossible situation that we can never get "right".

I think your concern regarding caching is overblown anyways, by definition cached data is "out-of-date", and when using master/slave replicated database, the slave database is rarely more than a few milliseconds out of date, whereas cached data in ADODB will often be many minutes, hours or days out of date. In fact with some PGSQL setups, the slave can be in synchronous mode, so the master won't return from a transaction until the slave as received the data and acknowledged it (this is for disaster recovery purposes among other things), so there literally is no discernible "delay" between the two.

I think a simple message in the documentation concerning caching and how the developer must consider caching more carefully when using a replicated database system will suffice.

**Comment by mnewnham** [comment]
**A few more points**
I've been testing this now against a mysql replication setup with a master and 2 slaves to emulate a cluster. 

1. Are you actually using a cluster? Do you think that this product actually gives a benefit over using the normal cluster management tools ?
2. Do you think it would be better to route the metaFunctions through a slave rather than run them over the master.
3. Right now, if you define multiple connections with the same priority, all connections except the last created are ignored. Would it be difficult to create a connection randomiser? 

Other than that, I've found no issues

**Comment by ipso** [comment]
> Are you actually using a cluster? Do you think that this product actually gives a benefit over using the normal cluster management tools ?

We use it primarily on a master/slave setup with just two servers. We have benchmarked it against middle layer tools for query splitting and its significantly faster since the load is handled by the many web servers instead of consolidated at fewer middle layer servers or even the DB servers themselves. It doesn't have all the advantages of some of these tools though of course, but it should be much less complicated to setup.

> Do you think it would be better to route the metaFunctions through a slave rather than run them over the master.

I will need to look at the code more closely, but as long as we are 100% certain they run read only queries they should be load balanced through all connections. Though for the initial release I did try to stay on he conservative side when handling the routing.

> Right now, if you define multiple connections with the same priority, all connections except the last created are ignored. Would it be difficult to create a connection randomiser?

That shouldn't be the case, I will need to replicate this and get it fixed.

**Comment by mnewnham** [comment]
I'd also like to raise whether you would consider changing the terminology from master/slave to something like primary/secondary

**Comment by mnewnham** [comment]
I wondered if you had had the time to look at/resolve the last issue I raised?

**Comment by ipso** [comment]
Master/Slave is pretty widely used and understood terminology, switching to primary/secondary would seem to confuse things, as the class supports unlimited nodes, so what do we call node 3,4,5,6,7,...? Master/Slave seems to fit this well, as slaves are read-only and masters are read/write, and any number of each are supported if necessary.

Do you still want this changed?

**Comment by mnewnham** [comment]
I totally accept the traditional acceptable usage is appropriate, I am wondering about [this](https://en.wikipedia.org/wiki/Master/slave_(technology)#Appropriateness_of_usage)

**Comment by ipso** [comment]
I don't see the point in wasting cycles pandering to political correctness especially when it diverges from the commonly known and understood terminology. Both MySql and Postgresql documentation refers to master/slave, and ADODB isn't exactly a end-user facing product that any marketing department needs to care about.

I'm happy to change the naming if something more obvious can be suggested, but primary/secondary and 'replica' don't seem like improvements to me.

**Comment by mnewnham** [comment]
Well that escalated quickly, didn't it. It's easy to bang on about 'political correctness' when you don't have to sit in a room full of US government entity Systems Admins, IT managers and directors from the southern US states and say the word 'slave' 20 or 30 times. I have had to, and I felt like crawling under a desk each time I said it.

From a marketing perspective, you've asked us to add your programs to the core ADOdb distribution. You didn't have to, and you could have just put it out for download, but I suspect it would just languish there without any traffic.  I don't have time to grab the source and change it, I barely have time to look after the core software, but I have invested a good amount of time in testing your software. We're nearly there, so I'm going to add a simple precondition to going further with it and getting it into the main distribution: To satisfy my PC sensibilities, I'd like you to change the terminology from master/slave to something else. I don't care what you change it to, but the change must be made.

**Comment by ipso** [comment]
I understand what you are saying, but the mysql and pgsql manuals at least refer to master/slave, so I don't see how using some other terminology changes that, and it just adds further confusion to everything, which is an ongoing cost, not just a one-time cost. 

I'm not 100% sure what you are actually objecting too though, are wanting to remove any mention of the word "slave" from the source code? Or are you wanting a developer to just not have to use the word "slave" when using this class in their own code? Do you object to "master" too? Are you okay with allowing aliases, like master or parent, and slave or child?

I'm fine with playing by your rules to contribute to the project, that is expected, but not having them clearly laid out just turns me into your slave as I try to jump through hidden hoops until one of us eventually tires.

**Comment by mnewnham** [comment]
Let's just make sure that all the exposed functions and variables have appropriate names. You can choose whatever you wish. Once that it is done I can complete the documentation and get it merged.

**Comment by ipso** [comment]
Okay, 'slave' has been removed from the code other than a single comment at the top to simply state that 'readonly' may be more commonly known as a 'slave' database.

**Comment by dregad** [comment]
@mnewnham following your recent message, I'm targeting this at 5.21.

I noticed a couple [related pages in the wiki](http://adodb.org/dokuwiki/doku.php?id=v5:loadbalancer:index), but they look like a draft (missing links, incomplete tutorial, etc). It might also be worth mentioning in the doc that the feature is only available as of 5.21.

Let me know if you want to finalize the doc, or if there are any changes needed in the code, before I merge the PR.

**Comment by ipso** [comment]
Friendly bump, any chance of getting this merged? 

Its been pending for two years now and doesn't actually touch or affect any existing code, so breakage isn't a concern... Perhaps we could label it a "beta" feature in the initial release so we can leave the door open for changes later on if necessary?

**Comment by dregad** [comment]
@ipso sorry about the delay. Busy life and all that, you know how it is I'm sure... I personally have no objections with merging this new feature (although I did not actually test it).

@mnewnham please answer my question in https://github.com/ADOdb/ADOdb/pull/111#issuecomment-252871716

**Comment by mnewnham** [comment]
@ipso, A while ago in testing I raised this issue:

if you define multiple connections with the same priority, all connections except the last created are ignored. Would it be difficult to create a connection randomiser? You said would look at it. Has there been a resolution?

@dregad I'll try to start looking at the docs after i've finished with memcached

**Comment by mnewnham** [comment]
@dregad , could you create a  'load balancer' tag in the issues please? @ipso , as I go through and document, I'm going to be testing all the available options and posting problems in the issues, You'll need to keep an eye on the list

**Comment by dregad** [comment]
> could you create a 'load balancer' tag in the issue

Done https://github.com/ADOdb/ADOdb/labels/load-balancer

**Comment by ipso** [comment]
@mnewnham Do you have example code that replicates the multiple connection issue you described earlier?

**Comment by ipso** [comment]
@mnewnham I looked into the whole connection randomizer thing you mentioned earlier, and what you may be seeing is the "sticky session" feature. 

By default "sticky sessions" are enabled, so once a write or readonly connection is established, it won't bother creating new ones, it will send all queries to one of those two connections. If multiple readonly connections are defined, it randomly (based on weight) picks *one* connection and uses it for all subsequent readonly queries within that request. 

So the load balancing is handled on a "per request" basis by default, rather than a "per query" basis. This has many advantages, such as avoiding the overhead of connecting to many databases within a single request, especially if the requests are short lived, which they mostly are in the web-server world of course. It doesn't make sense to spend 50ms connecting to 10 different databases to run 10 queries (1 on each), when the 10 queries can be run a single database in 10ms. Not to mention all the issues with cache coherency too. 

Now, if you set "$load_balancer_obj->enable_sticky_sessions = false", that changes the load balancing to "per query" mode, and it will try to load balance each individual query between all available connections, even within the same request. 

I did just realize that "$enable_sticky_sessions" was a protected rather than public variable, so I have just fixed that and committed it.

**Comment by ipso** [comment]
Below is some sample output from my tests based on the last comment.

With sticky sessions enabled:
```
$ php test.php
Using Connection ID: 3
string(6) "query1"
string(6) "query2"
string(6) "query3"
string(6) "query4"
string(6) "query5"
string(6) "query6"
string(6) "query7"
string(6) "query8"
string(6) "query9"
string(7) "query10"
$ php test.php
Using Connection ID: 5
string(6) "query1"
string(6) "query2"
string(6) "query3"
string(6) "query4"
string(6) "query5"
string(6) "query6"
string(6) "query7"
string(6) "query8"
string(6) "query9"
string(7) "query10"
$ php test.php
Using Connection ID: 4
string(6) "query1"
string(6) "query2"
string(6) "query3"
string(6) "query4"
string(6) "query5"
string(6) "query6"
string(6) "query7"
string(6) "query8"
string(6) "query9"
string(7) "query10"
$ php test.php
Using Connection ID: 2
string(6) "query1"
string(6) "query2"
string(6) "query3"
string(6) "query4"
string(6) "query5"
string(6) "query6"
string(6) "query7"
string(6) "query8"
string(6) "query9"
string(7) "query10"
```
With sticky sessions disabled:
```
$ php test.php
Using Connection ID: 4
string(6) "query1"
Using Connection ID: 0
string(6) "query2"
Using Connection ID: 5
string(6) "query3"
Using Connection ID: 3
string(6) "query4"
Using Connection ID: 0
string(6) "query5"
Using Connection ID: 0
string(6) "query6"
Using Connection ID: 5
string(6) "query7"
Using Connection ID: 0
string(6) "query8"
Using Connection ID: 2
string(6) "query9"
Using Connection ID: 4
string(7) "query10"
mikeb@dev1:~/public_html/ADOdb$ php test.php
Using Connection ID: 5
string(6) "query1"
Using Connection ID: 0
string(6) "query2"
Using Connection ID: 1
string(6) "query3"
Using Connection ID: 2
string(6) "query4"
Using Connection ID: 1
string(6) "query5"
Using Connection ID: 1
string(6) "query6"
Using Connection ID: 0
string(6) "query7"
Using Connection ID: 2
string(6) "query8"
Using Connection ID: 5
string(6) "query9"
Using Connection ID: 1
string(7) "query10"
```

Here is the code I used:
```
<?php
require_once( 'adodb.inc.php');
require_once( 'adodb-exceptions.inc.php');
require_once( 'adodb-loadbalancer.inc.php');

//$driver = 'mysqli';
$driver = 'postgres8';
$db = new ADOdbLoadBalancer( $driver );
$database = 'adodb';
$user     = 'adodb';
$password = 'adodb';
$host     = '127.0.0.1';
$db_connection_obj = new ADOdbLoadBalancerConnection( $driver, 'write', 10, false, $host, $user, $password, $database ); //Master with weight of 10
$db->addConnection( $db_connection_obj );

$db_connection_obj = new ADOdbLoadBalancerConnection( $driver, 'readonly', 10, false, $host, $user, $password, $database ); //Master with weight of 10
$db->addConnection( $db_connection_obj );

$db_connection_obj = new ADOdbLoadBalancerConnection( $driver, 'readonly', 10, false, $host, $user, $password, $database ); //Master with weight of 10
$db->addConnection( $db_connection_obj );

$db_connection_obj = new ADOdbLoadBalancerConnection( $driver, 'readonly', 10, false, $host, $user, $password, $database ); //Master with weight of 10
$db->addConnection( $db_connection_obj );

$db_connection_obj = new ADOdbLoadBalancerConnection( $driver, 'readonly', 10, false, $host, $user, $password, $database ); //Master with weight of 10
$db->addConnection( $db_connection_obj );

$db_connection_obj = new ADOdbLoadBalancerConnection( $driver, 'readonly', 10, false, $host, $user, $password, $database ); //Master with weight of 10
$db->addConnection( $db_connection_obj );

$db->enable_sticky_sessions = false;
//$db->debug = true;

for( $i = 1; $i <= 10; $i ) {
	var_dump( $db->GetOne( 'select \'query'. $i .'\'' ) );
	$i++;
}
//$db->debug = true;
?>
```

dregad:
@mnewnham this is tagged for 5.21.0 - do you think it can be merged now, or should it be postponed to a later release ?

mnewnham:
@dregad , As this is a standalone product, There's no reason why it couldn't go in now, as long as we can get the rebasing fixed

dregad:
@Mike-Benoit I (finally) got around to merging this... Sorry it took so long and many thanks for your contribution.

I made some changes to the code (mostly cosmetic / whitespace), but also a few issues that were identified by static analysis. See the details in the commits I just merged (e1a80b5356fc3145c6f200db997c73d4f816bc2b..13b946c30ab6d19b611e8fcd4ed4547f6549a549).

There's a few left, that I was not 100% sure how to fix:

- line 318: Variable $connection_id might have not been defined 
- lines 321, 332: Variable $adodb_obj might have not been defined 

Please have a look and decide whether any action is necessary.

PR #112, opened 2015-05-27
Noblob optimization

ipso:


drivers/adodb-postgres64.inc.php
```diff
@@ -737,7 +737,7 @@ function _connect($str,$user='',$pwd='',$db='',$ctype=0)
 		# PHP does not handle 'hex' properly ('x74657374' is returned as 't657374')
 		# https://bugs.php.net/bug.php?id=59831 states this is in fact not a bug,
 		# so we manually set bytea_output
-		if (version_compare($info['version'], '9.0', '>=')) {
+		if ( !empty($this->connection->noBlobs) AND version_compare($info['version'], '9.0', '>=')) {
```
dregad: `&&` not AND

dregad:
@Mike-Benoit I had a closer look at  this old contribution of yours today, while checking a static analysis warning.

I didn't realize it back then, but ADOConnection object does not have a _connection_ property, and this _noBlobs_ flag is not documented or used anywhere else... 

Can you please shed some light on this ?

dregad:
@Mike-Benoit since you never provided any feedback on my earlier note https://github.com/ADOdb/ADOdb/pull/112#issuecomment-902028719, I'm going to revert this change.

Mike-Benoit:
My apologies for the lack of response, I must had missed the original notification. 

```
                # PostgreSQL 9.0 changed the default output for bytea from 'escape' to 'hex'
		# PHP does not handle 'hex' properly ('x74657374' is returned as 't657374')
		# https://bugs.php.net/bug.php?id=59831 states this is in fact not a bug,
		# so we manually set bytea_output
		if (version_compare($info['version'], '9.0', '>=')
			&& version_compare($info['client'], '9.2', '<')
		) {
			$this->Execute('set bytea_output=escape');
		}
```

It appears the above code is to work around some changes in PostgreSQL 9.0 to 9.2. v9.2 went end-of-life as of November 2017. Would you be okay with removing this code completely? It does cause performance issues when not using blobs (I would argue the most common situation) and if I recall causes problems with the ADODb load balancer due to deferred query handling.

While we are it, this code could probably be removed too:

```
		$info = $this->ServerInfo(false);

		if (version_compare($info['version'], '7.1', '>=')) {
			$this->_nestedSQL = true;
		}
```

PR #116, opened 2015-06-10
Added data dictionary support for PDO.

CreatrixTeam:
Added data dictionary support for PDO which should allow ADODB extensions using the data dictionary to run successfully when using PDO. However the support is unlikely to run on the older versions of PHP that ADODB was originally designed for before.


CreatrixTeam:
Closing pull request. Class no longer required in our latest commit to our fork's master branch.

PR #118, opened 2015-06-14
Make NOT NULL/DEFAULT support for BLOB/TEXT colums configurable per datadict

mjonuschat:
Instead of globally disabling the support for NOT NULL and DEFAULT for
BLOB and TEXT columns introduce two flags in the datadict that allow
enabling these features individually per dictionary.

Enable NOT NULL and DEFAULT for PostgreSQL BLOB/TEXT
Enable NOT NULL for MySQL BLOB/TEXT


dregad:
@mnewnham I'm thinking of merging this in, any objections, suggestions for better implementation before I do ?

mnewnham:
I think that's fine, generally we haven't cared in the past whether a particular feature wasn't supported when building schemas, so I'm surprised  that it was there in the first place. We should probably research which databases don't support not null, It could be an old feature we can eliminate completely

Issue #121, opened 2015-06-16
Broken FetchField function in adodb-pdo.inc.php


CreatrixTeam:
The commit https://github.com/ADOdb/ADOdb/commit/a36f5c4276f7bbf8fc79039a950b0c68d06d2099 which was intended for coding guidelines broke the FetchField method. The commit replaced:

if (ADODB_ASSOC_CASE == 0) $o->name = strtolower($o->name);
else if (ADODB_ASSOC_CASE == 1) $o->name = strtoupper($o->name);
return $o;

with 

switch(ADODB_ASSOC_CASE == 0) {
	case ADODB_ASSOC_CASE_LOWER:
		$o->name = strtolower($o->name);
		break;
	case ADODB_ASSOC_CASE_UPPER:
		$o->name = strtoupper($o->name);
		break;
	default:
		return $o;
}

which are not equivalent. The function now returns nothing sometimes.

dregad:
Should have been `switch(ADODB_ASSOC_CASE) {` and move the return out of the switch statement. Sorry about that.

dregad:
Proposed fix in PR #122

PR #122, opened 2015-06-16
Fixes issue https://github.com/ADOdb/ADOdb/issues/121

CreatrixTeam:


dregad:
For future patch submissions, may I kindly ask you to 

1. refrain from opening *both* an issue *and* a pull request like you just did, it's redundant...  
If you have a patch, it is sufficient to put everything in the PR; the only case where we would have both, is if you're submitting a patch for a pre-existing issue
2. make sure each commit message includes a complete description of the issue it addresses. I suggest reading the [Git commit message guidelines](https://www.mantisbt.org/wiki/doku.php/mantisbt:git_commit_messages) I wrote some time ago for MantisBT.

Thanks.

CreatrixTeam:
Thank you.

Regarding 1,  if you would give us permission, we would like to continue doing things the way we are doing them now. We use issue to discuss, track and confirm. A submitted issue does not guarantee an issue exists. An issue exists regardless whether we patch or not. And a patch does not guarantee a request for pull. If you do not give permission, we shall abide by your request. We are waiting for your response.

dregad:
From my perspective, tracking and confirmation can be done just as effectively in a pull request. Opening an issue on top of the PR just creates additional overhead, even more so if the PR is just a series of references without any actual text (like this one). That makes keeping track and reviewing things more difficult for me. If a patch is rejected, or the underlying issue determined not to be valid, the PR will simply be closed without merging.

CreatrixTeam:
Thank you. 

Perhaps we did not explain our selves well. One might think there is an issue, but it would be better to wait for confirmation before starting to program something. In any case, we shall abide by your request, at least in the case where we are sure an issue is indeed an issue like in this case, and we shall do our best to abide in other cases.

dregad:
Clearly in such case where you're not sure about an issue and the patch is not something trivial like this PR, then a 2-step approach is the better approach... thus no objection to have both issue + PR when necessary. Use common sense but please make the PR title more descriptive than a simple link.

Issue #126, opened 2015-06-18
Double quotation with xmlschema03


CreatrixTeam:
Going through code and documentation, it has become apparent that the library allows a person to include fields, table names and other identifiers in the xml schema quoted with `. The data dictionary part of the ADODB is then responsible to replace ` with `,  ", ' or whatever quotation used by the underlying database. This can be seen in the code, such as the function NameQuote in the file "adodb-datadict.inc.php", and in documentation, http://adodb.sourceforge.net/docs-datadict.htm. For example, one should be allowed to declare a table using:
     <table name="tableName">
or
     <table name="`tableName`">
where the second form would ensure the table name is properly quoted, which would avoid syntax errors if the table name is a reserved keyword, and in some databases would make the table name case sensitive. We have become dependent on this feature.

It appears that the data dictionary feature has been introduced later along with AXMLS, and AXMLS relies on this feature. Unfortunately ADODB itself does not account for this feature, at least not everywhere. A particular problem, for example, has been found with the function MetaColumns in the file "adodb-mysql.inc.php" where that function hard codes the quotation on the table on the sql statements it creates. This can been seen in the line:
	var $metaColumnsSQL = "SHOW COLUMNS FROM `%s`";
That line was previously:
              var $metaColumnsSQL = "SHOW COLUMNS FROM %s";

The change has been introduced with version 491, and although we do not know why, the change has been in line with all other drivers, apart from apparently "adodb-fbsql.inc.php". The function never expected the ` quotation feature introduced by data dictionary. To further understand the problem, please examine and contrast the contradicting behavior of the function MetaColumns in the class ADODB_DataDict in the file "adodb-datadict.inc.php". 

When the XML with quoted table names is parsed and the MetaColumns function is called, the created query contains the data dictionary feature quotation, `, processed and replaced with the correct identifier quotations per the data dictionary feature, but then quoted again by MetaColumns in  "adodb-mysql.inc.php" (and other drivers), causing errors. To fix this properly, the ADODB library needs to be normalized with this feature in mind.

We are unsure where to proceed from here. We are still contemplating whether we should propose our own fix or not. The fix might require changing a good number of files, and progress has appeared to be very slow with this project so far. We do hope the project continues to move forward.

dregad:
I originally took over this library's from John Lim, who lacked time and lost interest in maintaining it. It has never been my objective to make it evolve significantly, my initial priority was to fix issues which would have been difficult or impossible to circumvent downstream.

[...]

PR #127, opened 2015-06-19
Regarding "https://github.com/ADOdb/ADOdb/issues/126".

CreatrixTeam:
(Please note that the issue "https://github.com/ADOdb/ADOdb/issues/126" has been deemed a new feature request not a bug fix by the Maintainer of Adodb)

Going through code and documentation, it has become apparent that the library allows a person to include fields, table names and other identifiers in the AXMLS xml schema quoted with back-quotes (`). ADODB Data Dictionary is then responsible to replace back-quotes(`) with back-quotes(`), ("), (') or whatever quotation used by the underlying database. This can be seen in the code such as the function NameQuote in the file "adodb-datadict.inc.php", and in documentation, http://adodb.sourceforge.net/docs-datadict.htm. For example, one should be allowed to declare a table using:

     <table name="tableName">
or
     <table name="`tableName`">

where the second form would ensure the table name is properly quoted, which would avoid syntax errors if the table name is a reserved keyword, and in some databases would make the table name case sensitive.

It appears that ADODB Data Dictionary has been introduced later along with AXMLS, and AXMLS relies on this ADODB Data Dictionary. Unfortunately ADODB itself does not account for ADODB Data Dictionary's quotation feature, at least no where we have examined. A particular problem, for example, has been found with the function MetaColumns in the file "adodb-mysql.inc.php" where that function hard codes the quotation on the table in the sql statements it creates. This can been seen in the line:
var $metaColumnsSQL = "SHOW COLUMNS FROM `%s`"

That line was previously:
var $metaColumnsSQL = "SHOW COLUMNS FROM %s";

The change has been introduced with version 491, and although we do not know why, the change has been in line with all other drivers, apart from apparently "adodb-fbsql.inc.php". The function never expected the (`) quotation feature introduced by ADODB Data Dictionary. To further understand the problem, please examine and contrast the contradicting behavior of the function MetaColumns in the class ADODB_DataDict in the file "adodb-datadict.inc.php". This contradicting behavior is why we are convinced this is a bug.

When the XML with quoted table names is parsed and the MetaColumns function is called, the created query contains the ADODB Data Dictionary feature quotation, (`), processed and replaced with the correct identifier quotations per the ADODB Data Dictionary feature, but then quoted again by MetaColumns in "adodb-mysql.inc.php" (and other drivers), causing errors. To fix this properly, the ADODB library needs to be normalized with this feature in mind.

Although the ultimate aim is to normalize this feature in the entirety of ADODB, we are taking a simpler approach of fixing the calls made by ADODB Data Dictionary to ADODB. Please note the following:
- The approach assumes ADODB is completely unaware of the quotation feature. This has proven true everywhere in the code we looked.
- Un awareness of the feature is assumed when the code adds its own quotation, whether identifier quotation or string quotation. When no quotation is found, it could be awareness, a bug, or simply something specific to the particular database sql syntax.
- It appears the function ADOConnection::MetaColumns is unaware of the quotation feature, apart from possibly the driver "adodb-fbsql.inc.php". All drivers were examined to arrive at this conclusion.
- It appears the function ADOConnection::MetaIndexes is unaware of the quotation feature, apart from the drivers "adodb-mysqli.inc.php" and "adodb-mysql.inc.php", which we can conclude to be bugs, not awareness, due to our familiarity with Mysql. All drivers were examined to arrive at this conclusion.
- It appears the function ADOConnection::MetaPrimaryKeys is unaware of the quotation feature, apart from possibly the driver "adodb-odbtp.inc.php". All drivers were examined to arrive at this conclusion.

### DIFF
--- a/adodb-datadict.inc.php
+++ b/adodb-datadict.inc.php
@@ -197,22 +197,25 @@ class ADODB_DataDict {
 		return $this->connection->MetaTables();
 	}
 
+	//NOTE: THIS FUNCTION PROMISES TO QUOTE THE TABLE NAME, BUT THE PROMISE IS KEPT BY THE CORE ADODB
 	function MetaColumns($tab, $upper=true, $schema=false)
 	{
 		if (!$this->connection->IsConnected()) return array();
-		return $this->connection->MetaColumns($this->TableName($tab), $upper, $schema);
+		return $this->connection->MetaColumns($this->TableNameWithNoNameQuote($tab), $upper, $schema);
 	}
 
+	//NOTE: THIS FUNCTION PROMISES TO QUOTE THE TABLE NAME, BUT THE PROMISE IS KEPT BY THE CORE ADODB
 	function MetaPrimaryKeys($tab,$owner=false,$intkey=false)
 	{
 		if (!$this->connection->IsConnected()) return array();
-		return $this->connection->MetaPrimaryKeys($this->TableName($tab), $owner, $intkey);
+		return $this->connection->MetaPrimaryKeys($this->TableNameWithNoNameQuote($tab), $owner, $intkey);
 	}
 
+	//NOTE: THIS FUNCTION PROMISES TO QUOTE THE TABLE NAME, BUT THE PROMISE IS KEPT BY THE CORE ADODB
 	function MetaIndexes($table, $primary = false, $owner = false)
 	{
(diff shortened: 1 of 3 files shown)


CreatrixTeam:
Closing to avoid unforeseen changes from our current master should this pull get accepted; changes shall be re proposed with master.

PR #128, opened 2015-06-20
Added MetaxIndexes support for PDO drivers. The code is simply verbat…

CreatrixTeam:
Added MetaxIndexes support for PDO drivers. The code is simply verbatim from other drivers and should work as well and as bad as those others drivers. We confirmed this for both Mysql and PostgreSQL. The support should allow code relying on MetaxIndexes, such AXMLS, to work properly.


CreatrixTeam:
Closing to avoid unforeseen changes from our current master should this pull get accepted; changes shall be re proposed with master.

PR #130, opened 2015-06-27
Sqlite driver array error.

CreatrixTeam:
Sqlite has automatically created indexes which can be safely assumed to be unaccounted for by calling code, such as AXMLS, as well as different to normal indices in terms of what may be done with them. Without this fix, a number of failures happen starting with the line

"$cols = explode(")",$cols[1]);"
later in the same function (MetaIndexes), as well as in AXMLS at least. The fix is a simple check to ensure the element '1' in the $row variable is not empty, which we observed happens with the automatically created indexes.

### DIFF
--- a/drivers/adodb-sqlite.inc.php
+++ b/drivers/adodb-sqlite.inc.php
@@ -327,6 +327,9 @@ class ADODB_sqlite extends ADOConnection {
 			if ($primary && preg_match("/primary/i",$row[1]) == 0) {
 				continue;
 			}
+			//IGNORE AUTOMATICALLY CREATED INDICES
+			if (empty($row[1]))
+				{continue;}
 			if (!isset($indexes[$row[0]])) {
 				$indexes[$row[0]] = array(
 					'unique' => preg_match("/unique/i",$row[1]),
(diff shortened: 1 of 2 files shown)


dregad:
Can you please 

1. make sure your commit messages comply with rules defined in https://www.mantisbt.org/wiki/doku.php/mantisbt:git_commit_messages
2. always rebase your branch on top of master prior to submitting

CreatrixTeam:
Thank you. Regarding:
1- Regarding one, we shall read it and adhere as much as feasible.
2- We are trying hard with this one. We shall re update our branches with master as soon as possible. You are using the word rebase, which we are assuming is a git term since we came across it. At some point it was decided against. If you can explain your motivation for a re base instead of mergers as we are currently doing them, we appreciate it. Note that our mergers are such as, only the changes in our original proposal are changes in the after merger.

dregad:
I don't have time for a lengthy dissertation on merge vs rebase. In short this prevents merge conflicts while maintaining a clean (read: as linear as possible) history - it allows me to perform a fast-forward merge so the history looks like (where *a* is the submitted commit)
```
* -- a
```
vs a "regular" merge
```
* --- m 
 \   /
   a
```

CreatrixTeam:
Thank you. Whatever time you manage to offer us, we are appreciating it. We shall look into the re base functionality again, and hope we can eventually meet your requirements or reach a compromise.

PR #150, opened 2015-09-11
Making ? (parameter) detection in queries more robust

orcinus:
This patch makes changes in commit 1a4befc14c52a96591eca2424ff87eabe7d233e2 a little less "destructive" and more in line with the existing parameters mismatch errors; and makes the "?" token (parameter) detection in queries more robust.

1) 1a4befc14c52a96591eca2424ff87eabe7d233e2 would throw an exception / error and return false if a query with no parameters was passed alongside a non-empty input array. This patch makes it simply throw an error and continue, but IGNORING the input array completely. Disabling error handling / exceptions will, thus, not break an application, but will still behave correctly.

2) Previously, queries were simply exploded over "?" token as part of the query parsing flow. This means that having a hardcoded string in a query, with a ? in it could result in unpredictable behaviour. This has been dealt away with, by first normalizing the query so it only uses single quotes (while protecting double-escaped double quotes), then exploding it by quotes, then processing the parts that are outside of quotes normally (explode by "?") while ignoring the parts that are inside quotes.

No regexps were used, only array manipulations and str_\* functions.

It's not perfect, but it should be a tiny bit more robust than the original version.
It's been tested on an actual live app (in a dev environment) and via a few ad-hoc test scripts, but i might've missed something - more thorough testing is advised.

### DIFF
--- a/adodb.inc.php
+++ b/adodb.inc.php
@@ -1102,6 +1102,7 @@ if (!defined('_ADODB_LAYER')) {
           $j++;
         }
         $sql = implode("'", $sql);  // restore original query
+				
 
 				// Get number of parameters (tokens)
 				$nparams = sizeof($sqlarr)-1;
--- a/adodb.inc.php
+++ b/adodb.inc.php
@@ -1076,8 +1076,34 @@ if (!defined('_ADODB_LAYER')) {
 			unset($element0);
 
 			if (!is_array($sql) && !$this->_bindInputArray) {
-				// @TODO this would consider a '?' within a string as a parameter...
-				$sqlarr = explode('?',$sql);
+				// Normalize any hardcoded strings in the query to single quotes
+				$sql = explode("\\\"", $sql);	// protect escaped double quotes
+				while(list(, $x) = each($sql))
+				  $sqlClean[] = str_replace('"', "'", $x);
+				$sql = implode('\\\"', $sqlClean);	// restore escaped double quotes
+
+				// Split by tokens, ignoring tokens in quotes
+				$sql = explode("'", $sql);
+				$j = 0; $prevSeg = 0;
+				while(list(, $v) = each($sql)) {
+				  if($j%2 == 0) {
+				    // if even, we're outside the quotes
+				    $exp = explode("?", $v);
+				    $sqlarr[$prevSeg] .= $exp[0];
+				    unset($exp[0]);
+				    array_merge($sqlarr, $exp);
+				    while(list(, $w) = each($exp))
+				      $sqlarr[] = $w;
+				    $prevSeg = $j + (count($exp));
+				  } else {
+				    // if odd, we're inside the quotes
+				    $sqlarr[$prevSeg] .= "'".$v."'";
+				  }
+				  $j++;
+				}
+				$sql = implode("'", $sql);	// restore original query
+
+				// Get number of parameters (tokens)
 				$nparams = sizeof($sqlarr)-1;
 
 				// If there are no '?' tokens in the query and there is a an input array (has to be
--- a/adodb.inc.php
+++ b/adodb.inc.php
@@ -1076,32 +1076,32 @@ if (!defined('_ADODB_LAYER')) {
 			unset($element0);
 
 			if (!is_array($sql) && !$this->_bindInputArray) {
-				// Normalize any hardcoded strings in the query to single quotes
-				$sql = explode("\\\"", $sql);	// protect escaped double quotes
-				while(list(, $x) = each($sql))
-				  $sqlClean[] = str_replace('"', "'", $x);
-				$sql = implode('\\\"', $sqlClean);	// restore escaped double quotes
+        // Normalize any hardcoded strings in the query to single quotes
+        $sql = explode("\\\"", $sql); // protect escaped double quotes
+        while(list(, $x) = each($sql))
+          $sqlClean[] = str_replace('"', "'", $x);
+        $sql = implode('\\\"', $sqlClean);  // restore escaped double quotes
 
-				// Split by tokens, ignoring tokens in quotes
-				$sql = explode("'", $sql);
-				$j = 0; $prevSeg = 0;
-				while(list(, $v) = each($sql)) {
-				  if($j%2 == 0) {
-				    // if even, we're outside the quotes
-				    $exp = explode("?", $v);
-				    $sqlarr[$prevSeg] .= $exp[0];
-				    unset($exp[0]);
-				    array_merge($sqlarr, $exp);
-				    while(list(, $w) = each($exp))
-				      $sqlarr[] = $w;
-				    $prevSeg = $j + (count($exp));
-				  } else {
-				    // if odd, we're inside the quotes
-				    $sqlarr[$prevSeg] .= "'".$v."'";
-				  }
-				  $j++;
-				}
-				$sql = implode("'", $sql);	// restore original query
+        // Split by tokens, ignoring tokens in quotes
+        $sql = explode("'", $sql);
+        $j = 0; $prevSeg = 0;
+        while(list(, $v) = each($sql)) {
+          if($j%2 == 0) {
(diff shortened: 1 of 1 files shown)


orcinus:
Pardon the indent mess and subsequent fixes.

dregad:
> 1) 1a4befc would throw an exception / error and return false if a query with no parameters was passed alongside a non-empty input array. This patch makes it simply throw an error and continue, but IGNORING the input array completely. Disabling error handling / exceptions will, thus, not break an application, but will still behave correctly.

I would much rather have the code break, instead of having some query getting executed with the wrong params and without my knowledge (could happen with your patch, e.g. with a query having 1 param and an array with 2).

As a matter of fact, this is also how PDO handles this situation [1]: "You cannot bind more values than specified; if more keys exist in input_parameters than in the SQL specified in the PDO::prepare(), then the statement will fail and an error is emitted. "

I would argue that the way I implemented this, is the safer (dare I say proper ?) way to deal with an invalid parameter array. 

[1] http://php.net/manual/en/pdostatement.execute.php

orcinus:
> I would much rather have the code break, instead of having some query getting executed with the wrong params and without my knowledge (could happen with your patch, e.g. with a query having 1 param and an array with 2).

It will break if error or exception handling is included. If it's not, it will not break, but will return 
That is *exactly* how it's been behaving up until now. 1a4befc changed that behaviour, but inconsistently - it would've failed miserably for a 2D parameter array.

You're right, returning a false is safer, and i'd personally rather have it break too, regardless of error / exception. I was just trying to stay consistent with the old behaviour, which is - to respect the exception handler.

Easily tweaked, though - all it needs is a `return false;` after both, the new and the old exceptions.

mnewnham:
Ok, I'm a bit late into this discussion, but:

[This recent commit to the oracle driver] (https://github.com/ADOdb/ADOdb/commit/eb9d83915fcfdcd4a4fb3f65a586e6c60be598f7) gives guidance on how to break queries up using regexps and process them in order to match  the number of actual bind parameters to ? characters in a query and prevent exactly the exploit that is being discussed.

This statement:

> This has been dealt away with, by first normalizing the query so it only uses single quotes (while protecting double-escaped double quotes),

Concerns me. Is this part of this suggested merge? I could not find changes related to this

orcinus:
> This recent commit to the oracle driver gives guidance on how to break queries up using regexps
Parsing queries using regexps is easy. Maintaining a low profile and doing it *without* regexps is not ;)

>Concerns me. Is this part of this suggested merge? I could not find changes related to this
Yes it is. Line 1079 onwards. Right after the comment saying "Normalize any hardcoded strings in the query to single quotes".

orcinus:
By the way, this (from eb9d83915fcfdcd4a4fb3f65a586e6c60be598f7):

```php
$sql = str_replace('-QUESTIONMARK-', '?', $sql);
```

... is a kludge.

mnewnham:
1. Agree totally, but it is a 15 year-old (now a teenager) successful kludge
2. Your method of arbitrarily exchanging unescaped double quotes for single quotes is going to break the DB2 driver. In order to gain access to mixed and lower case table and column names, they have to be 'protected' by double quotes, for example, you may have 2 tables, **ACCOUNTS** and **accounts** in the same database.  

a statement like:
```php
SELECT column1, column2 from accounts where account_name = 'ADODB'
```
will always find the table **ACCOUNTS**

in order to access the table **accounts**, we must write the SQL like this:
```php
SELECT column1, column2 from "accounts" where account_name = 'ADODB'. 
```
note that the table name is double quoted and that the string is single quoted. In fact, if we created the column names in lower case, we must double quote them as well:

```php
SELECT "column1", "column2" from "accounts" where account_name = 'ADODB'
```

Use of double quotes to obtain access to the non-uppercase table name is, unfortunately for us, mandatory. So, whilst I agree that the non-regexp method is a lower footprint, It is simply not going to achieve the portable result we need.

orcinus:
Ah! True, ANSI SQL standard calls for double quotes around tables and columns.
Got used to backticks so much i completely forgot about that.

I've got an idea for a regexpless solution (replacing and storing original delimiters in a LUT as you go through the query, then reverting them after the query is exploded), but will need to work on it a bit.

I know this all seems like obscene micro-optimization, but i really really don't feel comfortable with suddenly adding ~70 preg_matches per page load (using the app i'm currently working on as an example) :)

mnewnham:
While I might have agreed with your premise 20 years ago where every optimization counted, I'm not sure now that with processor speeds, it's really worth your time, for example:

my core I7 running Windows 10 takes 0.06 seconds to explode 100,000 different SQL statements by a single quote - preg_split takes 0.16 seconds. 

my core I5 running Centos takes 0.06 seconds to explode 100,000 different SQL statements by a single quote - preg_split takes only 0.08 seconds. 

If I was in a transaction processing environment where that .1 of a second really counted, I probably wouldn't be using ADOdb in the first place, and let's not lose sight of our mission, which is portability first, then speed

orcinus:
That benchmark isn't a very good argument, as your i7/i5 is not handling hundreds of threads doing that simultaneously, a second. Things add up very quickly once you go beyond a single process.

But i understand your point.
If i can figure out an easy / non-overly-complex way to fix it, i'll do that, if not, i'll rewrite with a regexp. Would that be okay?

mnewnham:
Yes of course, we're never going to refuse a fix, but I have to reinforce that a patch to adodb.inc.php affects every single driver, so there's always going to be a discussion like this. We accept that the code is not optimal in many places, which means that any patches need to be scrutinized for unexpected effects. That being said - go for it.

peterdd:
Another approach would be:

For databases that support prepared statements, enhance the adodb driver to use it.
(have mysqli driver in my mind..)

So we don't need to fallback to a fake emulation hurting our brains. :-)

orcinus:
Great idea, but there are pitfalls (as always).

Native prepared statements don't support all queries. The driver would have to check the first word against a list of allowed statements and fall back to emulation if it's not one of them.

Also, it would likely be somewhat slower for setups that don't keep the DB on the same machine as the app (two trips to DB per query, i.e. one extra trip, unless you're already using e.g. mysqli::real_escape_string for escaping, which wouldn't be necessary anymore with a native prepared statement).

It would be one helluva cleaner, though :)

orcinus:
@mnewnham That opens another venue to investigate, though - the whole parameter parsing logic could be moved to the driver, or - better yet - overridden by driver. So it always falls back to, say, the regexp, which is guaranteed to work, but if a driver specifies its own prepared statement handler, it uses that.

mnewnham:
I don't want to get too negative on this, but I think you could easily be letting the scope of your fix get completely out of hand. Perhaps we could look at it like this:

1. Lets complete the minimal amount of code necessary to address the parameter mismatch/ possible attack vector issue. This could even be a short-term regexp fix while you test out and benchmark a non-regexp solution.

2 A more generalized investigation/prototyping of a better way of handling prepared statements  as a much more long-term discussion. You need to take into account that there are differences in prepared statement handling across databases, and there are commonalities in the ADOdb code in prepared statement and stored procedure handling.  As a general observation, I believe the code that deals with this functionality is quite fragile (viz. --QUESTIONMARK--)  so we need to tread lightly. This isn't something we would want to just pull into master.

orcinus:
Sorry for dropping off radar. Don't take it as giving up, i had quite a busy week and couldn't really find any time to devote to this. You're right, i might be going way off as far as the scope is concerned.

For what it's worth, i did manage to do some more testing on the non-regexp patch and found a pretty serious bug (that causes it to append a single quote at the end of a statement under some circumstances).

dregad:
@orcinus what should we do with this pull request ?

orcinus:
Argh, sorry again, i'm swamped with work and never got around to fixing the bug.
Please close, can always be reopened when there's something to work with.

dregad:
OK, thanks for the feedback.

PR #151, opened 2015-09-14
Return -1 by default, for sqlite3 _numOfRows

paulproteus:
This relies on the built-in ability to rely on ADODB_COUNTRECS.

Before this patch, we were setting _numOfRows to positive 1, which meant
that any query that checked how many rows were available in a queryset
would see literally the number 1, which was not accurate for queries that
returned 0 rows or queries that returned more rows.

Note that returning -1 is also what the ADOdb Oracle driver does, for
example. In the case of sqlite3, there appears to be no efficient
way to ask the PHP sqlite3 driver how many rows come back from
a query, so returning -1 is the best we can do.

### DIFF
--- a/drivers/adodb-sqlite3.inc.php
+++ b/drivers/adodb-sqlite3.inc.php
@@ -394,7 +394,6 @@ class ADORecordset_sqlite3 extends ADORecordSet {
 
 	function _initrs()
 	{
-		$this->_numOfRows = 1;
 		$this->_numOfFields = $this->_queryID->numColumns();
 
 	}


paulproteus:
Hi team! Let me know if there's anything I can do to help you review/comment on/approve/reject this pull request.

I've manually tested it by calling the `->RecordCount()` function (with the help of @zarvox) and it reliably returns the number of rows in a query in my testing.

I'm happy to look into other tests as needed for it to be something the ADOdb team is willing to merge. So let me know, and thanks for reviewing!

mnewnham:
If a driver does not natively support record counting for a  query, it should indeed always return -1 

This is actually defined in the base ADORecordSet class as -1, (adodb.inc.php)  so the correct implementation should be to actually remove the reference in sqlite3, so that it defaults to the class value

paulproteus:
Thanks for the comment @mnewnham ! I've added a commit on top that removes this line entirely. @zarvox has tested that removing the line works, too.

I could submit a pull request with just 1 commit by squashing these together, or I could leave this as-is with 2 commits. Your call @mnewnham -- I'm so grateful for your very speedy review!

mnewnham:
Could you bring these into a single commit please. @dregad , we could pull this into master, I think. Thanks for testing, @zarvox

paulproteus:
I clarified the commit log message and squashed to one commit. Ready for fresh review @mnewnham . Thanks!

paulproteus:
:cake: :penguin: :yellow_heart: 

WAHOO!

Thanks a great deal. This is super exciting for me. (And hopefully makes @zarvox's life easier as he packages a ADOdb-using web app for Sandstorm.)

dregad:
You're welcome. Thank YOU for your contribution !

PR #158, opened 2015-10-24
utf8 charset support in mssql native

azghanvi:
add utf8 charset support - http://www.databaseskill.com/228701/

### DIFF
--- a/drivers/adodb-mssqlnative.inc.php
+++ b/drivers/adodb-mssqlnative.inc.php
@@ -480,6 +480,10 @@ class ADODB_mssqlnative extends ADOConnection {
 		$connectionInfo["Database"]=$argDatabasename;
 		$connectionInfo["UID"]=$argUsername;
 		$connectionInfo["PWD"]=$argPassword;
+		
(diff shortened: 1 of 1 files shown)


mnewnham:
I agree that this is an valuable enhancement to the driver.

 Unfortunately, the connect() method already has a 5th argument defined ($forceNew), which forces the driver to establish a new connection. In order to avoid PHP strict standards errors, we cannot add further arguments to the connect statement just for the mssqlnative driver.

We do, however, have a plan to add functionality that will allow configuration of connection parameters prior to issuing the connection request. this will be available in a later release of ADOdb. I'm going to hold this pull request until we have that available.

azghanvi:
Thanks for the feedback!
Appreciated.

mnewnham:
I have added a new public method , setConnectionParameter($parameter,$value) which addresses this:

```php
$db = newAdoConnection(......);
$db->setConnectionParameter('CharacterSet','UTF-8');
$db->connect(.....);
```

You can pull this from trunk or wait for a new release

PR #161, opened 2015-11-11
start for using travis-ci

peterdd:
no real tests or real config/setup done, but at least travis-ci runs and gives green buttons as a first start for further development. :-)


peterdd:
currently the travis-ci webhook is done for my account, can be replaced when someone of organisation ADOdb does it for ADOdb/ADOdb.

Can be done in the project settings on github.com: ADOdb/ADOdb ->Settings->'Webhook & services'->
'Add service', select travis-ci from the dropdown list and follow further steps on travis-ci.

dregad:
Considering 
- Travis' change in policy (migration from travis.org to travis.com and their current open-source support policy which I think is crap, 
- the fact that we don't even have a test suite worth the name 
 
I don't see the point in this PR anymore, so I'm closing it. We can investigate alternative approaches (GitHub Actions, CircleCI, etc) if and when ready with a test suite.

Issue #168, opened 2015-12-14
PHP7 compatibility issues: PHP4 constructors, pcre '/e' modifier


marinaglancy:
Hello,
I thought you might be interested in our experience when upgrading ADOdb in Moodle to version 5.21.
You can find full history on the issue in moodle tracker: https://tracker.moodle.org/browse/MDL-52286

First of all, one of our unittests failed on Mysql and it looks like an ADOdb bug when closing recordset. This line https://github.com/ADOdb/ADOdb/blob/master/drivers/adodb-mysqli.inc.php#L1060 produces a notice "mysqli_more_results() expects parameter 1 to be mysqli, boolean given".
*EDIT [dregad] this is covered in #170*

Second, there are still two places that php7cc picked up as incompatible with PHP7 (/e modifier is removed from PHP7 and forgotten old-style constructor):
https://github.com/ADOdb/ADOdb/blob/master/drivers/adodb-db2ora.inc.php#L41
https://github.com/ADOdb/ADOdb/blob/master/drivers/adodb-mysqlt.inc.php#L133

dregad:
Hello Marina,

Many thanks for your feedback and contribution, it is greatly appreciated. I will look into these as soon as possible.

In future bug reports, may I kindly ask you to please keep things separate, i.e. 1 bug = 1 issue report ? TIA.

marinaglancy:
oh, thank you. I was sure I read 5.21 in the latest tag but can't see it now. I corrected the version number in moodle

PR #169, opened 2015-12-14
Mysqli notice and two PHP7 fixes

marinaglancy:
Reported also in https://github.com/ADOdb/ADOdb/issues/168

### DIFF
--- a/drivers/adodb-db2ora.inc.php
+++ b/drivers/adodb-db2ora.inc.php
@@ -38,7 +38,7 @@ global $_COLONARR,$_COLONSZ;
 	$_COLONARR = array();
 	$_COLONSZ = sizeof($arr);
 
-	$sql2 = preg_replace("/(:[0-9]+)/e","_colontrack('\\1')",$sql);
+	$sql2 = preg_replace_callback('/(:[0-9]+)/', create_function('$m', 'return _colontrack($m[0]);'), $sql);
 
 	if (empty($_COLONARR)) return array($sql,$arr);
 
--- a/drivers/adodb-mysqli.inc.php
+++ b/drivers/adodb-mysqli.inc.php
@@ -1057,7 +1057,7 @@ class ADORecordSet_mysqli extends ADORecordSet{
 		//if results are attached to this pointer from Stored Proceedure calls, the next standard query will die 2014
 		//only a problem with persistant connections
 
-		while(mysqli_more_results($this->connection->_connectionID)){
+		while(@mysqli_more_results($this->connection->_connectionID)){
 			@mysqli_next_result($this->connection->_connectionID);
 		}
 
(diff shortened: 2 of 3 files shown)


dregad:
*Note to self: next time, comment on the* files changed *tab and not directly in the commits...*

marinaglancy:
apologies for putting separate changes in the same commit. I absolutely do not mind you replacing my commit(s) with another solution and rejecting this pull request. We can also pull your changes into moodle instead of my fix, we are only testing now and will not release until the end of this week.

I added a unittest for the _colonscope function in moodle (because there is pretty much no other way we can test it), I have just run it with $m[1] and it works the same: https://github.com/marinaglancy/moodle/commit/cc3048eb410396c3230c8a76824c859b5f597c3e
Also the use of "create_function()" is inline with moodle coding style, I know that most of other projects prefer anonymous functions.

As for the constructor, I thought that you may prefer to remove it completely because it repeats the parent one.

marinaglancy:
Also no preferences on @ operator. The bigger question is if this is actually the correct place to put the fix for the error we are experiencing. We were trying to figure out why it is happening at all.

This is the error we were having: https://tracker.moodle.org/browse/MDL-52286?focusedCommentId=385668&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-385668
My first intention was to remove destructor but then I started looking why other db, such as postgres, pass our unittests and noticed the @ in the _close() method.

marinaglancy:
update from our integration team - we just hit the same on mssql (php5): https://tracker.moodle.org/browse/MDL-52512
http://pastebin.com/UaLnGV2c
I'll create a separate request, I promise :)

marinaglancy:
I created #170 for the destructor-related failures

dregad:
> the use of "create_function()" is inline with moodle coding style, I know that most of other projects prefer anonymous functions

My point was not against use of `create_function()` vs a closure, but rather to call `_colontrack()` directly as explained in https://github.com/marinaglancy/ADOdb/commit/dff32013c1f2f1e51ea8be59d447280d60362e33#commitcomment-14962888

> As for the constructor, I thought that you may prefer to remove it completely because it repeats the parent one.

You're right.

> Also no preferences on @ operator. The bigger question is if this is actually the correct place to put the fix for the error we are experiencing. 

There is way too much use of @ operator for my taste in this library ;-)

dregad:
preg_replace issue (#168) has been fixed in bdeb28f214c29a54a88cef995e251702f7181ab7 and follow-up commit c32f1525136ff3afcf1e3ed30a17818b1bcd1731.

Destructor issue (#170) is fixed in 59e93d9b2ed7e8fc040f35a1eb228ae222116112, dcbe478834b607f0f572b934fbc1b4e137688a7f and 745a02a7533b744c1a14a8bb0b41fea96c41ac14

PR #190, opened 2016-01-08
Security Improvements

SecMyth:
There might be the possibility to perform SQL-Injection attacks on SelectLimit via nrows and offset.
Additionally the session manager misses to cover the cookie flag httponly.


dregad:
Many thanks for your contribution !

Issue #199, opened 2016-02-17
BulkBind is broken in 5.20.X


10n:
In Execute method after $array_2d is detected as TRUE (line:1109)  the param count validation is not compatible with bulk bind:

				$sqlarr = explode('?',$sql);
				$nparams = sizeof($sqlarr)-1;
				// Make sure the number of parameters provided in the input
				// array matches what the query expects
				if ($nparams != count($inputarr)) {

Issue #132, opened 2015-06-28
Unwanted replacement of '?' within strings in ocipo driver


dregad:
As reported by @mnewnham by e-mail.

Issue with the adodb-oci8po.inc.php driver,  the scenario is this:

An SQL statement is sent that contains a ? character in a string field, e.g. 

```
SELECT * FROM users WHERE user_code = 'ABC?DEF'
```

adodb sees the ? field, and assumes it is for a bind variable. The number of variables then no longer matches the number passed, and crashes the function. I recognize it's pretty obscure (I've been using the driver for more than 10 years and this is the first time i've seen it) There may however be some security repercussions as it may allow bind variables to be passed in via fillin fields.

CreatrixTeam:
One could perhaps check the immediate previous and after characters of ? and act accordingly. If the following assumption holds, this should likely work for most scenarios:
- The ? character is followed and preceded by either a space, a comma, a bracket, an operation, or nothing.

A full solution would require encoding of any inputs, processing of the ? characters, and then decoding.

PR #200, opened 2016-02-17
Patch for - BulkBind is broken in 5.20.X #199

10n:
Bulk bind fix proposal (see #199).

Before checking the number of arguments if  array_2d is detected then we should make sure all list of arguments have the same size.

### DIFF
--- a/adodb.inc.php
+++ b/adodb.inc.php
@@ -1156,11 +1156,11 @@ if (!defined('_ADODB_LAYER')) {
     					);
     					return false;
 				    }
-
+				    unset($countElements);
 				}
 				// Make sure the number of parameters provided in the input
 				// array matches what the query expects
-			    $element0 = reset($inputarr);
+			    	$element0 = reset($inputarr);
 				if ($nparams != count($element0)) {
 					$this->outp_throw(
 						"Input array has " . count($element0) .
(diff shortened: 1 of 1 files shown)


mnewnham:
Thanks, can you give a bit of background, such as what driver you are using and if possible a simple
code snippet to give some context to the fix. I only ask because there has been substantial discussion about counting and checking bind variables #132 #150

10n:
Hello,
(sorry for the late reply)

I'm using MySQL driver. To replicate the issue do an bulk insert with at least 3 columns. 
See attachments also.

```
// Table Structure:
// CREATE TABLE testing ( colA int(11) NOT NULL, colB varchar(20) DEFAULT NULL, colC varchar(12) DEFAULT NULL );

$db = NewADOConnection('mysqli://user:pass@host/tmp');
if (!$db) die("Connection failed"); 
$db->bulkBind = true;
$data = array(
    array('value row 1', date('c'), '1'),
    array('value row 2', date('c'), '2'),
);

$db->Execute("INSERT INTO `testing` (`colA`, `colB`, `colC`) VALUES (?, ?, ?);", $data);
```


[AdodbTest.zip](https://github.com/ADOdb/ADOdb/files/183844/AdodbTest.zip)
[index.zip](https://github.com/ADOdb/ADOdb/files/183841/index.zip)

mnewnham:
Ok, I has time to check this with mysql and sql server and it seems fine. 

@dregad , Its a good few commits behind master but looks like that will not be a problem. Could we pull this please?

PR #207, opened 2016-02-21
Some more specialized casts for _recreate_copy_table as newer PostgreSQL

ralfbecker:
Current PostgreSQL versions wont allow an ALTER COLUMN between varchar and bytea (blob) without an explicit cast. This pull request adds this casts for both directions and fixes sequence name detection to understand newer ::regclass instead of ::text in defaults.

Ralf

### DIFF
--- a/datadict/datadict-postgres.inc.php
+++ b/datadict/datadict-postgres.inc.php
@@ -311,18 +311,27 @@ class ADODB2_postgres extends ADODB_DataDict {
 	{
 		if ($dropflds && !is_array($dropflds)) $dropflds = explode(',',$dropflds);
 		$copyflds = array();
-		foreach($this->MetaColumns($tabname) as $fld) {
+		foreach(($meta=$this->MetaColumns($tabname)) as $fld) {
 			if (!$dropflds || !in_array($fld->name,$dropflds)) {
 				// we need to explicit convert varchar to a number to be able to do an AlterColumn of a char column to a nummeric one
 				if (preg_match('/'.$fld->name.' (I|I2|I4|I8|N|F)/i',$tableflds,$matches) &&
 					in_array($fld->type,array('varchar','char','text','bytea'))) {
 					$copyflds[] = "to_number($fld->name,'S9999999999999D99')";
+				} elseif (preg_match('/'.$fld->name.' ([\w]+)/i',$tableflds,$matches) &&
+					strtoupper($fld->type) != ($type = $this->ActualType($matches[1]))) {
+					if ($type == 'BYTEA' && $fld->type == 'text') {
+						$copyflds[] = "DECODE($fld->name, 'escape')";
+					} elseif ($fld->type == 'bytea' && $type == 'TEXT') {
+						$copyflds[] = "ENCODE($fld->name, 'escape')";
+					} else {
+						$copyflds[] = "CAST($fld->name AS $type)";
+					}
 				} else {
 					$copyflds[] = $fld->name;
 				}
 				// identify the sequence name and the fld its on
 				if ($fld->primary_key && $fld->has_default &&
-					preg_match("/nextval\('([^']+)'::text\)/",$fld->default_value,$matches)) {
+					preg_match("/nextval\('([^']+)'::(text|regclass)\)/",$fld->default_value,$matches)) {
 					$seq_name = $matches[1];
 					$seq_fld = $fld->name;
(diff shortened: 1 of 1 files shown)


datadict/datadict-postgres.inc.php
```diff
@@ -311,18 +311,27 @@ function _recreate_copy_table($tabname,$dropflds,$tableflds,$tableoptions='')
 	{
 		if ($dropflds && !is_array($dropflds)) $dropflds = explode(',',$dropflds);
 		$copyflds = array();
-		foreach($this->MetaColumns($tabname) as $fld) {
+		foreach(($meta=$this->MetaColumns($tabname)) as $fld) {
```
dregad: What is the purpose of this assignment ? `$meta` does not appear to be used anywhere.

datadict/datadict-postgres.inc.php
```diff
@@ -311,18 +311,27 @@ function _recreate_copy_table($tabname,$dropflds,$tableflds,$tableoptions='')
...
 			if (!$dropflds || !in_array($fld->name,$dropflds)) {
 				// we need to explicit convert varchar to a number to be able to do an AlterColumn of a char column to a nummeric one
 				if (preg_match('/'.$fld->name.' (I|I2|I4|I8|N|F)/i',$tableflds,$matches) &&
 					in_array($fld->type,array('varchar','char','text','bytea'))) {
 					$copyflds[] = "to_number($fld->name,'S9999999999999D99')";
+				} elseif (preg_match('/'.$fld->name.' ([\w]+)/i',$tableflds,$matches) &&
```
dregad: I may be nit-picking and over-engineering things, but we're doing nearly the same regex match twice in a row, it would be more efficient to run preg_match() only once and use the capturing pattern `$matches[1]` in the if statements' conditions.

Moreover, in `[\w]+`, the square brackets are redundant - `\w+` is enough.

datadict/datadict-postgres.inc.php
```diff
@@ -311,18 +311,27 @@ function _recreate_copy_table($tabname,$dropflds,$tableflds,$tableoptions='')
...
+					strtoupper($fld->type) != ($type = $this->ActualType($matches[1]))) {
```
dregad: Try to avoid assignments within conditional statements if possible, they are a source of confusion.

datadict/datadict-postgres.inc.php
ralfbecker: Looks like some left over debug code ...

I'll remove it.

datadict/datadict-postgres.inc.php
ralfbecker: Ok, that one is fixed with updated pull request too.

datadict/datadict-postgres.inc.php
ralfbecker: Removed the square brackets.

If you insist I can change it to a single preg_match, but it makes code harder to understand in my opinion.

dregad:
> If you insist I can change it to a single preg_match, but it makes code harder to understand in my opinion.

I don't think so, consider something like this (not tested):

``` php
if (preg_match('/'.$fld->name.' (\w+)/i', $tableflds, $matches)) {
	$new_type = strtoupper($matches[1]);
	if (in_array($new_type, array('I', 'I2', 'I4', 'I8', 'N', 'F')) && 
		in_array($fld->type, array('varchar','char','text','bytea'))
	) {
		$copyflds[] = "to_number($fld->name,'S9999999999999D99')";
	} else {
		$new_actual_type = $this->ActualType($new_type);
		if (strtoupper($fld->type) != $new_actual_type) {
			if ($new_actual_type == 'BYTEA' && $fld->type == 'text') {
				$copyflds[] = "DECODE($fld->name, 'escape')";
			} elseif ($fld->type == 'bytea' && $new_actual_type == 'TEXT') {
				$copyflds[] = "ENCODE($fld->name, 'escape')";
			} else {
				$copyflds[] = "CAST($fld->name AS $new_actual_type)";
			}
		}
	}
} else {
	$copyflds[] = $fld->name;
}
```

ralfbecker:
Misunderstood what you wanted to change, that makes sense and is changed now in pull request.

Ralf

mnewnham:
@dregad , whats the status on this? With no expertise in postgresql, obviously I have no opinion to offer

PR #209, opened 2016-02-21
Always name quote

ralfbecker:
This pull-requests consists of two part:

a) new feature: always quote column and table names, instead of only if they contain special characters.
This allows to use reserved words of the database as column- or table-names eg. "group" or "timestamp".
I know this is a bad idea and we don't use that in our own software, but we have to create schemata for other projects we have no influence on too.

b) bug fix: PostgreSQL datadictionary queries columns and indexes from PostgreSQL's internal tables. If a table-name is quoted (either always because of above patch, or because it contains special chars!) the comparison will fail and no columns or indexes get reported.

As a workaround for a) we could already pass table- or column-name MySQL quoted eg. "`group`", which then get's unconditionally replaced by DB specific quotes in existing nameQuote implementation.
Then we would still run into bug fixed with b), which exists in current code-base independent of a).

Ralf

### DIFF
--- a/adodb-datadict.inc.php
+++ b/adodb-datadict.inc.php
@@ -350,14 +350,13 @@ class ADODB_DataDict {
 			return $quote . $matches[1] . $quote;
 		}
 
-		// if name contains special characters, quote it
-		$regex = ($allowBrackets) ? $this->nameRegexBrackets : $this->nameRegex;
-
-		if ( !preg_match('/^[' . $regex . ']+$/', $name) ) {
-			return $quote . $name . $quote;
+		// allways quote column names to kope with column names being reserved words like eg. "timestamp"
+		// if brackets are allowed, quote only the rest,
+		// to allow to limit indexes on colums, eg "column(32)"
+		if ($allowBrackets && preg_match('/^(.*) *(\(\d+\))$/',$name,$matches)) {
+			return $quote . $matches[1] . $quote . ' '. $matches[2];
 		}
-
-		return $name;
+		return $quote . $name . $quote;
 	}
 
 	function TableName($name)
@@ -665,11 +664,11 @@ class ADODB_DataDict {
 			//-----------------
 			// Parse attributes
 			foreach($fld as $attr => $v) {
-				if ($attr == 2 && is_numeric($v)) 
+				if ($attr == 2 && is_numeric($v))
 					$attr = 'SIZE';
-				elseif ($attr == 2 && strtoupper($ftype) == 'ENUM') 
+				elseif ($attr == 2 && strtoupper($ftype) == 'ENUM')
 					$attr = 'ENUM';
-				else if (is_numeric($attr) && $attr > 1 && !is_numeric($v)) 
+				else if (is_numeric($attr) && $attr > 1 && !is_numeric($v))
 					$attr = strtoupper($v);
 
 				switch($attr) {
@@ -820,7 +819,7 @@ class ADODB_DataDict {
 			if (strlen($fprec)) $ftype .= ",".$fprec;
 			$ftype .= ')';
 		}
-		
+
 		/*
 		* Handle additional options
 		*/
@@ -833,12 +832,12 @@ class ADODB_DataDict {
 					case 'ENUM':
 					$ftype .= '(' . $value . ')';
 					break;
-					
+
 					default:
 				}
 			}
 		}
-		
+
 		return $ftype;
 	}
 
--- a/drivers/adodb-postgres64.inc.php
+++ b/drivers/adodb-postgres64.inc.php
@@ -616,6 +616,11 @@ class ADODB_postgres64 extends ADOConnection{
 	{
 		global $ADODB_FETCH_MODE;
 
+		//if tablenames are quoted, remove the quotes as the tablenames here are used for comparsion of content of fields in postgres system tables
+		if (!empty($this->nameQuote) && !(strpos($table,$this->nameQuote)===false)) {
+			$table = str_replace($this->nameQuote,'',$table);
+		}
+
 		$schema = false;
 		$this->_findschema($table,$schema);
 
--- a/drivers/adodb-postgres64.inc.php
+++ b/drivers/adodb-postgres64.inc.php
@@ -492,6 +492,9 @@ class ADODB_postgres64 extends ADOConnection{
 	{
 		global $ADODB_FETCH_MODE;
 
+		// table-name must NOT be quoted, otherwise we will not find any index
+		$table = str_replace($this->nameQuote,'',$table);
+
 		$schema = false;
 		$false = false;
 		$this->_findschema($table,$schema);


ralfbecker:
Any comment / opinion on that one?

mnewnham:
I agree totally with the principle of always quoting table names, In fact  there is a much deeper issue with IBM DB2 where we use quoting to distinguish between tables named "TABLENAME" and "Tablename" in the the same database. I do have a substantial amount of code related to that issue that I have not yet committed that may conflict with your work, however I have not had time to review your submission. 

You'll have to excuse delays in responding to your updates, the release of the new website and the associated work of updating documentation etc has had an impact on our workload associated with this project. We will get to them all eventually, and I would say that in principle, there is no reason that we would decline to add them to the project, so you can continue to work assuming your submitted features will be added to the project.

ralfbecker:
Just pushed an other bugfix around quoting of table names: datadict-mysql ADODB2_mysql::MetaColumns uses nameQuote on table-name before calling connection->MetaColumns, which uses "SHOW COLUMNS FROM `%s`".
This results in double quoting, if nameQuote quotes the table name. This does not happen as often, if only certain table names get quoted (probably been overseen so far), but it causes the datadict method to fail, if we always quote. 

Cleaner approach would be to move nameQuote into connection and call it in connections MetaColumns, instead of putting the quotes in the used SQL template.

Anyway current code with my fix produces always quoted table names in MetaColumns.

Neustradamus:
@ralfbecker: Have you seen the @mnewnham comment?

Can you do it?

Thanks in advance.

ralfbecker:
@Neustradamus no I have not, sorry

To answer the questions:
The two are closely linked, the bug-fix in PostgreSQL is only required, if you (always) quote names, or if the existing name contains a "special" char for which it was already quoted (and then failed with PostgreSQL!).

Anyway does the PostgreSQL stuff still matters for you, as ADOdb's PostgreSQL implementation for autoincrement depends on OIDs, which PostgreSQL dropped with version 12?
I did not have time to fix that for EGroupware and (the last time I checked) it's not fixed in ADOdb either ...

Ralf

dregad:
@ralfbecker 
> Anyway does the PostgreSQL stuff still matters for you, as ADOdb's PostgreSQL implementation for autoincrement depends on OIDs, which PostgreSQL dropped with version 12

I stopped using PostgreSQL several years ago, and nowadays only fire it up occasionally for dev/test purposes so I'm must admit that I am not keeping up with the changes as I used to.

That being said, I'm not sure why you are saying that we use OID for autoincrement... An autoincrement column would be defined as something like `id I AUTO`, which the postgres datadict will translate to `id SERIAL`. No OID there.

We do still reference *oid* columns in various SQL statements querying the system catalog (e.g. pg_class, pg_type, and others), but those are still officially documented as row identifier in [current (v16) documentation](https://www.postgresql.org/docs/16/catalogs.html).

Maybe you care to clarify if you believe ADOdb is broken (which is what your statement seems to imply - or at least that is how I understand it) and in what way. With your help, I'm sure we can fix it.

ralfbecker:
Hi Damien,
just checked the code and you're right: https://github.com/ADOdb/ADOdb/blob/master/drivers/adodb-postgres8.inc.php#L40

Looking at my PostgreSQL bug reports: problem is that a lot of the meta* queries in ADOdb still use OID's:
https://github.com/ADOdb/ADOdb/blob/master/drivers/adodb-postgres7.inc.php#L46
There are many more, if you grep for "oid" in the drivers/adodb-postgres7.inc.php.

So this part of ADOdb no longer works with PostgreSQL 12+ :(
Unfortunately EGroupware needs these functionality for schema-updates and I have no longer access to PostgreSQL developers to get replacements for the queries used in ADOdb.

But all that is not a contradiction of this pull-request, only of the usefulness of ADOdb in it's current state with recent PostgreSQL versions. 

Ralf

ralfbecker:
Sorry, I read over your statement about the oid columns, but corrected now my post, shortly before you replied :)

Please reread my changed comment and I'm happy if the problem no longer exist in current ADOdb :)

Ralf

ralfbecker:
> Is that a fact ?

I honestly don't know at the moment, maybe all our PostgreSQL problems are related to that [missing fix for metaDefaultSQL](https://github.com/ADOdb/ADOdb/blob/master/drivers/adodb-postgres8.inc.php#L33).

> This` discussion is getting somewhat off-topic... I would appreciate if you could file new Issues for specific compatibility problems you have encountered with PostgreSQL 12+, if possible with steps to reproduce, and we can look into it and hopefully fix them.

Don't get me wrong, I have no expectation of anyone fixing problems or bugs for EGroupware. 

I'm really happy if I can get back to a recent ADOdb version independent if it works with PostgreSQL or not and happy to send pull requests with fixes, if I need something fixed e.g. due to newer PHP versions.

Thanks :)

dregad:
Cross-posting hell :wink:
I read your edited post, and I think we are aligned. 

> * SELECT d.adnum as num, d.adsrc as def from pg_attrdef d, pg_class c where d.adrelid=c.oid and c.relname='egw_accounts' order by d.adnum --> Database error
>       --> that one seems to be fixed in your ADOdb version :)
>       https://github.com/ADOdb/ADOdb/blob/master/drivers/adodb-postgres8.inc.php#L33

Yes, I fixed this a few years ago, see #562. Not related to OID though.

> * need to check, that all my [PHP version related fixes](https://github.com/EGroupware/ADOdb/commits/master/) are also in place, as we run PHP 8.3

We have fixed a lot of [PHP 8.x compatibility issues](https://github.com/ADOdb/ADOdb/issues?q=is%3Aissue+label%3APHP8%2CPHP8.1%2CPHP8.2%2CPHP8.3%2CPHP8.4) in the past few years, and to my knowledge there are none remaining except #1036 (but that's 8.4). I'm not running 8.3 in production though, only 8.2, so f you do find any remaining issues, just open a ticket and we'll take care of it.

> But all that is not a contradiction of this pull-request ...

Indeed 😃 

For the record, regarding removal of OIDs in v12+ - after a bit of research my understanding is that they actually removed the `WITH OID` table attribute, and got rid of the special handling for this column (which was hidden). I believe our usage of the oid column in meta queries is fine, according to [this(https://www.postgresql.org/message-id/E1gPG3j-0003cy-03%40gemulon.postgresql.org):

*Remove WITH OIDS support, change oid catalog column visibility.
[...]
The biggest user of WITH OID columns was postgres' catalog. This
commit **changes all 'magic' oid columns to be columns that are normally
declared and stored**. To reduce unnecessary query breakage all the
newly added columns are still named 'oid' [...]*

ralfbecker:
Hallo Damien,

I was able to get EGroupware working on ADOdb master under MariaDB/MySQL with the following 6 patches:
https://github.com/EGroupware/ADOdb/commits/egroupware-new/

Some of them have been send as pull requests back in 2016, others are later additions.

Can you please have a quick look, if it makes sense from opinion to send these as pull requests?

I would like to do that first, before testing and looking at the bigger amount of PostgreSQL specific commits.

What do you think?

Regards 

Ralf

dregad:
Thanks for the feedback. I'll have a look and get back to you.

Neustradamus:
@ralfbecker: Nice, good work! :)

Have you other changes in your fork which can be added in upstream?

@dregad: Hope that you can quickly look it.

cc: @peterdd.

ralfbecker:
Yes, I have more PostgreSQL related changes, but I want to keep it in manageable pieces.

I started with the minimum we (EGroupware) need to work on MariaDB or MySQL.

Ralf

Neustradamus:
@dregad: Have you looked since 3 months?

I think that @ralfbecker waits you...

ralfbecker:
Not sure what you asking?

EGroupware is now using the current ADOdb, with the changes I submitted as pull requests to you.

The exact version we use, is always available in our [forked repos `egroupware-new` branch](https://github.com/EGroupware/ADOdb/tree/egroupware-new) and via [packegist.org egroupware/adodb-php packages](https://packagist.org/packages/egroupware/adodb-php#5.22.7.3)

If there's anything I can do, please let me know.

Ralf

Neustradamus:
@ralfbecker: Thanks for your quick answer! :)
I would like that this PR and all tickets linked to will be solved.
The official ADOdb has not the fix.

@dregad must solve it upstream...

Neustradamus:
@dregad: Can you look?
We are in 2025 now.

@ralfbecker: All your improvements exist in PRs?

PR #210, opened 2016-02-22
PostgreSQL >= 9.0 gives an error when droping a table with auto-increment

ralfbecker:
Sequences are not detected as dependent on the table and therefore give an error when trying to drop them, after table was dropped.
They are reported as deptype "a"=automatic and "n"=normal, but sequence is droped automatic as for "i"=internal.
Fortunate 9.0+ supports DROP SEQUENCE IF EXITS, so we can use that to not get an error in that case.

### DIFF
--- a/datadict/datadict-postgres.inc.php
+++ b/datadict/datadict-postgres.inc.php
@@ -287,8 +287,8 @@ class ADODB2_postgres extends ADODB_DataDict {
 		$has_drop_column = 7.3 <= (float) @$this->serverInfo['version'];
 		if (!$has_drop_column && !$tableflds) {
 			if ($this->debug) ADOConnection::outp("DropColumnSQL needs complete table-definiton for PostgreSQL < 7.3");
-		return array();
-	}
+			return array();
+		}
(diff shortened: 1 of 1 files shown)


ralfbecker:
Don't pull that right now, there is an other problem in that area, which I want to fix first.

ADOdb code to find the index is a bit simplistic, it assumes index name contain table name, but does not check index is from the concerned table. I have eg. two tables "prefix_name" and "prefix_name_extra", current code will try to drop an index of "prefix_name_extra" when asked to drop it from "prefix_name".

That's easy to fix by checking the table the index is from.

dregad:
OK.

dregad:
@ralfbecker any update on this ? Are you still planning to provide a fix ?

dregad:
@ralfbecker ping

mnewnham:
@ralfbecker , if you have abandoned this, I will close it. You can reopen it if you would like to re-start work on it

mnewnham:
As there has been no response on this, I am closing off the request. Feel free to re-activate

PR #215, opened 2016-03-09
fix maria year values to not default to Y-m-d, and stay YYYY

matt-land:
This alters year column behavior, to stop it from being force formatted into Y-m-d. On maria, the engine issues an "Out of range value for column '<column>' at row <row>" error, and casts the value to "0000"

### DIFF
--- a/adodb-datadict.inc.php
+++ b/adodb-datadict.inc.php
@@ -250,7 +250,7 @@ class ADODB_DataDict {
 		'LONGBINARY' => 'B',
 		'B' => 'B',
 		##
-		'YEAR' => 'D', // mysql
+		'YEAR' => 'Y', // mysql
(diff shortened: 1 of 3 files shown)


mnewnham:
I'm a little confused here. I'm assuming that for mariadb, you are using the mysqli ADOdb class.

The method ``metatype()`` is extended in both ``drivers/adodb-mysqli.inc.php`` and ``datadict/datadict.mysql.inc.php``, and it is those methods that control the returned metatype, not the base classes.

In addition, is this a behavior where mariadb differs from mysql?

dregad:
@matt-land you have not provided any feedback, so I'm going to close this PR without merging it. Feel free to reopen if you feel this is still an issue.

Issue #232, opened 2016-04-24
sqlDate does not return correct day of week with Microsoft SQL Server


mnewnham:
The code to return the 'l' format in sqlDate is missing from the procedure in drivers/adodb-mssqlnative.inc.php

PR #233, opened 2016-04-28
Driver does not support 'l' format in sqlDate

mnewnham:
The code for the 'l' format (day of week) is missing from the sqlDate function

see #232


dregad:
No problems with the change, but I'm curious as to why you're targeting it at 5.20.5 while all the other sqlDate-related fixes were merged in master (5.21)

dregad:
Removed labels and milestone as this is already referenced under #232

dregad:
Fixed in 5.21 (see f0d060dbe0ce029d7800d97394da7fd66d23370e)

Issue #236, opened 2016-05-03
_adodb_getcount fails in v5.20


superhigi:
_adodb_getcount used to work with v5.18 for the following queries:
SELECT &#x60;event_28&#x60;.* FROM &#x60;event&#x60; &#x60;event_28&#x60; ORDER BY (datestart) DESC
SELECT &#x60;eventreservation_35&#x60;.* FROM &#x60;eventreservation&#x60; &#x60;eventreservation_35&#x60; ORDER BY (SELECT CONCAT((datestart)) FROM event WHERE event.id=eventreservation_35.id_event) DESC, &#x60;ts&#x60; DESC

RegExp was: `/^\s*SELECT\s.*\s+FROM\s/Uis`

The new RegExp `/^\s*?SELECT\s.*?\s+(.*?)\s+FROM\s/is` creates faulty SQL.

ipso:
This may have been caused by a commit I sent in a while back, I will dig into it more later today.

ipso:
Yeah, this REGEX was changed in Pull Request #88 that I submitted a year ago.

I'm not sure there is much we can do here, ADODB will never be able to handle much more than the most basic types of queries. Once you start adding unions/sub-selects/JOINs/etc... there will always be substantial number of failure cases. 

The REGEX was originally changed a year ago (the code for which was posted to the PHPLENS forums in 2012) to try and catch a few more cases of more complex queries as it was low hanging fruit at the time, however I neglected to include examples of those queries then, but I think it was related to JOINs.

A bit more checking, it appears the old REGEX fails for a simple query like this:
`select col1,col2,col3,col4,
from table1
LEFT JOIN ( select col5 from table3 where col3=col4 ) as tmp;
`

But the new REGEX does work on it. So reverting back to the old REGEX isn't something I would recommend, as it would break other likely more used cases.

Anyways, to be honest, I'm not even sure I understand what your query is attempting to do, it seems to be like the previous count(*) behavior may have been valid SQL but also may return ambiguous results.

What the above mentioned pull request did add though was support to manually handle even the most complex queries, and that is by adding the _ADODB_COUNT keyword to the query, ie:

`SELECT _ADODB_COUNT `event_28`.* _ADODB_COUNT  FROM `event` `event_28` ORDER BY (datestart) DESC
[...]

PR #237, opened 2016-05-04
Complex query pagination fix

ipso:
Reverts regular expression back to its original as it seems to work in more cases, and doesn't break simple cases. Since we support _ADODB_COUNT now anyways we should be able to handle any possible situation. 

See Issue #236

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -5,4 +5,5 @@
 *.tmp
 .project
 .settings/
+/nbproject/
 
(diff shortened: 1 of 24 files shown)


dregad:
I didn't look in detail as I'm on my cell phone buzz there seems to be an awful lot of commits for a simple revert... Maybe you should rebase your branch?

ipso:
Will try again...

PR #238, opened 2016-05-05
Query pagination optimizatio nc

ipso:
REBASED from upstream master...

Reverts regular expression back to its original as it seems to work in more cases, and doesn't break simple cases. Since we support _ADODB_COUNT now anyways we should be able to handle any possible situation.

See Issue #236


ipso:
Grr... The preview shows one line changed and a single commit, yet when click to submit the pull request it changes to to show a bunch of other commits...

dregad:
Have a look at https://github.com/ipso/ADOdb/network.

The problem is due to your keeping your local *master* branch in sync with ours by means of merge commits. As a result, the code is the same, but the commits history to get to it is different.

I recommend that you always use rebase (or fast-forward merge) to keep your local *master* aligned with upstream, or alternatively create your feature branches off of upstream/master.

ipso:
Ya, I think I figured it out finally, see PR 239

dregad:
I saw that - just thought I'd give you an explanation as to why you were experiencing the behavior and how to avoid it in the future.

PR #241, opened 2016-05-11
Calling ADONewConnection with pdo_* causes a fatal error

gohanman:
If you pass `pdo_mssql`, `pdo_mysql`, etc to `ADONewConnection` there's a fatal error that occurs when `ADOLoadCode` includes the subclass file because `ADODB_pdo` is undefined.

Maybe those `pdo_*` strings aren't valid arguments to the function but the @ operator on the include that crashes makes it a bit of a nasty gotcha.

### DIFF
--- a/adodb.inc.php
+++ b/adodb.inc.php
@@ -4874,6 +4874,15 @@ http://www.stanford.edu/dept/itss/docs/oracle/10g/server.101/b10759/statements_1
 				$class = $db = 'postgres8';
 				break;
 
+			case 'pdo_mssql':
+			case 'pdo_mysql':
+			case 'pdo_oci':
+			case 'pdo_pgsql':
+			case 'pdo_sqlite':
+			case 'pdo_sqlsrv':
+				$class = $db = 'pdo';
+				break;
+
 			default:
 				$class = $db; break;
 		}
--- a/adodb.inc.php
+++ b/adodb.inc.php
@@ -4874,17 +4874,13 @@ http://www.stanford.edu/dept/itss/docs/oracle/10g/server.101/b10759/statements_1
 				$class = $db = 'postgres8';
 				break;
 
-			case 'pdo_mssql':
-			case 'pdo_mysql':
-			case 'pdo_oci':
-			case 'pdo_pgsql':
-			case 'pdo_sqlite':
-			case 'pdo_sqlsrv':
-				$class = $db = 'pdo';
-				break;
-
 			default:
 				$class = $db; break;
+		}
+		
+		if (substr($db, 0, 4) === 'pdo_') {
+			ADOConnection::outp("Invalid database type $db");
+			return false;
 		}
 
 		$file = ADODB_DIR."/drivers/adodb-".$db.".inc.php";


mnewnham:
The correct way to load the PDO driver for a specific databases is:
````
$db = new ADOconnection('pdo');

$db->connect('<databaseType>:<dsn options>');

````
See [The ADOdb PDO driver documentation](http://adodb.org/dokuwiki/doku.php?id=v5:database:pdo). When used this way, the correct sub type of the driver is loaded automatcally

**Comment by gohanman** [comment]
Nevertheless having magic strings that cause a suppressed fatal error seems problematic. If the argument shouldn't be adjusted to `pdo` then it should check for these values and reject them with a visible error. Dying in complete silence is the worst possible behavior.

**Comment by mnewnham** [comment]
If you'd like to help by submitting some patches for the driver so that the behavior is a little more graceful, it would be much appreciated.

**Comment by gohanman** [comment]
I've updated the PR so those arguments are rejected with an error rather than converted to `pdo`.

**Comment by mnewnham** [comment]
Thanks, @dregad , could we pull this into 5.21? I'm not sure it needs to go into a hotfix

**Comment by dregad** [comment]
That's an ugly hack over an even uglier piece of code... I think the way ADOdb loads its classes is really outdated and deserves to be refactored. 

That being said, I don't think it's worth the effort to do this now, considering we're planning to switch to PSR-2, at which point we'll be able to rely on standard autoloading mechanisms.

Another way to fix this error, would be to include the *ADODB_pdo* base class in the subclasses

**Comment by gohanman** [comment]
> Another way to fix this error, would be to include the ADODB_pdo base class in the subclasses

For what it's worth, this is what I was doing originally. Calling the constructor on for example *ADODB_pdo_mysql* instead of *ADODB_pdo* creates some divergences in behavior. The way the base class wraps a non-inherited inner driver object means changing which class constructor is called can change the way a given method call interacts with the outer vs inner object.

In other words `ADONewConnection('pdo')` and `ADONewConnection('pdo_mysql')` will both be able to connect successfully but will have edge cases where the resulting objects' behavior is not identical. That seems like it could create additional headaches down the road.

**Comment by dregad** [comment]
OK I guess my tests didn't hit any of these edge cases then. Thanks for the feedback, and your contribution too of course.

**Comment by gohanman** [comment]
#248 and #242 are both to resolve this kind of issue. The latter seems more likely to occur in practice.

```php
class ADODB_pdo
{
   function foo()
   {
       return $this->_driver->foo();
   }

   function bar()
   {
       return false;
   }
}

class ADODB_pdo_mysql extends ADODB_pdo
{
    function foo()
    {
        return true;
    }

    function bar()
    {
        return true;
    }
}
```

*foo* is the same regardless of which constructor you call; *bar* isn't. The result of overriding methods can change when the outer class changes. 

Also, what `$this->_driver` means can vary depending whether you're calling a method on the outer or inner object. Subclasses probably shouldn't reference this property directly but it's easy to hit accidentally by calling an inherited method. Whether or not that property is initialized may vary depending whether the subclass is currently the inner or outer object.

PR #246, opened 2016-05-17
Support for T-SQL styled wrapping column names in square brackets

vanilla-thunder:
Microsoft wraps column names in square brackets in their t-sql statements, which is not really obligatory but it is necessary when columns have names same as reserved keywords (e.g. 'BEGIN' or 'CURRENT').

While it is still possible to manually wrap these column names with square brackets and use them in Execute() statement, the AutoExecute() will ignore these columns and may cause bad data or sql errors.

i can provide you some example code for better explanation and reproducing this problem, if you need.

### DIFF
--- a/adodb-lib.inc.php
+++ b/adodb-lib.inc.php
@@ -861,6 +861,8 @@ static $cacheCols;
 			$bad = false;
 			if ((strpos($upperfname,' ') !== false) || ($ADODB_QUOTE_FIELDNAMES)) {
 				switch ($ADODB_QUOTE_FIELDNAMES) {
+				case 'BRACKETS':
+					$fnameq = $zthis->leftBracket.$upperfname.$zthis->rightBracket;break;
 				case 'LOWER':
 					$fnameq = $zthis->nameQuote.strtolower($field->name).$zthis->nameQuote;break;
 				case 'NATIVE':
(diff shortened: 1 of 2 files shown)


mnewnham:
I think you also need to add the support into the _adodb_getupdatesql() function, also, is the same effect not achieved using the standard backtick?

vanilla-thunder:
You are absolutely right about the _adodb_getupdatesql(), i added a new commit
With my code backtick would not work with AutoExecute, too, because it fetches the column names from the table and checks passed values array against them:
https://github.com/ADOdb/ADOdb/blob/master/adodb-lib.inc.php#L860

E.g. there are 3 columns: column1 | begin | column3
this would cause an sql error like "wrong syntax next to begin":
```php
$values = [
'column1' => 'value1',
'begin' => 'value2',
'column3' => 'value3'
];
$db->AutoExecute('TABLE',$values);
```
this code will ignore begin column completely because $values has no '[begin]' key:
```php
$values = [
'column1' => 'value1',
'[begin]' => 'value2',
'column3' => 'value3'
];
$db->AutoExecute('TABLE',$values);
```
i got same results when using backtick 
```php
'`begin`' => 'value2',
```

but maybe you had another code in mind? I'm not very familiar with adodb, i'm just using it for some mssql migration scripts.

mnewnham:
If you are doing bulk migrations, I would definitely counsel against using autoExecute see the warning at the top of [This page](http://adodb.org/dokuwiki/doku.php?id=v5:reference:connection:autoexecute)

vanilla-thunder:
I have noticed the impact on performance compared to regular execute() functions, but the advantages of AutoExecute() prevail the slower performance.
Actually this column-array_keys matching and simple possibility to chose between insert or update is exactly what i need for transferring and synchronising data

mnewnham:
OK, just so you are aware

PR #248, opened 2016-05-19
Eliminate notice that occurs when any ADO_pdo subclass calls Execute()

gohanman:
Call chain is:
1. ADOConnection::Execute()
2. ADOConnection::_Execute()
3. ADO_pdo::_query()

Since $this is an ADO_pdo subclass rather than an ADO_pdo and $_driver isn't actually an inherited property of the parent class, $this->_driver doesn't exist.


dregad:

PR #259, opened 2016-07-12
oci8po: prevent segfault on php7

damyon:
The OCIFetchinto function is causing segfaults on php7 - probably because the fields array
is not initialised or it is optimised out. This fixes just changes to use the safer function
oci_fetch_array instead.

The attached script will trigger the segfault.

[test.php.txt](https://github.com/ADOdb/ADOdb/files/358900/test.php.txt)


dregad:
Thanks for the report. I don't have a working Oracle setup I can use to confirm the issue and test your patch at the moment and no time to set one up, so I'll trust your tests and apply the change as-is.

PR #267, opened 2016-08-10
xml: support table 'opt' attribute with mysqli

peterdd:
This is a quickfix just to make it work without impact on other things.
Currently the opt tag is just ignored for mysqli!

### DIFF
--- a/adodb-xmlschema03.inc.php
+++ b/adodb-xmlschema03.inc.php
@@ -505,7 +505,11 @@ class dbTable extends dbObject {
 	*/
 	function addTableOpt( $opt ) {
 		if(isset($this->currentPlatform)) {
-			$this->opts[$this->parent->db->databaseType] = $opt;
+                        if('mysqli' === $this->parent->db->databaseType){
+                                $this->opts['mysql'] = $opt;
+                        }else{
+                                $this->opts[$this->parent->db->databaseType] = $opt;
+                        }
 		}
 		return $this->opts;
 	}


dregad:
This feels like a hack... I'm wondering (keep in mind that I don't really use XML schema) if it wouldn't be more appropriate to use `$this->parent->db->dataProvider` instead ?

mnewnham:
I agree, My own experience of having to migrate from mysqlt to mysqli in xmlschema files makes me believe that the dataprovider should always be used for xmlschema internals, so it should be:

````php
 $this->opts[$this->parent->db->dataProvider] = $opt;
````

dregad:
Reopening this, following https://github.com/ADOdb/ADOdb/commit/27bcd7d09f25850ac6e77847a32f04e0012f4671#commitcomment-19029593.

@peterdd 
> switching to 'dataProvider' would probably break other db drivers usage in xmlschema03 platform attributes

Are you recommending that I revert 27bcd7d09f25850ac6e77847a32f04e0012f4671 ? As mentioned previously I don't use XML schema much, so I just went with @mnewnham's reply, since you didn't object to it.

mnewnham:
Point taken about backward compatibility. Perhaps something where we could match either the databaseType or dataProvider might work. 

I personally believe that moving forward, we should have a dataprovider option and the users should be encouraged to use that (to avoid the issues that I myself encountered). 

Just looking at the code in addTableOpt
````php
if(isset($this->currentPlatform)) {
	$this->opts[$this->parent->db->databaseType] = $opt;
}
````
This looks problematic because $this->currentPlatform is defined as a class variable, therefore it always exists. It is however set true/false by the constraint tag (but only this). This feels almost like incomplete code that needs deeper investigation.

dregad:
@peterdd you have never replied to https://github.com/ADOdb/ADOdb/pull/267#issuecomment-247263372. 

As it stands, commit 27bcd7d09f25850ac6e77847a32f04e0012f4671 is part of master branch, should it be reverted or not ? It's worth noting that noone reported any issues in over 3 years, following the switch to use of _dataProvider_ instead of _databaseType_.

What do I do with the PR ? Merge this, cancel or postpone to later release ?

peterdd:
Do I get time until next weekend?

dregad:
I was hoping to finish this week-end, but considering that 5.21 has been pending for about 3 years, a week is not going to make much difference :wink:

dregad:
> Do I get time until next weekend?

@peterdd ping

peterdd:
Sorry, got sidetracked by testing more basic xmlschema03 and session2 stuff (got sucked into the rabbithole) and RL.

I try, but I better do set no due date when focus on this. :-/

dregad:
> got sucked into the rabbithole) and RL

I know what it's like, trust me ;-)

I am fine if you don't have time to fully analyze and provide a proper solution, we can delay that to some later date. For now, I just need to know what I should do with 5.21.0

1. revert 27bcd7d09f25850ac6e77847a32f04e0012f4671, i.e. the "safe" option (i.e. no support for opt attribute on mysqli in 5.21.0
2. keep the change in and release as-is, i.e. we accept the risk that other drivers could be broken (as you explained in https://github.com/ADOdb/ADOdb/commit/27bcd7d09f25850ac6e77847a32f04e0012f4671#commitcomment-19029593)

The risk is quite low as it only affects XML feature, which is not so widely used.

Since you're the original requestor of the feature, and also the one who's worried about the risk, I guess it's your call, just tell me what you want me to do.

dregad:
@peterdd without feedback from you, I'll go with option 2, i.e. keep the change in and release as-is. Please respond by Sunday evening.

dregad:
@peterdd since 5.21.0 has been released and you never followed up with this discussion, I'm going to close the PR. Feel free to reopen it if needed.

peterdd:
Not forgotten, shift to 5.22+

Issue #269, opened 2016-08-15
Undefined constant error when old ADOdb 5.04 Extension is loaded


peterdd:
I got a bug report on https://bugs.flyspray.org/2189 and trying to analyze that seems to be related to ADOdb:

```php
<?php print_r(get_defined_constants(true)); ?>
...
 [ADOdb] => Array
        (
            [ADODB_EXTENSION] => 5.04
...
```

in adodb.inc.php line 91:
```php
 if (!$ADODB_EXTENSION || ADODB_EXTENSION < 4.0) {
...
 line 116: 
define('ADODB_ASSOC_CASE_LOWER', 0);
define('ADODB_ASSOC_CASE_UPPER', 1);
define('ADODB_ASSOC_CASE_NATIVE', 2);
```

That constants seems to be introduced with this commit 4 Mar 2013 5.18/5.19: 
https://github.com/ADOdb/ADOdb/commit/e3c11ed98970408eae649f929898983c6b5d2a6a

I'm now a bit unsure how to solve this in a general and best future proof way..
I have hackish workarounds in my mind, but maybe you know a solid solution?

If you want test, all-inkl.com provides a free 1-week test account. They provide also mantisbt as "one click install", but their install seems to use ADOdb5.10, so I didn't get that undefined constant ADODB_ASSOC_CASE_NATIVE Warning there.

dregad:
The ADOdb Extension is no longer supported, and hasn't been updated for many years (to my knowledge, 5.04 was the last released version).

As a permanent fix, we should remove all references to its use from the code, so that the library relies exclusively on its own PHP objects.

Until then, as a workaround I suggest to work with the hosting provider so they discontinue use of obsolete software (i.e. disable ADOdb Extension from their PHP config). 

You may also want to try manually disabling the extension by modifying adodb.inc.php:
```
diff --git a/adodb.inc.php b/adodb.inc.php
index 821ab2c..0d8d079 100644
--- a/adodb.inc.php
+++ b/adodb.inc.php
@@ -70,7 +70,7 @@ if (!defined('_ADODB_LAYER')) {
    // GLOBAL SETUP
    //==============================================================================================
 
-   $ADODB_EXTENSION = defined('ADODB_EXTENSION');
+   $ADODB_EXTENSION = false;
 
    // ********************************************************
    // Controls $ADODB_FORCE_TYPE mode. Default is ADODB_FORCE_VALUE (3).
```

dregad:
[...]

PR #271, opened 2016-08-17
Exit with error/exception when the ADOdb Extension is loaded

dregad:
@peterdd, as discussed in #269.

Please test and let me know.


dregad:
Merged in 991a6aa3d0efbbbdf2658f5e990047531b3aa254

PR #275, opened 2016-09-03
FIX: ADODB_pdo_mysql::MetaTables fails w/ default args

gohanman:
The generated query is broken with the default method arguments. It issues this query:
```sql
SELECT
            TABLE_NAME,
            CASE WHEN TABLE_TYPE = 'VIEW' THEN 'V' ELSE 'T' END
        FROM INFORMATION_SCHEMA.TABLES
        WHERE TABLE_SCHEMA=
```
which of course fails.

The fix is taken from `ADODB_mysqli::MetaTables`.

### DIFF
--- a/drivers/adodb-pdo_mysql.inc.php
+++ b/drivers/adodb-pdo_mysql.inc.php
@@ -70,7 +70,7 @@ class ADODB_pdo_mysql extends ADODB_pdo {
 	{
 		$save = $this->metaTablesSQL;
 		if ($showSchema && is_string($showSchema)) {
-			$this->metaTablesSQL .= " from $showSchema";
+			$this->metaTablesSQL .= $this->qstr($showSchema);
 		} else {
 			$this->metaTablesSQL .= "schema()";
 		}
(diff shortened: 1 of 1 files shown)


dregad:
This is a regression introduced in v5.20.0 by 57399555f4dd991d6e010f0710fa91b3ec2d77a8 which changed the metaTableSQL but failed to update the MetaTables() function accordingly.

The proposed fix is in fact incomplete, as it still generates invalid SQL when the schema name is specified, e.g.
```
$db->metaTables(false, 'myschema');
```
generates
```
SELECT TABLE_NAME, CASE WHEN TABLE_TYPE = 'VIEW' THEN 'V' ELSE 'T' END 
FROM INFORMATION_SCHEMA.TABLES 
WHERE TABLE_SCHEMA= from myschema
                    ^
                    should be 'myschema'
```

gohanman:
I believe that's the other half of the fix then, also lifted from `ADODB_mysqli`.

dregad:
@gohanman thanks for the follow-up. I had already fixed it (bfb32f99) but forgot to push the branch and update the PR.

Your original commit is fixed in  0f17de19

Issue #270, opened 2016-08-16
Remove references to ADOdb Extension


dregad:
The ADOdb Extension's development stopped at version 5.04, and is no longer maintained or supported, yet the library still contains code referencing it.  If those code branches are executed, ADOdb will not perform properly, due to missing constant declarations (among other things).

We should remove all references to the Extension to avoid any problems (see #269).

dregad:
Proposed patch to remove the Extension code, see PR #276

PR #276, opened 2016-09-14
Remove references to ADOdb Extension

dregad:
For review. See #269, #270 .


dregad:
@mnewnham any comments ?
@peterdd if you have the opportunity to test this with the provider that was causing issues for your Flyspray user, it would be appreciated.

peterdd:
I will try, but prob. no time before october.

mnewnham:
We definitely do need to remove these references, in some sense I think the whole thing was a distraction from the core product. @dregad , I believe your approach to this is fine

dregad:
Thanks for the feedback. I will merge this in as-is, and if @peterdd later realizes this is causing issues, then we can make the necessary adjustments.

PR #278, opened 2016-09-20
ADORecordSet contructor takes only one parameter

mezaof:
ADORecordSet class contructor only takes $queryID parameter, but ADORecordSet_ado class sends two paremeter on constructor.


dregad:
There are several occurences of this behavior; I'd suggest you fix all of them while you're at it.

PR #284, opened 2016-11-02
Create adodb-oc.inc.php

Cedric31:
Please create occitan file.
Regards


Cedric31:
Please add occitan file

dregad:
Follow-up in #285.

@Cedric31 for next time, there is no need to create a separate pull request, just add new commits to branch you used to submit the original one.

PR #288, opened 2016-11-08
Inconsistent variable binding in SelectLimit and oci8po driver

pat-bateman:
Method SelectLimit in oci8po driver handles bind variables inconsistenly:

for $nrow < 1000: $inputarr can be indexed or associative array
for $nrow >= 1000: $inputarr must be indexed array

I tracked it down to line 777 of adodb-oci8.inc.php:

```php
oci_bind_by_name($stmt,":$k",$inputarr[$k],$len);
```

In this case ":$k" does not work with ? placeholder.

Proposed solution: for oci8po driver, always use index key values.

### DIFF
--- a/drivers/adodb-oci8.inc.php
+++ b/drivers/adodb-oci8.inc.php
@@ -755,10 +755,11 @@ END;
 
 			if (is_array($inputarr)) {
 				foreach($inputarr as $k => $v) {
+					$i=0;
 					if ($this->databaseType == 'oci8po') {
-						$bv_name = ":".array_search($k, array_keys($inputarr));
+						$bv_name = ":".$i++;
 					} else {
-						$bv_name = ":$k";
+						$bv_name = ":".$k;
 					}
 					if (is_array($v)) {
 						// suggested by g.giunta@libero.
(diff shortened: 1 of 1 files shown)


drivers/adodb-oci8.inc.php
```diff
@@ -755,13 +755,18 @@ function SelectLimit($sql,$nrows=-1,$offset=-1, $inputarr=false,$secs2cache=0)
 
 			if (is_array($inputarr)) {
 				foreach($inputarr as $k => $v) {
+					if ($this->databaseType == 'oci8po') {
+						$bv_name = ":".array_search($k, array_keys($inputarr));
```
dregad: To improve performance, since we already loop through $inputarr, I suggest to use a counter variable instead of calling array_search/array_keys in each iteration, i.e.
``` php
$i = 0;
foreach($inputarr as $k => $v) {
    if ($this->databaseType == 'oci8po') {
       $bv_name = ":" . $i++;
```

pat-bateman:
Agreed, as for ? placeholders in oci8po driver bind variables necessarily need to be in the right order.

pat-bateman:
Fixed it. Everything ok?

dregad:
Fixed in 0f490242e05e6a8e6aaa5291e98c9296d76b7218. Thanks !

PR #297, opened 2016-11-26
mssql: fix drop/alter column with existing default constraint

### LINKED ISSUE #290: mssqlnative: drop/alter column not working if column has default constraint
With MSSQL it is not possible to drop or alter a column with an existing constraint.
This constraint has to be removed before the drop or alter takes place. 
In 'datadict-mssqlnative.inc.php' AlterColumnSQL is commented out and DropColumnSQL
doesn't care.

The attached patch tries to remedy this with the smallest possible impact.

 - Fix DropColumnSQL, to allow the drop even with a constraint on that given column.
   We drop so any default doesn't matter.
   
 - Fix AlterColumnSql, to allow changes if and only if(!) either a 'new' default is
   given for an existing default, or there is no existing one at all. So something
   like 'ALTER TABLE t ALTER COLUMN c INT NOT NULL' with an existing constraint on c
   will still fail. I can't decide, if keep or remove the constraint is implied here.

[datadict-mssqlnative.inc.php.patch.txt](https://github.com/ADOdb/ADOdb/files/585675/datadict-mssqlnative.inc.php.patch.txt)

obmsch:
With MSSQL it is not possible to drop or alter a column with an existing default constraint. This constraint has to be removed before the drop or alter takes place.

This patch removes an existing default constraint before the drop/alter column is
filed and in case of alter creates a new one. after that, if a 'DEFAULT' is given.

Fixes #290

### DIFF
--- a/datadict/datadict-mssqlnative.inc.php
+++ b/datadict/datadict-mssqlnative.inc.php
@@ -148,19 +148,63 @@ class ADODB2_mssqlnative extends ADODB_DataDict {
 		return $sql;
 	}
 
-	/*
-	function AlterColumnSQL($tabname, $flds, $tableflds='', $tableoptions='')
+	function DefaultConstraintname($tabname, $colname)
+	{
+		$constraintname = false;
+		$rs = $this->connection->Execute( "select name from sys.default_constraints WHERE object_name(parent_object_id) = '" . $tabname ."' AND col_name(parent_object_id, parent_column_id) = '" . $colname . "'");
+		if ( is_object($rs) ) {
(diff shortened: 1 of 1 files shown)


mnewnham:
@obmsch , the code you're submitting in datadict-mssqlnative.inc.php is based on version 5.20.4, which is incompatible with both 5.21 and later releases of 5.20. It doesn't match the version of your own branch obmsch/mssql-issue-290 either. 

Could you rebase your fixes please.

Thanks

obmsch:
Done.

Issue #300, opened 2016-12-09
Issue on Sequence managment on mssqlnative


laurent-n:
Hello,

on adodb-mssqlnative.inc.php Method GenID do not transfert the parameter to GenID2008 and GenID2012 which can receive it.
It cause only one sequence is handled and forcely named 'adodbseq'
Here the existing code.
```function GenID($seq='adodbseq',$start=1) {
		if (!$this->mssql_version)
			$this->ServerVersion();
		switch($this->mssql_version){
		case 9:
		case 10:
			return $this->GenID2008();
			break;
		case 11:
		case 12:
			return $this->GenID2012();
			break;
		}
	}
```

mnewnham:
What version of SQL Server are you using?

laurent-n:
Hello,
Issue is on SQL Server 2012, but on 2005/2008 it was buggy too.
I've submitted a patch to fix them.

PR #301, opened 2016-12-11
Proposal for Issue on Sequence managment on mssqlnative #300

laurent-n:
Here a proposal to fix  Issue on Sequence managment on mssqlnative #300 
Also fixed a typo on line 232
A bug on line 301 it was returning 'true' instead of the next value on the sequence
Also fixed parameter forwarding on CreateSequence

Also initiated a file on test directory to make test using mssqlnative driver


dregad:
Thanks for the patch. Will review and merge as time allows.

dregad:
I'm going to close your PR.

- The individual bug fixes will be included in 5.20.8 release as separate commits (will push them shortly)
- I'll add the test script to 5.21 release (master branch) later on

PR #311, opened 2017-01-16
ensure `$nrows` and `$offset` parameters of `SelectLimit` are integers

nicoder:
as is done (at least for `$nrows`) in the `SelectLimit` method of `Adodb/adodb.inc.php`.

lowers the risk of SQL injection.


dregad:
Sorry I messed up while rewording the commit message and inadvertantly merged instead of squash. The actual fix is 7d43989dc95fd00fcf3ba73e08fdaf56ffa4f3e5

jedi58:
Doesn't this need doing for `mysqli` and the other drivers too?

PR #314, opened 2017-01-17
MSSQL driver: support alternative port in connect

merijnvdk:
Fix actually using the alternative port if it was specified. The base
driver class will strip the port from hostname and put it in port
property. So now add it again if it was set.

### DIFF
--- a/drivers/adodb-mssql.inc.php
+++ b/drivers/adodb-mssql.inc.php
@@ -620,6 +620,7 @@ order by constraint_name, referenced_table_name, keyno";
 	function _connect($argHostname, $argUsername, $argPassword, $argDatabasename,$newconnect=false)
 	{
 		if (!function_exists('mssql_pconnect')) return null;
+		if (!empty($this->port)) $argHostname .= ":".$this->port;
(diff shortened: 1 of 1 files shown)


merijnvdk:
As this is a pretty straightforward and small fix, is there a reason it has not been looked at? Something more I need to do maybe?

dregad:
> is there a reason it has not been looked at? 

Sorry for the delay, but please keep in mind that we maintain ADOdb in our spare time, and we all have busy lives...

With regards to the PR, I noticed you just pushed a bunch of commits unrelated to the original change, maybe you merged the master branch or something like that... this makes it inconvenient for us to review and merge the changes, so could you please reset your feature branch by rebasing it on top of master and force-push it to github ?

atrol:
@dregad you added a wrong label to it, should certainly be mssql.

merijnvdk:
Hi yes, just wanted to know if there was nothing wrong. As you just noticed I am not that experienced with git yet ;) Bad luck you just happend to look when I messed the feature branch. Should be okay now.

PR #316, opened 2017-03-07
mssql: string keys are not allowed in parameters arrays

athaden:
- The method "ErrorMsg" returns "false" for all other drivers if no error occurred.
- I added a switch in method "adodb_log_sql", because string keys are not allowed in parameters arrays for mssqlnative driver.

### DIFF
--- a/adodb-perf.inc.php
+++ b/adodb-perf.inc.php
@@ -147,7 +147,7 @@ function adodb_log_sql(&$connx,$sql,$inputarr)
 		if (empty($d)) $d = date("'Y-m-d H:i:s'");
 		if ($conn->dataProvider == 'oci8' && $dbT != 'oci8po') {
 			$isql = "insert into $perf_table values($d,:b,:c,:d,:e,:f)";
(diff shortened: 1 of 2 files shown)


dregad:
@athaden thanks for your contribution, and apologies for the late reply.

> The method "ErrorMsg" returns "false" for all other drivers if no error occurred.

Can you please clarify, I'm not sure I fully understand the rationale for the proposed change, and what the relationship with "other drivers" is. This function basically returns the results of the driver-specific error function (e.g. mysqli_error(), pg_result_error()...). 

If consistency is the goal here, then why not just return *null* since this is what [sqlsrv_errors](http://php.net/manual/en/function.sqlsrv-errors.php) does when there were no errors ?

> I added a switch in method "adodb_log_sql", because string keys are not allowed in parameters arrays for mssqlnative driver.

I will merge this separately.

dregad:
I will close this PR as a fix for the string keys issue (b4ece10049dc9130e65059c2a691c87dad46f36c).

Please open a separate to follow-up with the ErrorMsg() issue if needed.

Issue #192, opened 2016-01-10
Adopt PSR-4 autoloading standard


dregad:
Currently, inclusion of source files required prior to instantiating new classes is managed manually in several places in the code.

The library needs to be refactored to follow an industry standard for autoloading supported by Composer.

http://www.php-fig.org/psr/psr-4/

antoniopaisfernandes:
@dregad How's this issue?
Can you make the public that branch you wrote of? Perhaps we can help :)

dregad:
@antoniopaisfernandes This is kind of low priority, and due to lack of time to spare for ADOdb these days, I have not made any progress towards this. You can have a look at @gohanman's proposed approach in #317, but at this time I'm not yet certain it will be merged as-is.

codisart:
@dregad Are you still interested for this issue. A lot of work would be needed to do this. And I am quite interested to help. I think it should be more feasable to do with small steps. Do you think you could be available for some code review for some small PR ?

PR #317, opened 2017-03-18
PSR-4 approach acceptable?

gohanman:
Checking in if this is a reasonable approach to chipping away at #192 - take a file containing multiple classes, lift one out into its own file, and namespace it.

### DIFF
--- a/FieldObject.php
+++ b/FieldObject.php
@@ -1,0 +1,26 @@
+<?php
+
+namespace ADOdb;
+
+//==============================================================================================
+// CLASS ADOdb\FieldObject
+//==============================================================================================
+/**
+ * Helper class for FetchFields -- holds info on a column
+ */
+class FieldObject {
+	var $name = '';
+	var $max_length=0;
+	var $type="";
+/*
+	// additional fields by dannym... (danny_milo@yahoo.com)
+	var $not_null = false;
+	// actually, this has already been built-in in the postgres, fbsql AND mysql module? ^-^
+	// so we can as well make not_null standard (leaving it at "false" does not harm anyways)
+
+	var $has_default = false; // this one I have done only in mysql and postgres for now ...
+		// others to come (dannym)
+	var $default_value; // default, if any, and supported. Check has_default first.
+*/
+}
+
--- a/adodb-csvlib.inc.php
+++ b/adodb-csvlib.inc.php
@@ -1,5 +1,7 @@
 <?php
 
+use ADOdb\FieldObject;
+
 // security - hide paths
 if (!defined('ADODB_DIR')) die();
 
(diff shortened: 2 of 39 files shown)


dregad:
@gohanman thanks for your contribution. Yes, that is the general idea. 

However, before you spend too much time converting the files, please note that I have a work in progress (currently not publicly available branch), to add and standardize PHPDoc blocks that would likely cause a large number of merge conflicts.

I would also like to have a discussion regarding the namespaces before moving forward. If you have already given this some thought, please share your conclusions.

gohanman:
RE: namespaces:

I can see at least two options: dividing by base class e.g. `ADOdb\Connections` or `ADOdb\RecordSets` or dividing by driver e.g. `ADOdb\Drivers\MySQL`, `ADOdb\Drivers\Postgres`, etc. Going by base class is probably more common in most big PHP projects I've seen. It's less clear to me where something like `ADOFieldObject` should land. A catch-all like `ADOdb\Lib` or `ADOdb\Util` isn't that much tidier than leaving things in the root namespace.

For backward compatibility it's probably worth allowing exceptions to PSR-1 naming conventions. Having some lowercase class names like `ADOdb\Connections\postgres9` ought to make it easier to keep `ADONewConnection('postgres9')` working as expected. Maybe that's another upside to grouping things by base class.

dregad:
I haven't thought this all the way through, but it seems to me that dividing by driver offers the advantage of regrouping all code for that driver in a single directory, i.e. under `ADOdb\Drivers\MySQL` we would find classes Connection, RecordSet, DataDict, etc, so maybe easier for development and maintenance.

I am fine with keeping "global" classes in the root namespace for now, I agree that there is not much added value to use `ADOdb\lib` (unless there are a lot of classes).

With regards to BC and making exceptions to the PSR standard, I wouldn't worry about it too much considering that it should be fairly simple to provide compatibility within the ADONewConnection function.

Speaking of that, I was thinking of moving some of the functionality currently provided via standalone functions into a generic ADOdb class, so instead of `$db = ADONewConnection('mysql');` we would do `$db = new ADOdb('mysql'); or something similar.

gohanman:
OK. With the driver route, what's your opinion on closely related drivers - i.e., does `ADOdb\Drivers\Postgres9\Connection` extends `ADOdb\Drivers\Postgres8\Connection`, or is there a single `ADOdb\Drivers\Postgres` namespace containing classes for the different versions?

dregad:
A lot of these "related drivers" are legacy things that should be refactored. 

In the case of PostgreSQL, since the earliest supported version as of this writing is 9.2 (EOL in Sept 2017), the plan was to merge all the drivers into a single class. 

And if we do end up needing a specialized class for version-specific functionality, I think it would make sense to have it in the `ADOdb\Drivers\Postgres` namespace.

PR #320, opened 2017-03-27
Add support for new memcached extension.

crebelein:
I've been having trouble getting the legacy memcache extension (https://pecl.php.net/package/memcache) compiled in OEL7 with PHP 7.1.x. Getting the new memcached extension (https://pecl.php.net/package/memcached) compiled however, is trivial.

Based on the limited use of caching in ADOdb, adding support for the new memcached extension appears to be pretty straight forward. I have not done exhaustive tests of these changes but, depending on your desired implementation, this is a quick tweak and appears to work well.

Switching from legacy memcache to the new memcached library is simple for the end user;

// legacy memcache support (still functional if desired, 
// overrides new memcached library if both are configured)
$db->memCache=TRUE;
$db->memCacheHost=$cachehost;
$db->memCachePort=11211;
$db->memCacheCompress=FALSE;

// new memcached extension
$db->memCached=TRUE;
$db->memCachedHost=$cachehost;
$db->memCachedPort=11211;
//$db->memCachedCompress=FALSE; // note: not supported

There are many alternative methods to implement this. I'm very open to hearing your input on this one and look forward to getting support for the new memcached extension in ADOdb.

Thanks!

### DIFF
--- a/docs/changelog.md
+++ b/docs/changelog.md
@@ -49,6 +49,10 @@ Older changelogs:
 - session: add 'httponly' flag to cookie. #190
 - xml: support table 'opt' attribute with mysqli. #267
 
+## 5.20.9 - 21-Dec-2016
+
+- mssql: fix syntax error in version matching regex #305
+
 ## 5.20.8 - 17-Dec-2016
 
 - mssql: support MSSQL Server 2016 and later #294
--- a/drivers/adodb-mssqlnative.inc.php
+++ b/drivers/adodb-mssqlnative.inc.php
@@ -153,7 +153,7 @@ class ADODB_mssqlnative extends ADOConnection {
 	 */
 	function ServerVersion() {
 		$data = $this->ServerInfo();
-		preg_match('/^\d{2}', $data['version'], $matches);
+		preg_match('/^\d{2}/', $data['version'], $matches);
 		$version = (int)reset($matches);
 
 		// We only support SQL Server 2005 and up


mnewnham:
1. I'm assuming that you are using the memcached version 3 library for PHP 7.  Did you ever use the version 2 library for PHP 5?
2. One of the issues we have had in the past for memcached is lack of support for Windows in the libraries. Do you know if that is still the case?

crebelein:
1. I used the memchached extension linked above in conjunction with libmemcached-1.0.18. With PHP 5, I always used the memcache extension (also linked above).
2. That's a great question that I simply cannot answer. I don't even have a Windows system to test on at the moment. Unix is the only environment that I use PHP in, mainly CentOS and OEL at the moment.

mnewnham:
@dregad , I think we can pull this ok.  I can update the documentation to reflect that the memcached library is recommended for PHP7 and that we haven't currently tested under Windows.  I have a memcached vm server that I used to write up the documentation and could probably test it

dregad:
@crebelein Thanks for your contribution. I had a quick look a the code (but did not actually test it). 

What I don't really like is that this introduces quite a bit of code duplication in ADOdb base class and functions including redundant class properties, and considering that legacy memcache and the new memcached are mutually exclusive, I think it would make more sense to reuse the existing properties, and use $memCache as an integer to define the type of cache to use, via class constants e.g. CACHE_OFF = 0, CACHE_LEGACY = 1 (== true, so we maintain backwards compatibility), CACHE_MEMCACHED = 2...

Then the other properties can be reused as-is, and we can remove most of the duplicated code.

I'm also wondering (but not sure if it's a good idea or even feasible as I didn't analyze in detail), if we should maybe define an abstract cache base class, extended for specialized behavior: file or memory cache.

crebelein:
@degrad, I couldn't agree more. This could certainly be done in a much better way. The primary reason that I submitted it as "duplication" was that it was a quick and dirty proof of concept. Do you have a preference for how it's implemented in the end?

Another option similar to what you proposed with $memCache, would be to add a $memCacheExtension variable to adodb.inc.php (defaulting to 'memcache' in PHP<7 and 'memcached' in PHP=>7). It would be simple enough to add the relevant conditionals based on this variable to adodb-memcache.lib.inc.php thus significantly reducing the duplicated code. Although one could argue that this method of implementation could be more difficult to extend in the future, the use of caching is basic enough that I personally don't think that's an issue.

Thoughts on this? I'm happy to refactor the code myself and resubmit the PR once we decide on a direction. I'm less familiar with any precedence that may have been established within ADOdb than you obviously are so I will await your direction prior to proceeding.

Thanks!

dregad:
> add a $memCacheExtension variable to adodb.inc.php 

You mean a global variable ? If so, I'd rather avoid this approach. I think we have too many of these already, IMO it is better to define all options as class properties, so a given connection object is self-contained.

crebelein:
Sounds good. I'll rewrite this w/out the duplication and resubmit the PR. Thanks!

dregad:
Thank you !

PR #333, opened 2017-04-28
Geometry datatypes

Mike-Benoit:
In order to utilize some of the advanced datatypes (ie: geometry) with MySQL and PostgreSQL (more so MySQL in this case), add  'Z' (RAW) datatype that passes data through to the DB unmodified. 

Then teach ADOdb about the specific geometry datatypes for each database.


chc88:
As more db's are getting these types I think it would be a good idea to support them.

However every database seems to have it's own way of storing the geometry datatypes (even in your example of postgres and mysql. It looks to me that mssql has even a different approach.

Maybe this is for now an acceptable option, but for the future some sort of logic should be added for the different geometry types? To make the db-abstraction complete again?

Mike-Benoit:
Ran into another case where the default type being set to 'N' (numeric) causes problems with ENUM types in PGSQL.

While adding support directly into ADOdb for some of these types would be nice, it may not be practical (especially for geometry types), but we should at least have a work around for it, and default to a RAW ('Z') type I think is the only way to do that without having to add every type under the sun to ADODB ahead of time.

Mike-Benoit:
@mnewnham Any ideas why you reverted the "return ADODB_DEFAULT_METATYPE" change in https://github.com/ADOdb/ADOdb/commit/d365ba9aedb185b0226203fbda457a6ae7b04a08#diff-f44aea90186ce5422f8d671db6f0dee2L1214 ? It seems this breaks the entire point of ADODB_DEFAULT_METATYPE, which I was thinking of using to handle more advanced data-types here.

Mike-Benoit:
@dregad Would it be crazy to consider getting this into v5.21? 

Sadly ADODB currently can't be used with a PostgreSQL database that uses a UUID data type, or any geometry types.

mnewnham:
See https://github.com/ADOdb/ADOdb/commit/3c8d1e5f20aab232a68853f93532a5d840f38e09 for information about a generic method for defining the datatypes you need. This will appear in 5.22

Issue #58, opened 2014-11-11
connection to mssql with Port fails


schmanat:
Hi

Yesterday i tried to connect to an mssql (Microsoft Server 2005) with the following dsn string.
`mssql://user:password@10.10.10.19:63341/database`

I couldn't connect to the database because in the drivers/adodb-mssql.inc.php there is a function called

    function _connect($argHostname, $argUsername, $argPassword, $argDatabasename,$newconnect=false)
	{
		if (!function_exists('mssql_pconnect')) return null;
		$this->_connectionID = mssql_connect($argHostname,$argUsername,$argPassword,$newconnect);
		if ($this->_connectionID === false) return false;
		if ($argDatabasename) return $this->SelectDB($argDatabasename);
		return true;
	}

The parameter `$argHostname` holds only the hostname (10.10.10.19) and ignore the port.
So it's not possible to connect to an mssql server with a port.

dregad:
I don't use mssql so I can't test. Can you please check if this is still an issue in latest master ? Maybe #63 changes this behavior.

mnewnham:
Is this a connection from a windows machine? If so the correct syntax is:

`hostname, port`  and not `hostname:port`

Please let me know if it is a freetds connection and I will set up a test

dregad:
@schmanat you have not provided any feedback, so I'm closing this; feel free to reopen with the requested information.

PR #340, opened 2017-05-22
Ldap connection fix

Mike-Benoit:
Commit 53d13c0f64ad97ff87fa3b74fa4a4e896e691c9b fixing issues #58 and #63 introduced a bug that broke LDAP connection handling when using URIs such as ldap:// or ldaps://. 

This prevents host:port parsing when "://" exists in the host string as we then assume its a URI and that the driver itself will handle it directly.

Might be a good idea to add some unit tests for LDAP connections using full URIs, maybe some other DBs too.

EDIT [dregad]: referenced offending commit and reworded 1st paragraph


dregad:
@Mike-Benoit thanks for the PR; the offending commit 53d13c0f64ad97ff87fa3b74fa4a4e896e691c9b made changes not only to *Connect()* but also to *PConnect()* so I would expect a fix to be applied to both functions; was it intentional that you did not update PConnect ?

It's somewhat off topic of this PR's scope, but looking at the code I'm wondering whether it would not be better to rely on *parse_url()* instead of several strpos() calls to account for hostnames in the form of an URL. Something like (not tested):

``` php
$this->host = $argHostname;
$this->port = null;
$url = parse_url($argHostname);
if (is_array($url) && !isset($url['scheme']) && isset($url['host'])) {
    $this->host = $url['host'];
    if (isset($url['port'])) $this->port = $url['port'];
}
```

Thoughts ?

I don't see much benefit of the single vs double quotes change. Personally, but that's mostly out of habit from work in other projects, I tend to favor the use of single quotes unless escape sequences or variable substitutions are actually needed .

Mike-Benoit:
Yeah, I would agree that relying on parse_url() is probably a better long-term approach. 

Are you wanting to ditch this PR and create your own, or should re-do this PR from scratch?

Mike-Benoit:
@dregad Any update regarding this? Secure LDAP connections are broken without this fix, which seems like a pretty major regression to me. 

I'm concerned the ADODB project is dying a slow death here with bug fixes and push requests taking months, years, multiple years to be addressed.  I'm sure other projects that rely on ADODB have just lost all hope and maintain their own fork for features/bug fixes like we have essentially done. I understand you guys are busy, but are you or @mnewnham willing to open up the project to new committers to possibly help breathe some new life into the project?

dregad:
Hello Mike, sorry for the delay - I missed your Aug. 12th update.

>Are you wanting to ditch this PR and create your own, or should re-do this PR from scratch?

Feel free to amend the PR or create a new one, as you prefer. I do not have the bandwidth to do it myself (as you have guessed)

> are you or @mnewnham willing to open up the project to new committers to possibly help breathe some new life into the project?

I'm certainly open to the idea, will discuss it with Mark. Are you volunteering ?

Mike-Benoit:
> I'm certainly open to the idea, will discuss it with Mark. Are you volunteering ?

Yes.

Mike-Benoit:
@dregad Is this along the lines of what you were thinking?

adodb.inc.php
```diff
@@ -658,6 +659,29 @@ function Time() {
...
 	}
 
+	/**
+	 * Parses the hostname to extract the port.
+	 * Overwrites $this->host and $this->port, only if a port is specified.
+	 * The Hostname can be fully or partially qualified,
+	 * ie: "db.mydomain.com:5432" or "ldaps://ldap.mydomain.com:636"
+	 * Any specified scheme such as ldap:// or ldaps:// is maintained.
+	 *
+	 * @return true
```
dregad: Is there any added value of returning _true_ vs _void_ ? I would personally use the latter (and get rid of the _return_ statement at line 682). Thoughts ?

adodb.inc.php
```diff
@@ -658,6 +659,29 @@ function Time() {
...
+	 */
+	function parseHostNameAndPort() {
```
dregad: This is really an internal function, I don't see any benefit of a public method. Suggest to make it _protected_.

dregad:
> Is this along the lines of what you were thinking?

Precisely. Thanks ! 
I committed a few minor adjustments, and posted a couple of questions.

adodb.inc.php
Mike-Benoit: I was originally thinking it could return TRUE if a port was parsed, and FALSE if not. But we aren't checking the return value currently so I never bothered. My habit is always to have a return value no matter what, but I'm fine with no return at all.

adodb.inc.php
Mike-Benoit: I'm fine with that too.

PR #342, opened 2017-05-24
Fixed annotations for GetOne()

HaaseIT:


adodb.inc.php
```diff
@@ -1709,13 +1709,14 @@ function CacheGetAssoc($secs2cache, $sql=false, $inputarr=false,$force_array = f
...
-	* @param sql			SQL statement
-	* @param [inputarr]		input bind array
-	*/
-	function GetOne($sql,$inputarr=false) {
+	 * Return first element of first row of sql statement. Recordset is disposed
+	 * for you.
+	 *
+	 * @param string		$sql		SQL statement
+	 * @param array|bool	$inputarr	input bind array
+	 * @return string|int|bool
```
dregad: I think it would be more appropriate to use a more generic `@return mixed`.

HaaseIT: Whatever floats your boat :)

dregad:
Thanks for your contribution

HaaseIT:
Looking forward to use it in the next release :)

I will do more small fixes whenever they come up.

PR #346, opened 2017-06-01
Add "Update or Insert" (UPSERT) functionality in AutoExecute

chc88:
Very frequently we see the need to have an Update or Create functionality.
When this is the case you would like to update an existing record if possible, and create it when it is not yet present in the table.

In this pull request the AutoExecute function is extended with this "UPSERT" functionality.

To check if a record already exists the PO_RecordCount function is used.

### DIFF
--- a/adodb.inc.php
+++ b/adodb.inc.php
@@ -179,7 +179,7 @@ if (!defined('_ADODB_LAYER')) {
 	 */
 	define('DB_AUTOQUERY_INSERT', 1);
 	define('DB_AUTOQUERY_UPDATE', 2);
-
+	define('DB_AUTOQUERY_UPSERT', 3);
 
 	// PHP's version scheme makes converting to numbers difficult - workaround
 	$_adodb_ver = (float) PHP_VERSION;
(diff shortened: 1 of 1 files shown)


Mike-Benoit:
Doesn't this introduce race conditions between the select and insert/update calls?

What if a database supports UPSERT natively? Like recent versions of PostgreSQL and I'm pretty sure Oracle/MS-SQL too? It would be nice if we had an easy way to utilize the native functionality which should avoid the worst part of these race conditions. Of course this would require plumbing into the driver itself to generate the necessary DB specific SQL.

chc88:
Mike, thanks for your feedback. You have a point there.

I could relocate the logic to ADOConnection so it could be overridden in the driver-specific classes (and in AutoExecute it would be the same as for the getUpdateSQL or getInsertSQL).

As for the racing conditions, in my usecases I have not seen it happen (yet). 
The code in AutoExecute could be wrapped with the Transaction-functions when available (how could we check if the current driver accepts transctions?)

Let me know what you think.

Mike-Benoit:
Might want to let @dregad @mnewnham weigh in on the best way to delegate this functionality to the drivers.

As for the race condition issue, the only way I'm aware of fixing that is to start a serializable transaction which can really hurt performance, and should throw an exception if it happens to fail/timeout and offer a way to retry. My guess is that if we tried to implement this that way it would have so many drawbacks to be useless in the real world, at least for anyone that cares about performance and reliability. It also will cause problems for the load balancing class I proposed a few years ago...

Don't get me wrong, supporting upsert would be awesome, but it's not an easy thing to do for databases that don't already support it natively. See the following for a interesting read: https://www.depesz.com/2012/06/10/why-is-upsert-so-complicated/

chc88:
Ok, with the comments of dregad I agree on not taking further actions on this pull-request

PR #347, opened 2017-06-01
Fix: pdo sqlsrv auto-commit error

marcelto:
Fixes PDO Error: The auto-commit mode cannot be changed for this driver

Excludes calls setting attribute PDO::ATTR_AUTOCOMMIT

### DIFF
--- a/drivers/adodb-pdo_sqlsrv.inc.php
+++ b/drivers/adodb-pdo_sqlsrv.inc.php
@@ -21,52 +21,8 @@ class ADODB_pdo_sqlsrv extends ADODB_pdo
 
 	function BeginTrans()
 	{
-		if (!$this->hasTransactions) {
-			return false;
-		}
-		if ($this->transOff) {
-			return true;
-		}
-		$this->transCnt += 1;
-
-		return $this->_connectionID->beginTransaction();
-	}
-
-	function CommitTrans($ok = true)
-	{
-		if (!$this->hasTransactions) {
-			return false;
-		}
-		if ($this->transOff) {
-			return true;
-		}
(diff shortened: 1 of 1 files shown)


dregad:
Instead of duplicating all the code in 3 methods to exclude one setAttribute call, I'm wondering if would not make more sense to let the *ADODB_pdo* parent class handle the case where the underlying driver does not support setting ATTR_AUTOCOMMIT (by suppressing the error or catching the exception). 

Thoughts ?

marcelto:
I definitely understand the desire to avoid code duplication but for what it's worth I'm not a fan of suppressing errors/exceptions.  I've been burned before with errors that weren't originally anticipated cause the code to fail.  Then I'm left with a library that doesn't work, lots of unfamiliar code, and no error to go on.  

What would you think about a method at the driver level that sets ATTR_AUTOCOMMIT?  So the parent class would just call the driver method and the details would be handled by the driver.

dregad:
I was just making a suggestion; I agree that suppressing error is generally not a good idea, although catching a single, specific exception would be fine in my book. That being said, your approach sounds good too.

marcelto:
No worries, I understand and agree that catching a specific exception would be fine.  My suggestion would require updates to the other drivers used by the parent so it may not be ideal.  It's your call.  What would you like to do?

dregad:
Defining a method in the parent and overriding in the child classes is probably cleaner in terms of OOP. Let's go for that.

dregad:
Fixed in b741929818359e17b0b1725f6dfab1d97bccbdda. Thanks for your contribution !

PR #348, opened 2017-06-01
Fix typo and suppress error on db connect.

ryanpcmcquen:
This way if the connection fails, the failure can be captured and used as intended.

### DIFF
--- a/drivers/adodb-mysqli.inc.php
+++ b/drivers/adodb-mysqli.inc.php
@@ -121,7 +121,7 @@ class ADODB_mysqli extends ADOConnection {
 		if ($persist && PHP_VERSION > 5.2 && strncmp($argHostname,'p:',2) != 0) $argHostname = 'p:'.$argHostname;
 
 		#if (!empty($this->port)) $argHostname .= ":".$this->port;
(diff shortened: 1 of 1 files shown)


dregad:
To my knowledge, mysqli_real_connect() returns _false_ in case of error, which is checked just below, so I'm not sure why we'd need to suppress errors with `@`... Did you experience an actual problem without the error suppression operator ? If so, please share the details.

ryanpcmcquen:
Yes, without the error suppression operator I was unable to get past a warning that it created (I'm using Whoops on my API). I'll break things again to get you the exact error.

dregad:
@ryanpcmcquen I would appreciate if you could provide a code snippet to reproduce the problem

ryanpcmcquen:
Sorry, I have to break our database to do it. I'll try to carve out some time this week but have been swamped.

ryanpcmcquen:
I get a timeout error without ignoring warnings (`@`):

```
 Whoops \ Exception \ ErrorException (E_WARNING)
mysqli_real_connect(): (HY000/2002): Connection timed out
```

screen shot 2017-06-28 at 9 43 44 am

dregad:
Thanks for your contribution.

PR #353, opened 2017-06-16
Added support for connecting using windows authentication

andrerunge:
To connect to the SQL Server using windows authentication the username and password arguments have to be omitted. So i just check if these are null or have a length of 0. If both the username and password are not set I do not add them to the connection info. This way the native SQL driver connects using windows authentication.

### DIFF
--- a/drivers/adodb-mssqlnative.inc.php
+++ b/drivers/adodb-mssqlnative.inc.php
@@ -475,9 +475,12 @@ class ADODB_mssqlnative extends ADOConnection {
 
 		$connectionInfo 			= $this->connectionInfo;
 		$connectionInfo["Database"]	= $argDatabasename;
-		$connectionInfo["UID"]		= $argUsername;
-		$connectionInfo["PWD"]		= $argPassword;
-
+		if (!((is_null($argUsername) || (is_string($argUsername) && strlen($argUsername) === 0)) &&
+			   is_null($argPassword) || (is_string($argPassword) && strlen($argPassword) === 0))) {
+			$connectionInfo["UID"]=$argUsername;
+			$connectionInfo["PWD"]=$argPassword;
+		}
+		
 		/*
 		* Now merge in the passed connection parameters setting
 		*/


mnewnham:
Could you provide a code snippet so that we can update the documentation please

andrerunge:
For windows authentication just pass NULL or empty strings as username and password.

$db = ADONewConnection('mssqlnative');
$db->setConnectionParameter('ReturnDatesAsStrings',false);
$db->connect('SERVER\SQLEXPRESS‘,NULL,NULL,'NORTHWND’);
if ($db->isConnected()) {
	echo "Connected!";
} else {
	echo "Not Connected!";
}


or

$db = ADONewConnection('mssqlnative');
$db->setConnectionParameter('ReturnDatesAsStrings',false);
$db->connect('SERVER\SQLEXPRESS‘,’’,'','NORTHWND’);
if ($db->isConnected()) {
	echo "Connected!";
} else {
	echo "Not Connected!";
}

mnewnham:
Thanks

drivers/adodb-mssqlnative.inc.php
```diff
@@ -475,9 +475,12 @@ function _connect($argHostname, $argUsername, $argPassword, $argDatabasename)
 
 		$connectionInfo 			= $this->connectionInfo;
 		$connectionInfo["Database"]	= $argDatabasename;
-		$connectionInfo["UID"]		= $argUsername;
-		$connectionInfo["PWD"]		= $argPassword;
-
+		if (!((is_null($argUsername) || (is_string($argUsername) && strlen($argUsername) === 0)) &&
+			   is_null($argPassword) || (is_string($argPassword) && strlen($argPassword) === 0))) {
```
dregad: Maybe I'm missing something here, but considering that in PHP `null == ''`, it feels unnecessary to test separately if the parameter is null or empty string; likewise, why the _strlen()_ call and not just compare to empty string ? Just typecast the parameter to string and be done with it...

``` php
if((string)$argUsername != '' || (string)$argPassword != '') {
```

Also, a comment before the if statement explaining why we're doing this would be useful (along the lines of what you wrote in the PR's description).

andrerunge:
Hi,

but I don’t think that null == ’’.

Just try:

is_null(NULL) is TRUE
and is_null(‘') is FALSE

dregad:
> I don’t think that null == ’’.

It is. If you use a "normal" PHP comparison, then `0 == false == null == ''` - try it.

`is_null($a)` is the equivalent of a _strict-type_ comparison `$a === null`. 

To avoid the case where the caller would send 0 (since `0 == ''`), use of a typecast ensures we always compare strings: `(string)0 === '0' != '')`.

laurent-n:
Hello,

I've worked on the same subject today
Here is my patch

		if($connectionInfo["UID"]=='') { // Request windows Authentication if user is not populated
			unset($connectionInfo["UID"]);
			unset($connectionInfo["PWD"]); 			
			}

Placed just after that instruction on _connect method.
		$connectionInfo["UID"]		= $argUsername;
		$connectionInfo["PWD"]		= $argPassword;

mnewnham:
I have updated the SQL Server documentation to reflect this feature

mnewnham:
I have used you code and added some commenting on a recent patch to the driver, thanks. no need to pull this

PR #359, opened 2017-08-01
MSSQL requires table alias in _adodb_getcount() query

monzol:
Currently the _ADODB_ALIAS_ is only used for postgres and mysql, but it's also necessary for mssql, otherwise the query assigned to $rewritesql on line 422 fails and produces the following warning:
PHP Warning (2):  mssql_query(): message: Incorrect syntax near ')'. (severity 15) in /usr/share/php/adodb/drivers/adodb-mssql.inc.php on line 832


dregad:
Fixed in 7860e95ade59e5b83727def93dc7048f1c99e018. Thanks for your contribution !

PR #361, opened 2017-08-03
Use include_once instead of include

merijnvdk:
Most of the includes in adodb.inc.php are done with include_once but
in a couple locations its still done with include.

This is the case for adodb-lib.inc.php which is conditionally included
based on a global var. I ran into problems with running adodb in debug
mode while running phpuni tests. Phpunit will reset the global vars
and thus break the include mechanism. This is a problem of phpunit
and can be worked around. However it would make adodb more robust
to just use include_once everywhere. It will remove the dependency on
the $ADODB_INCLUDED_LIB global var.

I also replaced one include statement of adodb-time.inc.php so
include_once is used everywhere.

### DIFF
--- a/adodb-active-record.inc.php
+++ b/adodb-active-record.inc.php
@@ -522,7 +522,7 @@ class ADODB_Active_Record {
 			$activetab->_created = time();
 			$s = serialize($activetab);
 			if (!function_exists('adodb_write_file')) {
-				include(ADODB_DIR.'/adodb-csvlib.inc.php');
+				include_once(ADODB_DIR.'/adodb-csvlib.inc.php');
 			}
 			adodb_write_file($fname,$s);
 		}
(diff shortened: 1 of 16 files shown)


dregad:
Thanks for your contribution.

> I also replaced one include statement of adodb-time.inc.php so

I'm confused - your PR only includes changes to _adodb.inc.php_... am I missing something ?

A quick grep also shows calls to `include` in the following files, which should probably be updated as well for consistency:
- adodb-active-record.inc.php
- adodb-active-recordx.inc.php

merijnvdk:
no I was talking about the file being included, adodb-lib.inc.php and adodb-time.inc.php.

Okay I can fix adodb-active-record.inc.php and adodb-active-recordx.inc.php too for consistency.

merijnvdk:
Also went through the drivers and replaced a bunch of include statements. But this means the 'core' label you added is not 100% true anymore.

dregad:
Thanks for following up.

> this means the 'core' label you added is not 100% true anymore.

Technically true, but considering that the change is purely technical and should have no impact on the drivers' operation, I think that's fine.

PR #367, opened 2017-09-14
Optimize FieldTypesArray with static variable

joebordes:
I was profiling my application which uses ADOdb and this optimization dropped my timing significantly, but I would like to ask if this change is valid.
The tests I made seemed to be working correctly.

### DIFF
--- a/adodb.inc.php
+++ b/adodb.inc.php
@@ -4182,7 +4182,10 @@ http://www.stanford.edu/dept/itss/docs/oracle/10g/server.101/b10759/statements_1
 	 *
 	 */
 	function FieldTypesArray() {
-		$arr = array();
+		static $arr = array();
+		if (count($arr)>0) {
+			return $arr;
+		}
 		for ($i=0, $max=$this->_numOfFields; $i < $max; $i++)
 			$arr[] = $this->FetchField($i);
 		return $arr;
--- a/adodb.inc.php
+++ b/adodb.inc.php
@@ -4183,11 +4183,11 @@ http://www.stanford.edu/dept/itss/docs/oracle/10g/server.101/b10759/statements_1
 	 */
 	function FieldTypesArray() {
 		static $arr = array();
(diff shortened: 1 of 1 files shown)


dregad:
Looks OK to me but if since this is about performance then you should probably use `empty($arr)` instead of `count($arr) > 0`.

@mnewnham thoughts ?

joebordes:
Sounds correct to me. Have a look and let me know if you want me to change anything else.

Thanks!!

mnewnham:
Why not

````php
static $arr = false;
if ($arr) 
    return $arr;

$arr = array();

blah blah blah....
````

Mike-Benoit:
Would be nice to get at least some idea of the actual performance improvement too. Is this a 1% improvement with a certain use case or 50%?

Mike-Benoit:
I think this just introduced another problem along the same lines of: #381.

Anytime data is cached, even in memory, there needs to be a way to clear it when schema changes and static variables make that job much more difficult.

dregad:
Revert?

Mike-Benoit:
I would leave it in place for now, but I think we should figure out a better caching solution as part of #381.

dregad:
OK.

dregad:
> I think this just introduced another problem along the same lines of: #381.
> 
> Anytime data is cached, even in memory, there needs to be a way to clear it when schema changes and static variables make that job much more difficult.

@Mike-Benoit it turns out I've just been hit by this... See #687

peterdd:
@dregad If you revert this you can move out

`$max=$this->_numOfFields;`

out of the loop I think. But probably just a micro optimazation..

dregad:
> If you revert this you can move out
> 
> `$max=$this->_numOfFields;`
> 
> out of the loop I think. But probably just a micro optimization..

OK

Not yet sure if I'll revert or maybe try to find an alternative to using a static variable.

PR #368, opened 2017-09-14
LAST_INSERT_ID() is not accessible after the transaction

gjergjsheldija:
Insert_ID returns 0 when executed after a transaction has been committed, fixed with calling _insertid() before transaction get's committed and saving to parents $lastInsID

### DIFF
--- a/adodb.inc.php
+++ b/adodb.inc.php
@@ -1394,7 +1394,7 @@ if (!defined('_ADODB_LAYER')) {
 	 * @return  the last inserted ID. Not all databases support this.
 	 */
 	function Insert_ID($table='',$column='') {
-		if ($this->_logsql && $this->lastInsID) {
+		if ($this->_logsql || $this->lastInsID) {
 			return $this->lastInsID;
 		}
 		if ($this->hasInsertID) {
--- a/drivers/adodb-mysqli.inc.php
+++ b/drivers/adodb-mysqli.inc.php
@@ -203,6 +203,7 @@ class ADODB_mysqli extends ADOConnection {
 		if (!$ok) return $this->RollbackTrans();
 
 		if ($this->transCnt) $this->transCnt -= 1;
+		$this->lastInsID = $this->_insertid();
 		$this->Execute('COMMIT');
 
 		//$this->Execute('SET AUTOCOMMIT=1');


dregad:
@gjergjsheldija Thanks for your contribution. Can you please clarify why you opened 2 pull requests with the same title (see #369) ? Was it a mistake ? If not, what is the relationship between them ? If they are the same issue, then please regroup the commits in a single branch for review.

gjergjsheldija:
@dregad Merged both of the changesets in the patch-1 branch. 
Thank you

dregad:
It would appear that the mysqli driver is indeed resetting the last_insert_id when committing a transaction. This does not seem to be documented, as far as I can tell.

Need to think about the proposed fix.

gjergjsheldija:
hi @dregad 
not quite undocumented, [example](https://forum.codeigniter.com/thread-61881-post-319000.html#pid319000)

dregad:
> not quite undocumented, [example](https://forum.codeigniter.com/thread-61881-post-319000.html#pid319000)

A forum post that lacks references and written by an unknown contributor does not quite constitute official documentation in my book :wink:

gjergjsheldija:
@dregad you can test the example i've provided [here](https://github.com/ADOdb/ADOdb/pull/369#issuecomment-330139298)

ad1rie1:
Hello, can you merge ? i have the same problem :(

mnewnham:
I've had a think about this one, to me it makes no sense to extend out the availability of the last insert_id because of the way that there is a single thread that holds the last value of **any** table. This feature only has usability if you can absolutely determine the table on which the last insert occurred:
````php
$db->startTrans();
//Do a single insertion or multiple on the same table
$db->completeTrans();
//Now get the insert_id
`````
When in reality, its more like:
````php
$db->startTrans();
//Do 19,600 insertions on 16 different tables, in 10 nested classes, with no guarantee which table was last
$db->completeTrans();
//Now get the insert_id (but of what?)
````
I actually think that:
1. Making the feature available would have no clear benefit to the product
2. It's better addressed in the users own code
3. We would have to test it out on all the other drivers

@dregad , if you think I'm wrong on this, let's re-open and pull. There's nothing intrinsically wrong with the code, I just don't think its an appropriate use of the driver.

PR #369, opened 2017-09-14
LAST_INSERT_ID() is not accessible after the transaction

gjergjsheldija:
Insert_ID returns 0 when executed after a transaction has been committed, fixed with calling _insertid() before transaction get's committed and saving to parents $lastInsID

### DIFF
--- a/adodb.inc.php
+++ b/adodb.inc.php
@@ -1394,7 +1394,7 @@ if (!defined('_ADODB_LAYER')) {
 	 * @return  the last inserted ID. Not all databases support this.
 	 */
 	function Insert_ID($table='',$column='') {
-		if ($this->_logsql && $this->lastInsID) {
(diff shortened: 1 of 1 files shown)


mnewnham:
What database are you working with here? I've just scanned the source and it looks like lastInsId is only ever set in adodb-mssql.inc

gjergjsheldija:
with mysql 5.6, using the mysqli driver. the code is as follows :
begin;
insert into table() values();
commit;
select last_insert_id();
-- returns 0


begin;
insert into table() values();
select last_insert_id();
-- returns correct value
commit;

dregad:
See https://github.com/ADOdb/ADOdb/pull/368#issuecomment-346954849

gjergjsheldija:
closing pull request, due to merge of both patches in pull #368

PR #371, opened 2017-09-25
Warning. sybase_connect

zerojano:
Sybase Connect Server message, Changed database context.
Although it is only a warning, it generates an error in the log with connections to external Sybase databases.
The error message is generated in moodle login, with system debugging enabled.

### DIFF
--- a/drivers/adodb-sybase.inc.php
+++ b/drivers/adodb-sybase.inc.php
@@ -150,9 +150,9 @@ class ADODB_sybase extends ADOConnection {
 		}
 
 		if ($this->charSet) {
-			$this->_connectionID = sybase_pconnect($argHostname,$argUsername,$argPassword, $this->charSet);
+			$this->_connectionID = @sybase_pconnect($argHostname,$argUsername,$argPassword, $this->charSet);
 		} else {
-			$this->_connectionID = sybase_pconnect($argHostname,$argUsername,$argPassword);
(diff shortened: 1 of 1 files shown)


mnewnham:
Thanks for that patch. @dregad There seems no reason that we can't just merge this immediately

dregad:
@zerojano I'm fine with merging this, but before I do I was wondering if the same fix should also be applied to *connect()* method ?

Can you please clarify the exact error message you're getting, and the context to trigger it (ideally a code sample to reproduce)

zerojano:
@dregad I modified the connection method with Sybase, apparently there is where the error is generated, it is necessary to rescue that when modifying the debugging options> normal, the error is generated.

**Error message:**

https://example.moodle.cl/login/index.php?authldap_skipntlmsso=1

Warning: sybase_connect (): Sybase: Server message: Changed database context to 'webApplication'. (severity 10, procedure N / A) in /var/www/html/moodle/lib/adodb/drivers/adodb-sybase.inc.php on line 138
Skip to main content

× This page should redirect automatically. If nothing happens, please use the follow link below.
(Continue)

Output error, disable automatic redirection.

dregad:
I'm not sure to understand what you mean, maybe something got lost in translation ? 

As far as I can tell, you're not answering my question, which was - should the same fix also be applied to *connect()* method ?

zerojano:
@dregad I failed the translation.
Yes. it is necessary to modify both _connect () and _pconnect () methods,

Greetings.

PR #373, opened 2017-10-03
Replace each() with foreach for PHP7.2 compatibility

marinaglancy:
Function each() is deprecated in PHP 7.2
See https://wiki.php.net/rfc/deprecations_php_7_2

### DIFF
--- a/adodb-datadict.inc.php
+++ b/adodb-datadict.inc.php
@@ -520,7 +520,7 @@ class ADODB_DataDict {
 			list($lines,$pkey,$idxs) = $this->_GenFields($flds);
 			// genfields can return FALSE at times
 			if ($lines == null) $lines = array();
-			list(,$first) = each($lines);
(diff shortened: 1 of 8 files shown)


adodb-datadict.inc.php
```diff
@@ -520,7 +520,7 @@ function RenameColumnSQL($tabname,$oldcolumn,$newcolumn,$flds='')
 			list($lines,$pkey,$idxs) = $this->_GenFields($flds);
 			// genfields can return FALSE at times
 			if ($lines == null) $lines = array();
-			list(,$first) = each($lines);
+			$first  = current($lines);
```
dregad: This is not directly related to your change, but the *list()* call on the next line will throw a PHP notice if *preg_split()* returns an array with a single element (which can happen if $first is empty or does not contain any whitespace)

PR #376, opened 2017-10-14
soft year validation in adodb_validdate

### LINKED ISSUE #375: adodb_validdate: expression is always false
Hi!
This expression:

https://github.com/ADOdb/ADOdb/blob/6cafd5579929c6e7f5ad8a5dadc33ad2a3d87c9a/adodb-time.inc.php#L819

is always false.

Probably it should be:
`if ($y < 1000 || $y > 3000) return false;`

This possible defect found by [AppChecker](https://npo-echelon.ru/en/solutions/appchecker.php)

marcomenzel:
fixes #375 
given year can only be less than 3000 or greater than 1000

### DIFF
--- a/adodb-time.inc.php
+++ b/adodb-time.inc.php
@@ -816,7 +816,7 @@ global $_month_table_normal,$_month_table_leaf;
 
 	if ($marr[$m] < $d) return false;
 
-	if ($y < 1000 && $y > 3000) return false;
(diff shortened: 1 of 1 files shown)


dregad:
Fixed in f17a3cecbe2e1baafab5b68f06fca85ea918526c

dregad:
Thanks for your contribution

Issue #383, opened 2017-12-10
Incorrect Postgres ALTER TABLE Syntax (xmlschema)


boshrin:
On 5.20.9 (and prior versions), if I change a column definition in my schema file, eg

`<field name="bar" type="C" size="512" />`

to

`<field name="bar" type="C" size="1024" />`

incorrect SQL is generated for Postgres. Specifically, something like

`ALTER TABLE foo ALTER COLUMN bar VARCHAR(1024)`

instead of

`ALTER TABLE foo ALTER COLUMN bar TYPE VARCHAR(1024)`

The problem appears to be in `adodb-datadict.inc.php::ChangeTableSQL()`, which reimplements its own ALTER TABLE syntax rather than using the (potential) database specific syntax. A simple fix (though perhaps not the correct one?) is to use `$this->AlterColumnSQL()` instead. Here's a sample patch:

```--- a/app/Vendor/adodb5/adodb-datadict.inc.php
+++ b/app/Vendor/adodb5/adodb-datadict.inc.php
@@ -996,33 +996,8 @@ class ADODB_DataDict {
                        $flds = $holdflds;
                }

-
-               // already exists, alter table instead
-               list($lines,$pkey,$idxs) = $this->_GenFields($flds);
-               // genfields can return FALSE at times
-               if ($lines == null) $lines = array();
-               $alter = 'ALTER TABLE ' . $this->TableName($tablename);
-               $sql = array();
-
-               foreach ( $lines as $id => $v ) {
-                       if ( isset($cols[$id]) && is_object($cols[$id]) ) {
-
-                               $flds = Lens_ParseArgs($v,',');
-
-                               //  We are trying to change the size of the field, if not allowed, simply ignore the request.
-                               // $flds[1] holds the type, $flds[2] holds the size -postnuke addition
-                               if ($flds && in_array(strtoupper(substr($flds[0][1],0,4)),$this->invalidResizeTypes4)
-                                && (isset($flds[0][2]) && is_numeric($flds[0][2]))) {
-                                       if ($this->debug) ADOConnection::outp(sprintf("<h3>%s cannot be changed to %s currently</h3>", $flds[0][0], $flds[0][1]));
-                                       #echo "<h3>$this->alterCol cannot be changed to $flds currently</h3>";
-                                       continue;
-                               }
-                               $sql[] = $alter . $this->alterCol . ' ' . $v;
-                       } else {
-                               $sql[] = $alter . $this->addCol . ' ' . $v;
[...]

PR #385, opened 2017-12-22
Fix Incorrect Postgres ALTER TABLE Syntax

boshrin:
Proposed fix for issue #383 (incorrect ALTER TABLE syntax on Postgres in datadict).

### DIFF
--- a/adodb-datadict.inc.php
+++ b/adodb-datadict.inc.php
@@ -1022,32 +1022,7 @@ class ADODB_DataDict {
 			$flds = $holdflds;
 		}
 
-
-		// already exists, alter table instead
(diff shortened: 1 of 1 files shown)


boshrin:
I don't think this is the correct fix, as it appears to break adding a column to an existing table.

dregad:
@boshrin what is the status of this PR ? Are you working on an alternate fix, or should I close it ?

asmecher:
Here is a modification we've been using for some years on an old version of ADODB, forward-ported to v5.20.12: https://github.com/asmecher/ADOdb/commit/d3b3b3ebfcd76e52689e625f3973c204e3828aa4
If it would be more useful, I could foward-port this further to `master`.

Issue #390, opened 2018-01-30
Table/Column names missing quotes 


marvinwaterman:
In function "_adodb_replace" in file "adodb-lib.inc.php" the function fails to take into account column/table names with spaces / reserve keywords.  All columns and table name should be in quotes. Using PHP and MySQL.
Example code of that will fail:
```
/*
 * Connection assumed
*/
 
*
$tableName = 'products';
$indexColumn = 'Product Code';
$ar = array('Product Code'=>10201,
            'Product Name' =>'Nails',
            'price' => '3.99');
 
$status = $db->replace($tableName,$ar ,$indexColumn, true);
```

Below you'll find the new version of the function that resolves the issue:
```
function _adodb_replace(&$zthis, $table, $fieldArray, $keyCol, $autoQuote, $has_autoinc)
{
//MW - Add Quote for table name
 $table="$zthis->nameQuote$table$zthis->nameQuote";
		if (count($fieldArray) == 0) return 0;
		$first = true;
		$uSet = '';

		if (!is_array($keyCol)) {
			$keyCol = array($keyCol);
		}
		foreach($fieldArray as $k => $v) {
			if ($v === null) {
				$v = 'NULL';
				$fieldArray[$k] = $v;
			} else if ($autoQuote && /*!is_numeric($v) /*and strncmp($v,"'",1) !== 0 -- sql injection risk*/ strcasecmp($v,$zthis->null2null)!=0) {
				$v = $zthis->qstr($v);
				$fieldArray[$k] = $v;
			}
			if (in_array($k,$keyCol)) continue; // skip UPDATE if is key

// MW - Add Quote for column name
			if ($first) {
				$first = false;
				$uSet = "$zthis->nameQuote$k$zthis->nameQuote=$v";
			} else
				$uSet .= ",$zthis->nameQuote$k$zthis->nameQuote=$v";
		}

		$where = false;
// MW - fix quote issues in where clause
		foreach ($keyCol as $v) {
			if (isset($fieldArray[$v])) {
				if ($where) $where .= ' and '."$zthis->nameQuote$v$zthis->nameQuote".'='.$fieldArray[$v];
				else $where = "$zthis->nameQuote$v$zthis->nameQuote".'='.$fieldArray[$v];
			}
		}

		if ($uSet && $where) {
			$update = "UPDATE $table SET $uSet WHERE $where";

			$rs = $zthis->Execute($update);


			if ($rs) {
				if ($zthis->poorAffectedRows) {
[...]

PR #391, opened 2018-01-31
Table/Column names needs to be quoted in replace function

marvinwaterman:
Per issue #390, Using the replace function will fail if table or column name contains a space or is using a MySQL reserve keyword.


dregad:
Replaced by #393

Issue #395, opened 2018-02-18
How to use PDO driver with Session Management


KorbenDallas2:
It seems only 'mysqli' driver is supported for this. 

```
include 'adodb/session/adodb-cryptsession2.php';
$driver = 'pdo'; 
$host = 'localhost';
$user = 'auser'; 
$pwd = 'secret'; 
$database = 'sessiondb';
 
ADOdb_Session::config($driver, $host, $user, $password, $database,$options=false);
session_start();
```

The above fails unless $driver = 'mysqli';

_Fatal error: session_start(): Failed to initialize storage module:_

I am using PDO connections and want to use them here as well. 
If this cannot use $driver = 'pdo' is there a way to get this to use my existing DB connection:  $conn which is using PDO?

KorbenDallas2:
A colleague fixed this and committed a PR

[https://github.com/ADOdb/ADOdb/pull/396](url)

uses  $driver = 'pdo_mysql';

Eet verks!

mnewnham:
@KorbenDallas2 , I reviewed this PR, some extended character set text appears to have been introduced into the file at line 582. Could you ask your colleague to review the code, and if possible re-test the changes **not** using the pdo driver. We can move forward on this once this has been done

KorbenDallas2:
Not sure what you are asking. I have been using this code successfully
(daily) since it was created, and I am not seeing any extended character
set info in the version I have, or in the commit at:

https://github.com/ADOdb/ADOdb/pull/396/commits/9b40399825dd572524baf1802381ca334f5b4b2b

Perhaps github is doing whatever it is you are seeing?

PR #396, opened 2018-02-18
Session pdo support

therealryanbonham:
Add support for using PDO for ADODB Sessions.

To use just specify driver as 'pdo_mysql' in your ADOdb_Session::config call.  

eg.
ADOdb_Session::config('pdo_mysql', $host, $user, $password, $database,$options=false);

### DIFF
--- a/session/adodb-session2.php
+++ b/session/adodb-session2.php
@@ -564,28 +564,44 @@ class ADODB_Session {
 #		assert('$database');
 #		assert('$driver');
 #		assert('$host');
-
-		$conn = ADONewConnection($driver);
-
-		if ($debug) {
-			$conn->debug = true;
-			ADOConnection::outp( " driver=$driver user=$user db=$database ");
-		}
-
-		if (empty($conn->_connectionID)) { // not dsn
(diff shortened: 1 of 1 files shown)


session/adodb-session2.php
```diff
@@ -564,29 +564,44 @@ static function open($save_path, $session_name, $persist = null)
...
+				case 'P': $ok = $conn->PConnect($dsn,$user,$password); break;
+				case 'C': $ok = $conn->Connect($dsn,$user,$password); break;
+				case 'N': $ok = $conn->NConnect($dsn,$user,$password); break;
 				}
 			} else {
-				$ok = $conn->Connect($host, $user, $password, $database);
+				$ok = $conn->Connect($dsn,$user,$password);
+			}
+		}else{
+			$conn = ADONewConnection($driverÏ);
```
peterdd: discussion #395 I think here is the bad char, the I with 2 dots after $driver.

KorbenDallas2: That odd letter 'I' is not in the version I have, it should not be there.

PR #399, opened 2018-03-02
Remove statics from fetchField

leedm777:
In the pdo_sqlsrv driver, static locals were used to cache field names.
Unfortunately, this means that if the results for multiple queries
were mapped, the field names for the first query were used instead of
updated field names for the subsequent queries.

### DIFF
--- a/drivers/adodb-pdo_sqlsrv.inc.php
+++ b/drivers/adodb-pdo_sqlsrv.inc.php
@@ -62,15 +62,9 @@ class ADORecordSet_pdo_sqlsrv extends ADORecordSet_pdo
 	public function fetchField($fieldOffset = 0)
 	{
 
-		static $fieldObjects = array();
 		// Default behavior allows passing in of -1 offset, which crashes the method
 		if ($fieldOffset == -1) {
 			$fieldOffset++;
-		}
-
-		if (isset($fieldObjects[$fieldOffset])) {
-			// Look for cached field offset
-			return $fieldObjects[$fieldOffset];
 		}
 
 		$o = new ADOFieldObject();
@@ -102,8 +96,6 @@ class ADORecordSet_pdo_sqlsrv extends ADORecordSet_pdo
 				break;
 		}
 
-		// Add to the cache
-		$fieldObjects[$fieldOffset] = $o;
 		return $o;
 	}
 }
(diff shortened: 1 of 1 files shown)


drivers/adodb-pdo_sqlsrv.inc.php
```diff
@@ -62,17 +62,12 @@ class ADORecordSet_pdo_sqlsrv extends ADORecordSet_pdo
...
 
-		static $fieldObjects = array();
+		$fieldObjects = array();
 		// Default behavior allows passing in of -1 offset, which crashes the method
 		if ($fieldOffset == -1) {
 			$fieldOffset++;
 		}
 
-		if (isset($fieldObjects[$fieldOffset])) {
-			// Look for cached field offset
```
mbrooks: There are only two hard things in Computer Science: cache invalidation, naming things, and off by one errors.

dregad: @mbrooks joke aside, are you actually trying to make a point here ?

renuzit: No, it was just a joke. I hardly ever get to use that line. But, in all seriousness, how can we make this work with the caching model currently in place in ADODB?

mbrooks: @dregad It was just a joke. I hardly ever get to use that line. In all honesty, how can we help you with the cache problems?

leedm777:
@dregad Good point. I've updated the commit, and rebased against master.

Thanks!

drivers/adodb-pdo_sqlsrv.inc.php
```diff
@@ -122,17 +114,12 @@ class ADORecordSet_array_pdo_sqlsrv extends ADORecordSet_array_pdo
 	 */
 	public function fetchField($fieldOffset = 0)
 	{
-		static $fieldObjects = array();
+		$fieldObjects = array();
```
dregad: @leedm777 you forgot one...

leedm777: @dregad Oops. I think I got it now.

chadxz:
Any other changes needed to this PR?

dregad:
Thanks for your contribution

PR #401, opened 2018-03-06
Fixes SQL Injection vulnerability in multiple drivers

### LINKED ISSUE #400: SQL Injection vulnerability in MySQLi driver
Repeat of #311 but for MySQLi - this likely affects other drivers also

jedi58:
The `SelectLimit` function has a potential SQLi exploit through the use of the `nrows` and `offset` parameters which are not forced to integers.

Fixes #400


jedi58:
This is to fix drivers that have their own `Execute` methods instead of, or as well as `ADOConnection::SelectLimit` which has already been fixed in #311

dregad:
Thanks @jedi58

PR #402, opened 2018-03-09
Concat method missing for odbc_mssql driver

WiXSL:
An error may occur in comparitions that needs casting.


dregad:

PR #406, opened 2018-03-30
PHP-X-version compat: Don't attempt to call `dl()` when the function is disabled

jrfnl:
The `dl()`has been removed/disabled in most SAPIs since PHP 5.3, so chances of it being available are rather slim.

> 7.0.0 	dl() is disabled in PHP-FPM.
> 5.3.9 	dl() is enabled in PHP-FPM, albeit discouraged.
> 5.3.0 	dl() is now disabled in some SAPIs due to stability issues. The only SAPIs that allow dl() are CLI and Embed. Use the Extension Loading Directives instead.

Ref: http://php.net/manual/en/function.dl.php#refsect1-function.dl-changelog

### DIFF
--- a/adodb-pear.inc.php
+++ b/adodb-pear.inc.php
@@ -358,7 +358,7 @@ class DB
 	 */
 	function assertExtension($name)
 	{
-		if (!extension_loaded($name)) {
+		if (function_exists('dl') && !extension_loaded($name)) {
 			$dlext = (strncmp(PHP_OS,'WIN',3) === 0) ? '.dll' : '.so';
 			@dl($name . $dlext);
 		}


adodb-pear.inc.php
```diff
@@ -358,7 +358,7 @@ function parseDSN($dsn)
 	 */
 	function assertExtension($name)
 	{
-		if (!extension_loaded($name)) {
+		if (function_exists('dl') && !extension_loaded($name)) {
```
dregad: Probably better test for *extension_loaded* first, no need to call *function_exists* if the extension is loaded.

jrfnl: I thought exactly the opposite ;-) Why bother checking if an extension is loaded if there's no way to force load it anyway.

dregad: I guess that makes sense too. The question is, which one of the 2 conditions is the most likely to occur. But considering It would probably save about 1 nanosecond (according to the [benchmark posted by some guy on the Drupal project](https://www.drupal.org/project/drupal/issues/1942432#comment-10006777)), let's forget about it :wink:

PR #407, opened 2018-03-30
PHP-X-version compat: Don't bother with magic quotes when not available

jrfnl:
Magic quotes were removed from PHP in v5.4.0.

The return value of `get_magic_quotes_runtime()` will be `0` if `magic_quotes_runtime` is off, `1` otherwise.
And as of PHP 5.4.0, it will always returns `false`.

This minor change prevents trying to set the ini value when the ini directive is not available and cleans up some old/unused/commented out code.

Refs:
* http://php.net/manual/en/info.configuration.php#ini.magic-quotes-runtime
* http://php.net/manual/en/function.get-magic-quotes-runtime.php

### DIFF
--- a/adodb-xmlschema.inc.php
+++ b/adodb-xmlschema.inc.php
@@ -1305,8 +1305,9 @@ class adoSchema {
 	function __construct( $db ) {
 		// Initialize the environment
 		$this->mgq = get_magic_quotes_runtime();
-		ini_set("magic_quotes_runtime", 0);
-		#set_magic_quotes_runtime(0);
+		if ($this->mgq !== false) {
+			ini_set('magic_quotes_runtime', 0);
+		}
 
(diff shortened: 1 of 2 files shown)


dregad:

jrfnl:
@dregad You're welcome. Been a while. I contributed long ago in the SourceForge days, but hadn't looked at the project much over the past few years (until today).

dregad:
Welcome back, then ! Looking forward for more contributions in the future :smiley:

jrfnl:
Not making any promises. Came across ADOdb again yesterday as I was looking to upgrade the PHP version of the server some old sites are running on, but who knows what the future will bring ;-)

dregad:
It would be great if you do, but no worries if you don't. Thanks again ! :smiley:

PR #408, opened 2018-03-30
PHP-X-version compat: Replace use of `each()` with foreach()

jrfnl:
`each()` has been deprecated as of PHP 7.2.
The recommended replacement is using `foreach()`.

Looks like one instance was missed in the previous PR related to this. (#373) Fixed now.

Refs:
* http://php.net/manual/en/migration72.deprecated.php#migration72.deprecated.each-function
* http://php.net/manual/en/function.each.php
* http://php.net/manual/en/control-structures.foreach.php

### DIFF
--- a/drivers/adodb-text.inc.php
+++ b/drivers/adodb-text.inc.php
@@ -255,8 +255,7 @@ class ADODB_text extends ADOConnection {
 					$projtypes = array($this->_types[$i]);
 					$projnames = array($n);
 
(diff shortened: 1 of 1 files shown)


dregad:
Already fixed in 316f4abe2cefdf024ddc197b615136fd7d84177b

jrfnl:
Serendipity - looks like we were doing the same things at the same time. It happens.

PR #409, opened 2018-03-30
PHP-X-version compat: Replace usages of deprecated `$php_errormsg`

### LINKED ISSUE #405: $php_errormsg deprecated in PHP 7.2
http://php.net/manual/en/reserved.variables.phperrormsg.php

jrfnl:
The PHP native `$php_errormsg` variable has been deprecated as of PHP 7.2.
The recommended replacement is `error_get_last()`.

A number of the driver classes used  `$php_errormsg` to track if a DB action caused an error.
As the `error_get_last()` function returns an array and comparing the message therefore is slightly more involved, two new methods have been introduced into the abstract parent class `ADOConnection` to handle the message comparison.

The `resetLastError()` method retrieves the initial (old) error message and resets it to enable checking whether a new error has occurred, even when the new error is the same as the previous one.
The new `getChangedErrorMsg()` method compares the old error message with the current one to determine whether a new message has occurred, and if so, returns the message.

Refs:
* http://php.net/manual/en/migration72.deprecated.php#migration72.deprecated.track_errors-and-php_errormsg
* http://php.net/manual/en/reserved.variables.phperrormsg.php
* http://php.net/manual/en/function.error-get-last.php

Additional notes:
* In the `ADODB_ads::UpdateBlob()` method, the original (old) message was not being retrieved which could result in an unrelated error being recorded. This has been fixed.
* In the `ADODB_db2::_connect()` and `ADODB_db2::_pconnect()` methods, the `$php_errormsg` was being reset, but subsequently not checked for changes, so as the variable in effect was not being used, I've removed the related code.
* In the `ADODB_sybase::ErrorMsg()` method, the `$php_errormsg` was being checked, but no functions which could potential cause errors are being called. As `$php_errormsg` is only available in the scope an error occurred in, this meant that `$php_errormsg` would always be empty, so I've removed the related code.

Fixes #405

### DIFF
--- a/adodb.inc.php
+++ b/adodb.inc.php
@@ -3042,6 +3042,61 @@ http://www.stanford.edu/dept/itss/docs/oracle/10g/server.101/b10759/statements_1
 		return $rs;
 	}
 
+	/**
+	 * Get the last error recorded by PHP and clear the message.
+	 *
+	 * By clearing the message, it becomes possible to detect whether a new error
+	 * has occurred, even when it is the same error as before being repeated.
+	 *
+	 * @return array|null Array if an error has previously occurred. Null otherwise.
+	 */
+	protected function resetLastError() {
+		$error = error_get_last();
+
+		if (is_array($error)) {
+			$error['message'] = '';
+		}
+
+		return $error;
+	}
+
+	/**
+	 * Compare a previously stored error message with the last error recorded by PHP
+	 * to determine whether a new error has occured.
+	 *
+	 * @param array|null $old Optional. Previously stored return value of error_get_last().
+	 *
+	 * @return string The error message if a new error has occured
+	 *                or an empty string if no (new) errors have occured..
+	 */
+	protected function getChangedErrorMsg($old = null) {
+		$new = error_get_last();
+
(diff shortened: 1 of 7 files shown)


dregad:
@jrfnl could you target the hotfix/5.20 branch and resolve conflicts please ?

jrfnl:
@dregad No problem. On it right now.

jrfnl:
@dregad Ok, done. Cherrypicked to the hotfix branch.

Note:
* I stayed closer to the original code regarding the resetting of `$this->errorMsg` in various places. Feel free to remove those if you think they're not necessary.
* Regarding Sybase - I removed the call to `error_get_last()`/`$php_errormsg` as - without something to compare to - there is no way to tell whether this is a _recent_ error or an unrelated older error.

Testing appreciated & highly recommended as I don't have a test situation with databases of all types set up to do so myself.

dregad:
Many thanks !

>  I don't have a test situation with databases of all types set up to do so myself.

Neither do I, actually. Especially considering that some of these drivers are deprecated / not actively supported.

Issue #193, opened 2016-01-10
Adopt PSR-2 coding style


dregad:
http://www.php-fig.org/psr/psr-2/

Note that PSR-2 requires compliance with PSR-1, which in turns states that *Namespaces and classes MUST follow an "autoloading" PSR*. Consequently, we'll also adopt PSR-4 (see #192).

jrfnl:
@dregad What with PSR-12 expected soon, is PSR-2 still the preference for a first step ?

And if so, would you be open to me having a go ? I'm a bit of PHPCS wiz, so I will probably be able to get this set up, fixed & auto-checked via Travis quite easily.

dregad:
@jrfnl many thanks for your offer !

> What with PSR-12 expected soon, is PSR-2 still the preference for a first step ?

I don't know when the new standard will be approved. I'm fine with switching to PSR-2 now, and adopt PSR-12 later on.

> I'm a bit of PHPCS wiz, so I will probably be able to get this set up, fixed & auto-checked via Travis quite easily.

That would be awesome !
If you're the expert, then you'll probably laugh at my initial attempts at implementing PHPCS but anyway you might want to have a look (https://github.com/dregad/ADOdb/commits/psr2).

> initially namespaces/autoloading can just be excluded

Indeed, but that's still something that needs to be done eventually (#192).

The only concern with the adoption of a coding standard, is that the changes will generate a lot of churn, and make existing PRs unmergeable. So before you get started, I should probably have a look at what should be merged and do that first, what do you think ?

jrfnl:
> The only concern with the adoption of a coding standard, is that the changes will generate a lot of churn, and make existing PRs unmergeable. So before you get started, I should probably have a look at what should be merged and do that first, what do you think ?

That would certainly be advisable. At the same time, I noticed most PRs have been open for quite a while, so I can imagine there may be merge conflicts already.
GH also offers a "rebase & merge" option, so that could possibly be a good alternative.

It's more a case of _when will you have time to look over & merge those PRs_ ? And will you have time to look over & merge the CS PR(s) relatively quickly to prevent that one getting into a conflict state ? which would happen fairly quickly.

On that note: would you prefer one big PR with separate commits to make reviewing easier, or would you prefer small PRs which would follow each other ? Most of those would need to be pulled one at the time as they will most likely have merge conflicts otherwise.

jrfnl:
*grin* Just did a little test run... this will be fun :sunglasses: 
```
[...]

PR #411, opened 2018-04-03
#192 - Add autoload and extract Exception

codisart:
I committed the vendor folder to keep the same installation process.
I added some tests to prove nothing is broken.


codisart:
Hello @mnewnham,

This PR is a proposal to begin the work needed for #192 and #193. 
I think it would be very hard to do this migration without using composer.

mnewnham:
Although I was really adding a vague suggestion when I mentioned about not using composer, I would say that **having** to use it is probably a non-starter. You would get no push-back from me on it's greatness, but for those of us who work in large-scale, commercial software houses that depend on masses of legacy code interactions, integrating composer into them would be a complete nightmare.  The other observation is that I've never come across a piece of software where the optimal implementation was composer that couldn't be installed with out it.

dregad:
With regards to not using composer, for the time being (and foreseeable future too) we're still offering ADOdb for download and standalone installation, via the SourceForge downloads site. 

I see no reason to force using Composer at all.

codisart:
@dregad 
I propose to simplify the autoloader generated by composer.
And maybe removing all mentions of composer. 

@mnewnham 
I'm sorry but I didn't understand what did you mean by :

dregad:
> maybe removing all mentions of composer

That would be too much, I think.

IMO, Composer would be the standard, default way to use the library in the future, but we need to provide a fallback way of using the library to accomodate usage scenarios like the one Mark described. This could indeed be a simplified autoloader script derived from the default Composer one.

codisart:
Hello @dregad 

I took some time to think about a solution to simplify the composer autoloader. But I must say that the composer code is quite complex and I did not find a good solution.
Furthermore I am not sure that this simplification is a good idea anymore.

By using the real code of composer, we are sure that no bug is introduced in the autoloading process. Except the ones from composer itself, but I think that a project as big as this one should be a good garanty of quality. The only downside of this current solution IMO is the autoloading will need to be updated to follow the composer history. 

About the way the adobd project should be installed, my proposal changes nothing for now. And it allows to change this situation more easily in the future if it is needed.

What do you think ?

gggeek:
In case it might be of help: I faced a similar problem with the phpxmlrpc/phpxmlrpc package, which also dates to the dark ages of php 3. It is a smaller project than adodb in terms of classes and complexity of loading (ie. no drivers or plugins involved), but the basic issues were the same.

My approach:
- rename existing classes, use proper namespaced class names and one-class-per-file (PRS-4)
- allow young whippersnappers to use the lib via composer fullstop
- keep the old .inc file available. in that file, classes are defined which extend the new classes while preserving the old class names
- methods which you want renamed can be modified so in teh new classes, and old-names methods kept in the .inc file, which call the new ones
- the .inc file uses include_once calls to load the pssr4 classes without going through a working autoloader config
- add to the lib a custom autoloader which can be used by anyone wanting to use the psr4 classes but bypass composer

See:
https://github.com/gggeek/phpxmlrpc/blob/master/doc/api_changes_v4.md
https://github.com/gggeek/phpxmlrpc/blob/master/lib/xmlrpc.inc
https://github.com/gggeek/phpxmlrpc/blob/master/src/Autoloader.php

PR #414, opened 2018-04-04
Add phpunit dependency

codisart:


dregad:
Thanks for your PR, but quite frankly I don't see the point of adding the composer dependency before we have actual PHPUnit tests to run...

PR #416, opened 2018-04-09
SSL for mysqli connections

colinmo:
MySQL has the ability to have secured connections via a certificate/ key/ etc. This is quite simple in Azure to configure. However ADODB's connect function inits-and-connects in a single function call, and the SSL trigger needs to happen between the two if it's not part of the my.cnf. I'd like to propose an extension to the adodb-mysqli.inc.php

### DIFF
--- a/drivers/adodb-mysqli.inc.php
+++ b/drivers/adodb-mysqli.inc.php
@@ -59,11 +59,11 @@ class ADODB_mysqli extends ADOConnection {
 	var $optionFlags = array(array(MYSQLI_READ_DEFAULT_GROUP,0));
 	var $arrayClass = 'ADORecordSet_array_mysqli';
 	var $multiQuery = false;
-	var $ssl_key = false;
-	var $ssl_cer = false;
-	var $ssl_ca = false;
-	var $ssl_capath = false;
-	var $ssl_cipher = false;
+	var $ssl_key = null;
+	var $ssl_cer = null;
+	var $ssl_ca = null;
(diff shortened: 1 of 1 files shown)


drivers/adodb-mysqli.inc.php
```diff
@@ -120,6 +125,11 @@ function _connect($argHostname = NULL,
 		//http ://php.net/manual/en/mysqli.persistconns.php
 		if ($persist && PHP_VERSION > 5.2 && strncmp($argHostname,'p:',2) != 0) $argHostname = 'p:'.$argHostname;
 
+		// SSL Connections for MySQLI
+		if ($this->ssl_key || $this->ssl_cer || $this->ssl_ca || $this->ssl_ccapath || $this->ca_cipher) {
```
peterdd: ssl_ccapath? typo?

colinmo: 100% correct, thank you peterdd. Repaired.

peterdd:
@mnewnham Have you seen my comment about $ssl_cer --> $ssl_cert ? IMHO should be changed before release; saw your change at http://adodb.org/dokuwiki/doku.php?id=v5:database:mysql

dregad:
> should be changed before release

Done.

mnewnham:
does the documentation need updating?

peterdd:
I modified the example at http://adodb.org/dokuwiki/doku.php?id=v5:database:mysql now.

PR #423, opened 2018-06-07
making the count wrapper sqlsrver compliant

ladams1776:
We use ADODB for Oracle and MSSQL, but the count wrapping is broken for MSSQL. MSSQL requires an alias at the end and the way we fall into the conditional flow is forcing us to hack the adodb-lib.inc.php.

I hope this is good enough to get it into master, but if not please advise me on how I can accommodate the project needs.

Thank you,

-Larry Adams

### DIFF
--- a/adodb-lib.inc.php
+++ b/adodb-lib.inc.php
@@ -418,6 +418,7 @@ function _adodb_getcount(&$zthis, $sql,$inputarr=false,$secs2cache=0)
 
         } else if (strncmp($zthis->databaseType,'postgres',8) == 0
             || strncmp($zthis->databaseType,'mysql',5) == 0
+	    || strncmp($zthis->databaseType,'mssql',5) == 0
             || strncmp($zthis->dsnType,'sqlsrv',5) == 0
         ){
 		    $rewritesql = "SELECT COUNT(*) FROM ($rewritesql) _ADODB_ALIAS_";
--- a/adodb-lib.inc.php
+++ b/adodb-lib.inc.php
@@ -416,15 +416,13 @@ function _adodb_getcount(&$zthis, $sql,$inputarr=false,$secs2cache=0)
 			} else
 				$rewritesql = "SELECT COUNT(*) FROM (".$rewritesql.")";
 
-		} else if (strncmp($zthis->databaseType,'postgres',8) == 0
-			|| strncmp($zthis->databaseType,'mysql',5) == 0
-			|| strncmp($zthis->databaseType,'mssql',5) == 0
-		) {
-			$rewritesql = "SELECT COUNT(*) FROM ($rewritesql) _ADODB_ALIAS_";
-		} else {
-		    // SqlSrver compliant - Larry Adams on 6/7/2018
-			$rewritesql = "SELECT COUNT(*) FROM ($rewritesql) _ADODB_ALIAS_";
-		}
+        } else if (strncmp($zthis->databaseType,'postgres',8) == 0
+            || strncmp($zthis->databaseType,'mysql',5) == 0
+            || strncmp($zthis->databaseType,'mssql',5) == 0)  {
+            $rewritesql = "SELECT COUNT(*) FROM ($rewritesql) _ADODB_ALIAS_";
+        } else {
+            $rewritesql = "SELECT COUNT(*) FROM ($rewritesql)";
+        }
 	} else {
 		// now replace SELECT ... FROM with SELECT COUNT(*) FROM
 		if ( strpos($sql, '_ADODB_COUNT') !== FALSE ) {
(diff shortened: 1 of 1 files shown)


mnewnham:
We shouldn't be breaking the default (unaliased) behavior here. The function already has an elseif to support databases that support aliasing, so try this change instead:
````php
} else if (strncmp($zthis->databaseType,'postgres',8) == 0 
          || strncmp($zthis->databaseType,'mysql',5) == 0 
          || strncmp($zthis->databaseType,'mssql',5) == 0)  {
	$rewritesql = "SELECT COUNT(*) FROM ($rewritesql) _ADODB_ALIAS_";
} else {
	$rewritesql = "SELECT COUNT(*) FROM ($rewritesql)";
}
````
This will allow support for both mssql and mssqlnative drivers

**Comment by ladams1776** [comment]
Ok. I made the change. I am not sure if I need to do anything else. do I need to resubmit the pull request?

**Comment by dregad** [comment]
@ladams1776 no need to resubmit the PR.
@mnewnham let me know if this is OK to merge after the correction.

**Comment by ladams1776** [comment]
Well... I am not sure if this will work. In the SS you can see that we are set to PDO, but the dsnType actually informs us that we are using Sqlsrv, could we adjust the conditional to catch that instead?

image

I will change it to something like this, if you guys approve it:

} else if (strncmp($zthis->databaseType,'postgres',8) == 0 
          || strncmp($zthis->databaseType,'mysql',5) == 0 
          || strncmp($zthis->dsnType,'sqlsrv',5) == 0)  {
	$rewritesql = "SELECT COUNT(*) FROM ($rewritesql) _ADODB_ALIAS_";
} else {
	$rewritesql = "SELECT COUNT(*) FROM ($rewritesql)";
}

**Comment by ladams1776** [comment]
I just updated the branch

**Comment by mnewnham** [comment]
Are we sure that this is not an issue with the PDO driver rather than sql server? If you are unable to test with a native mode driver, could you post a code snippet that we could look at?

**Comment by marcelto** [comment]
@mnewnham this is definitely as issue with the allowed sytnax within SqlSrv.  This can be seen by running these queries directly within any SqlSrv database.

```
--with alias, works
select count(*) from (select 1 as 'a',2 as 'b' ,3 as 'c')  x

--without alias, fails
select count(*) from (select 1 as 'a',2 as 'b' ,3 as 'c')
```

**Comment by marcelto** [comment]
Sorry @mnewnham, I think misunderstood your question.

File: adodb-lib.inc.php [hunk 1]
```diff
@@ -416,14 +416,15 @@ function _adodb_getcount(&$zthis, $sql,$inputarr=false,$secs2cache=0)
...
-			|| strncmp($zthis->databaseType,'mssql',5) == 0
-		) {
-			$rewritesql = "SELECT COUNT(*) FROM ($rewritesql) _ADODB_ALIAS_";
-		} else {
-			$rewritesql = "SELECT COUNT(*) FROM ($rewritesql)";
-		}
+        } else if (strncmp($zthis->databaseType,'postgres',8) == 0
+            || strncmp($zthis->databaseType,'mysql',5) == 0
+	    || strncmp($zthis->databaseType,'mssql',5) == 0
+            || strncmp($zthis->dsnType,'sqlsrv',5) == 0
```
dregad: @mnewnham if my understanding is correct, this would cover the case of pdo_sqlsrv driver, but what about pdo_mssql ?

ladams1776: I do not have a test case but it ought to, because 'mssql' is in the conditional.

ladams1776:
Additionally, on line 432 we have this line:

			$rewritesql = preg_replace('/^\s*SELECT\s.*\s+FROM\s/Uis','SELECT COUNT(*) FROM ',$sql);

But it does not accomodate T-SQL's syntax for formatted named parameters, because T-SQL starts with DECLARE and SET for those named parameters.

We have found if we adjust line 432 by just removing the caret (^) we can get it to work. Like so:

			$rewritesql = preg_replace('/\s*SELECT\s.*\s+FROM\s/Uis','SELECT COUNT(*) FROM ',$sql);


Is this something we can sneak into the pull request as well?

adodb-lib.inc.php
mnewnham: @dregad , sorry, I keep forgetting about mssql/freetds under unix. I have re-updated the code

dregad:
> Is this something we can sneak into the pull request as well?

@ladams1776 I prefer to keep distinct issues in separate pull requests. I noticed you've opened #429 for this, so that's great.

I'll merge this PR soon.

mnewnham:
I don't understand the need for this. Because of the changes made above, execution of the code in a SQL Server environment will never occur,  so why change it to be T-SQL compliant?

mnewnham:
Thanks for your work on this

PR #429, opened 2018-06-19
removing the caret should allow us to respect the query, but accomoda…

ladams1776:
…te DECLARE/SET syntax that is germain to sqlsrver.


dregad:
@ladams1776 following merge of #423, and based on @mnewnham's note https://github.com/ADOdb/ADOdb/pull/423#issuecomment-399694622

> I don't understand the need for this. Because of the changes made above, execution of the code in a SQL Server environment will never occur, so why change it to be T-SQL compliant?

I'll close this PR. If you believe this change is still required, please explain why.

PR #438, opened 2018-07-24
PHP7.2: string parameters for `assert` are deprecated

nicoder:
in favor of booleans

see http://php.net/manual/en/migration72.deprecated.php#migration72.deprecated.assert-string-arg

### DIFF
--- a/session/adodb-compress-bzip2.php
+++ b/session/adodb-compress-bzip2.php
@@ -41,8 +41,8 @@ class ADODB_Compress_Bzip2 {
 	/**
 	 */
 	function setBlockSize($block_size) {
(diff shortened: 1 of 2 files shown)


mnewnham:
Can you confirm that there is no behavior change when this is used with earlier versions of PHP?

mnewnham:
Ok, Im going to pull this into 5.21 master, but not into the 5.20 hotfix. That would mean that the 5.20 would not be 7.3 compatible

mnewnham:
As I did not receive any further comments from you on this subject I am proceeding with my suggested methodology. In fact, this would imply that Version 5.20 would not be 7.2 compatible

PR #444, opened 2018-08-06
server.php: fix example query, change function case, mysqli for demo

peterdd:


dregad:
The commit obviously does more than the message says.

Please resubmit as distinct commits for each change (mysql -> mysqli, change functions case), or at least update commit message to reflect the actual changes to the code.

peterdd:
Sorry, the changes made were so cosmetic I skipped describing each.

dregad:
> the idea of an API that provides a "TRUNCATE TABLE" capability frankly makes my skin crawl. @dregad , I would prefer to see this whole feature killed from the system. I'm surprised there hasn't been a CVE alert about it.
> 
> If,however you want to keep it, I see no reason not to merge

I agree it's a dangerous feature. I have no idea how much actual use it gets in the field (I wouldn't use it myself). Should we mark it as deprecated for now (5.21), and kill it in the next release (5.22) ?

peterdd:
Well, what a setup of a server.php can actually do with a database connection depends on the db user and his db permission you setup in the server.php. If that user has only a select permission for a db table, no 'dangerous' operation can be done.

I played a while ago with it and made some multiple servers, multiple databases tests and after some fiddling it was fun.

```
user <==> client.php 
                                   <===> server1_csv.php
                                   <===> server1_mysql.php
                                   <===> server1_pg10.php
                                   <===> server1_sqlite.php
                                   <===> server2_pg12.php
                                   <===> server3_firebird.php
```

Maybe just disable the server.php by default. It has probably no CVE yet because the default server.php does not work as nobody would setup a db connection like that.

Sure I would currently only use within internal networks where client.php and server.php trust eachother and man-in-the-middle is not possible/irrelevant.

Also a 'server.php' doesn't necessarly need to be reachable by enduser. A client.php can implement user auth, csrf protection and so on.

mnewnham:
@peterdd , if you want to take ownership of this, then that's fine by me. But for this to continue in the product, I'd like it to have substantial security enhancements,

- Perhaps an auth key passed from the client
- FILTER_VALIDATE applied to input variables
- Perhaps a user-definable list of acceptable SQL statements.

@dregad , I think your suggestion of deprecation and removal is appropriate, but if @peterdd wants to own it and fix it, I'd accept it being re-added.

dregad:
I added a deprecation notice and a security warning to the [Proxy Server documentation](https://adodb.org/dokuwiki/doku.php?id=v5:proxy:proxy_index) page.

As it stands, this PR does not cause any harm so I'll merge it. 

@peterdd it's up to you if you want to improve it as per Mark's suggestions.

gggeek:
About the security risk of leaving the server.php file in the main distribution: what about adding it as export-ignore to the .gitattributes file?
This way it would not be included in common installs of the adodb lib - when as required as dependency.
It would still be available to people who either clone the project or install it using composer's `prefer-source` option.

PR #448, opened 2018-08-10
Sqlsrv identity

gohanman:
With the `sqlsrv` module the `identitySQL` method (`SELECT SCOPE_IDENTITY()`) frequently (always?) returns `NULL`. The existing implementation [accounts for this](https://github.com/ADOdb/ADOdb/blob/master/drivers/adodb-mssqlnative.inc.php#L614-L619) in `_query()` by checking the statement for an `INSERT` and appending the `identitySQL` query to it. Running an `INSERT` query actually returns the result of `SELECT SCOPE_IDENTITY()` instead of the result of the `INSERT` query.

This works fine if you're aware of this behavior and examine the result of the insert query instead of calling `Insert_ID()`, but it means code that works with the `mssql` driver won't work the same way with the `mssqlnative` driver and I think that's fixable.

Keeping a copy of the last insert result from `_query()` and using it in `_insertid()` when available makes the change in query behavior transparent to the caller and they can continue using `Insert_ID()`.

### DIFF
--- a/drivers/adodb-mssqlnative.inc.php
+++ b/drivers/adodb-mssqlnative.inc.php
@@ -642,7 +642,7 @@ class ADODB_mssqlnative extends ADOConnection {
 		if(!$rez)
 			$rez = false;
 		if ($insert)
-			$this->_lastInsertResult = $res;
+			$this->_lastInsertResult = $rez;
 
 		return $rez;
 	}
--- a/drivers/adodb-mssqlnative.inc.php
+++ b/drivers/adodb-mssqlnative.inc.php
@@ -131,6 +131,8 @@ class ADODB_mssqlnative extends ADOConnection {
 	var $sequences = false;
 	var $mssql_version = '';
 
+	var $_lastInsertResult = false;
+
 	function __construct()
 	{
 		if ($this->debug) {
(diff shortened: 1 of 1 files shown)


gohanman:
I'd say I'm trying to change the behavior of the mssqlnative driver to duplicate everything else. I do want it to behave like the mssql driver, but I also want it to behave like the mysql drivers and the postgres drivers and the sqlite drivers. This code should do the same thing across all of them:

```php
$con->Execute("INSERT INTO someTable (someColumn) VALUES (1);
$rowID = $con->Insert_ID();
```

As an abstraction layer ADOdb should make make this fairly basic situation "just work" without the caller needing to know anything about the underlying implementation. With mssqlnative the abstraction is leaking and you have to write different code to work with how the specific driver is implemented.

mnewnham:
This has been bugging me for a while. I have finally remembered what it is. This issue **Insert_ID() returns an empty result when using prepared queries #5** identified a serious problem, in that we could only access the insert_id of a record inserted by a prepared query by appending the insert code onto the end of the statement.

Although this is now 6 years old, I have not been made aware of any alternatives to this method, and until we can, this code must stay as is.

If you want to refactor your code and re-submit as a non-default behavior, feel free to do that. We appreciate the time you've spent on this.

gohanman:
Is there a project preferred approach to non-default behavior - optional argument to Insert_ID, a mode-setting method on the connection class, something else?

mnewnham:
Something as simple as
````
public $delayInsertIdFetch = false

// in your code
If ($this->delayInsertId) then ---->modify the current behavior to your methodology
````
will be fine

PR #458, opened 2018-10-12
Simplified pdo_mysql driver SelectLimit Method

s0fie:
* Removed magic number
* Removed code duplication
* Removed long-dead link to what I can only assume is an explanation as to why a magic number is needed (it isn't)
* Reversed $nrows logic to only concatenate a LIMIT statement to existing sql if a limit is set when the method is called.

### DIFF
--- a/drivers/adodb-pdo_mysql.inc.php
+++ b/drivers/adodb-pdo_mysql.inc.php
@@ -204,15 +204,15 @@ class ADODB_pdo_mysql extends ADODB_pdo {
 		$nrows = (int) $nrows;
 		$offset = (int) $offset;		
 		$offsetStr =($offset>=0) ? "$offset," : '';
-		// jason judge, see http://phplens.com/lens/lensforum/msgs.php?id=9220
-		if ($nrows < 0) {
-			$nrows = '18446744073709551615';
+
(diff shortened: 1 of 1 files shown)


dregad:
> long-dead link to what I can only assume is an explanation as to why a magic number is needed (it isn't)

Here's the discussion from the old support forum, from the web archive :
https://web.archive.org/web/20051217185255/http://phplens.com/lens/lensforum/msgs.php?id=9220

The use of a "magic number" 2^64-1 is actually suggested in [MySQL documentation](https://dev.mysql.com/doc/refman/8.0/en/select.html). 

>  To retrieve all rows from a certain offset up to the end of the result set, you can **use some large number for the second parameter**. This statement retrieves all rows from the 96th row to the last:
> ```
> SELECT * FROM tbl LIMIT 95,18446744073709551615;
> ```

Your code would cause MySQL to choke with ERROR 1064 (42000), if $offset is specified while $nrows is -1, e.g. to retrieve all rows starting from a given row (`selectLimit($sql, -1, 10);

dregad:
@s0fie since you never provided any feedback following my last reply, I'm going to close this PR. Feel free to reopen it.

Issue #378, opened 2017-10-19
include pdo_firebird driver


mawouzo:
since php 7.0.03 the firebird extension can't be used with ADOdb. It woud be good to include the pdo_firebird driver in ADOdb.

gohanman:
I was looking into this and it seems like the existing Interbase extension driver (`drivers/adodb-ibase.inc.php`) works fine and the Interbase extension is available in PHP7. I'm testing against Firebird 2.5.4.

Is there anything `pdo_firebird` can do that `interbase` can't? If not it seems like adding another Firebird driver is just creating extra maintenance work.

mawouzo:
Yes, it is. Please look at this thread:
https://bugs.php.net/bug.php?id=72175

mnewnham:
I've added a PDO driver for Firebird, this is an initial release. If you have the time to help testing, please do

dregad:
@mnewnham not sure why you closed this, the branch with the new driver was not actually merged into master.

PR #473, opened 2019-01-06
Add support for pdo-firebird, see #378, Add Meta extension points

mnewnham:
This change includes the first release of the pdo_firebird driver, as well as extension points for the following  Meta functions within the pdo driver:
 - metaIndexes()
- metaPrimaryKeys()
- metaForeignKeys()
- metaProcedures()

The first release of the pdo-firebird driver is known to have the following limitations:

- Only supports dialect3
- Only tested with PHP7
- Only tested with Firebird 3
- function getRowAssoc() fails because PDO::firebird does not support getColumnMeta()
- function moveFirst() currently fails for unknown reasons

### DIFF
--- a/drivers/adodb-pdo_firebird.inc.php
+++ b/drivers/adodb-pdo_firebird.inc.php
@@ -85,13 +85,10 @@ class ADODB_pdo_firebird extends ADODB_pdo {
 		}
 
 		$retarr = array();
-		//OPN STUFF start
-		$dialect3 = ($this->dialect==3 ? true : false);
-		//OPN STUFF end
+		$dialect3 = $this->dialect == 3;
 		while (!$rs->EOF) { //print_r($rs->fields);
 			$fld = new ADOFieldObject();
 			$fld->name = trim($rs->fields[0]);
-			//OPN STUFF start
 			$this->_ConvertFieldType($fld, $rs->fields[7], $rs->fields[3], $rs->fields[4], $rs->fields[5], $rs->fields[6], $dialect3);
 			if (isset($rs->fields[1]) && $rs->fields[1]) {
 				$fld->not_null = true;
@@ -119,7 +116,6 @@ class ADODB_pdo_firebird extends ADODB_pdo {
 			} else {
 				$fld->sub_type = null;
 			}
-			//OPN STUFF end
 			if ($ADODB_FETCH_MODE == ADODB_FETCH_NUM) $retarr[] = $fld;
 			else $retarr[strtoupper($fld->name)] = $fld;
 
--- a/drivers/adodb-pdo_firebird.inc.php
+++ b/drivers/adodb-pdo_firebird.inc.php
@@ -1,0 +1,423 @@
+<?php
+/**
+ * ADOdb PDO Firebird driver
+ *
+ * @version   v5.21.0-dev  ??-???-2016
+ * @copyright (c) 2019      Damien Regad, Mark Newnham and the ADOdb community
+ *
+ * Released under both BSD license and Lesser GPL library license.
+ * Whenever there is any discrepancy between the two licenses,
+ * the BSD license will take precedence. See License.txt.
+ *
+ * Set tabs to 4 for best viewing.
+ *
+ * Latest version is available at http://adodb.org/
+ *
+ * This version has only been tested on Firebird 3.0 and PHP 7
+ */
+
+class ADODB_pdo_firebird extends ADODB_pdo {
+
+	public $dialect = 3;
+
+	public $metaTablesSQL = "select lower(rdb\$relation_name) from rdb\$relations where rdb\$relation_name not like 'RDB\$%'";
+
+	public $metaColumnsSQL = "select lower(a.rdb\$field_name), a.rdb\$null_flag, a.rdb\$default_source, b.rdb\$field_length, b.rdb\$field_scale, b.rdb\$field_sub_type, b.rdb\$field_precision, b.rdb\$field_type from rdb\$relation_fields a, rdb\$fields b where a.rdb\$field_source = b.rdb\$field_name and a.rdb\$relation_name = '%s' order by a.rdb\$field_position asc";
+
+	var $arrayClass = 'ADORecordSet_array_pdo_firebird';
+
+	function _init($parentDriver)
+	{
+		$this->pdoDriver = $parentDriver;
+		//$parentDriver->_bindInputArray = true;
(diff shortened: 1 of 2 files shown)


dregad:
@mnewnham I have split your original commit to reflect the different work items it covers. 

I kept your original commit in _PDO-Firebird-Driver-original_, for reference.

drivers/adodb-pdo_firebird.inc.php
```diff
@@ -0,0 +1,423 @@
...
+		else
+			$rs = $this->execute($sql,$inputarr);
+
+		return $rs;
+	}
+
+	/**
+	* Sets the appropriate type into the $fld variable
+	*
+	* @param	string		$fld		By reference
```
dregad: Looking at the function call at line 95, it would seem that $fld is an `ADOFieldObject`, not a string

drivers/adodb-pdo_firebird.inc.php
```diff
@@ -0,0 +1,423 @@
...
+	* @param	int			$ftype
+	* @param	int			$flen
+	* @param	int			$fscale
+	* @param	int			$fsubtype
+	* @param	int			$fprecision
+	* @param	bool		$dialect3
```
dregad: Remove $dialect3 param (see earlier comment) ?

drivers/adodb-pdo_firebird.inc.php
```diff
@@ -0,0 +1,423 @@
...
+	*/
+	final private function _convertFieldType(&$fld, $ftype, $flen, $fscale, $fsubtype, $fprecision, $dialect3)
+	{
+		$fscale = abs($fscale);
+		$fld->max_length = $flen;
+		$fld->scale = null;
+		switch($ftype){
+			case 7:
+			case 8:
+				if ($dialect3) {
```
dregad: `if ($this->dialect == 3) {`

drivers/adodb-pdo_firebird.inc.php
```diff
@@ -0,0 +1,423 @@
...
+						$fld->type = 'decimal';
+						$fld->scale = $fscale;
+						$fld->max_length = ($ftype == 7 ? 4 : 9);
+					} else {
+						$fld->type = ($ftype == 7 ? 'smallint' : 'integer');
+					}
+				}
+				break;
+			case 16:
+				if ($dialect3) {
```
dregad: if ($this->dialect == 3) {

drivers/adodb-pdo_firebird.inc.php
```diff
@@ -0,0 +1,423 @@
...
+				if ($fscale !=0) {
+					$fld->type = 'decimal';
+					$fld->max_length = 15;
+					$fld->scale = 5;
+				} else {
+					$fld->type = 'double';
+				}
+				break;
+			case 35:
+				if ($dialect3) {
```
dregad: if ($this->dialect == 3) {

drivers/adodb-pdo_firebird.inc.php
dregad: Also, better use spaces for indentation within PHPDoc blocks

dregad:
@mnewnham waiting for your feedback / update of the PR following my review comments.

drivers/adodb-pdo_firebird.inc.php
```diff
@@ -0,0 +1,423 @@
...
+
+		$ADODB_FETCH_MODE = $save;
+		$false = false;
+		if ($rs === false) {
+			return $false;
+		}
+
+		$retarr = array();
+		//OPN STUFF start
+		$dialect3 = ($this->dialect==3 ? true : false);
```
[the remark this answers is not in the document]
mnewnham: No idea, it was added by the original contributor

drivers/adodb-pdo_firebird.inc.php
mnewnham: See my reply

mnewnham: lets let the fix-docblocks take care of it......

drivers/adodb-pdo_firebird.inc.php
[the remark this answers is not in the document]
mnewnham: >> Ternary operator is not necessary - $dialect3 = $this->dialect == 3; does the same more efficiently - but see later comments.

Interesting, but is it actually a bug?  I wont disagree with you, and my simplistic tests show that we would save approx 0.01 seconds per million iterations.

drivers/adodb-pdo_firebird.inc.php
```diff
@@ -0,0 +1,423 @@
...
+		//OPN STUFF end
+		while (!$rs->EOF) { //print_r($rs->fields);
+			$fld = new ADOFieldObject();
+			$fld->name = trim($rs->fields[0]);
+			//OPN STUFF start
+			$this->_ConvertFieldType($fld, $rs->fields[7], $rs->fields[3], $rs->fields[4], $rs->fields[5], $rs->fields[6], $dialect3);
```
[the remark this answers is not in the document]
dregad: > I'd prefer to keep it code compatible with the original firebird driver,

OK fine.
I still think it's bad design, but we can reconsider this at some other time (in 6.0, if we ever get around to it :wink:)

drivers/adodb-pdo_firebird.inc.php
```diff
@@ -0,0 +1,447 @@
...
+	public function dropSequence($seqname = 'adodbseq')
+	{
+		$seqname = strtoupper($seqname);
+		return $this->Execute("DROP SEQUENCE $seqname");
+	}
+
+
+	public function _affectedrows()
+	{
+		return fbird_affected_rows($this->_transactionID ? $this->_transactionID : $this->_connectionID);
```
dregad: @mnewnham Is this valid ? 

Function is not documented (there is [ibase_affected_rows()](https://www.php.net/manual/en/function.ibase-affected-rows.php))
Shouldn't we simply rely on the parent's ADODB_pdo::_affectedrows() ?

dregad:
Merged in 02192183acaa167d7986e0c143112c725ca8e552

drivers/adodb-pdo_firebird.inc.php
dregad: @mnewnham I merged this, but I realize now that you never answered the above question...

mnewnham: [fbird_affected_rows ](https://www.php.net/manual/en/function.fbird-affected-rows.php )appears to exist now

PR #477, opened 2019-01-10
Removal of session_module_name('user') for PHP 7.2+ database session support.

### LINKED ISSUE #449: adodb-session2 : session_module_name('user') throws an error for PHP 7.2
The `session_module_name('user')` code on line 467 in the `session/adodb-session2` file throws below error for PHP 7.2 on windows platform (not sure about linux):

```php
session_module_name(): Cannot set 'user' save handler by ini_set() or session_module_name() 
```

The same code works in PHP 5.6 on both the Windows and Linux platform.

mattsimpson:
Resolves #449 Removal of `session_module_name('user')` from session related files.

According to the documention:
http://php.net/session_module_name

> 7.2.0 - It is now explicitly forbidden to set the module name to "user". Formerly, this has been silently ignored.

I have tested this removal on PHP 7.1 and 7.2, and it works as expected.

### DIFF
--- a/session/adodb-session.php
+++ b/session/adodb-session.php
@@ -442,7 +442,6 @@ class ADODB_Session {
 	/*!
 	*/
 	function _init() {
-		session_module_name('user');
 		session_set_save_handler(
 			array('ADODB_Session', 'open'),
 			array('ADODB_Session', 'close'),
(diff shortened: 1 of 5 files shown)


dregad:
Fixed via 93f13fd8e92ce7349ad84752c68235b252b88468. Thanks for your contribution !

Issue #476, opened 2019-01-09
PDO driver: add support for $options array in PDO contructor


f7h:
Some driver specific connection options like MYSQL_ATTR_FOUND_ROWS can only be set in PDO::__construct and not later via PDO::setAttribute().

We need the ability to pass an options array to ADOConnection::Connect which is used in ADODB_pdo::_connect as follows `new PDO($argDSN, $argUsername, $argPassword, $options)`.

mnewnham:
Per the [Documentation](http://adodb.org/dokuwiki/doku.php?id=v5:database:pdo), the [setConnectionParameter](http://adodb.org/dokuwiki/doku.php?id=v5:reference:connection:setconnectionparameter) option is supported under PDO from version 5.21. Whilst the is currently no release,  you can simply drop in the code from a zip file of the master branch

tkhamez:
I didn't see ADOConnection::setConnectionParameter, but it uses PDO::setAttribute.

As I said, there are options that cannot be changed.

Example:

This works:
```
new PDO($argDSN, $argUsername, $argPassword, [\PDO::MYSQL_ATTR_FOUND_ROWS => true]) 
```

This doesn't work:
```
$pdo = new PDO($argDSN, $argUsername, $argPassword);
$pdo->setAttribute(\PDO::MYSQL_ATTR_FOUND_ROWS, true);
```

I tested this with PHP 5.6 and 7.2.

The latter even shows a warning with PHP 7.2, MySQL 5.7: `Warning: PDO::getAttribute(): SQLSTATE[IM001]: Driver does not support this function: driver does not support that attribute in ...`

f7h:
I didn't see ADOConnection::setConnectionParameter, but it uses PDO::setAttribute.

As I said, there are options that cannot be changed.

Example:

This works:
```
new PDO($argDSN, $argUsername, $argPassword, [\PDO::MYSQL_ATTR_FOUND_ROWS => true]) 
```

This doesn't work:
```
$pdo = new PDO($argDSN, $argUsername, $argPassword);
$pdo->setAttribute(\PDO::MYSQL_ATTR_FOUND_ROWS, true);
```

I tested this with PHP 5.6 and 7.2.

The latter even shows a warning with PHP 7.2, MySQL 5.7: `Warning: PDO::getAttribute(): SQLSTATE[IM001]: Driver does not support this function: driver does not support that attribute in ...`

mnewnham:
[...]

PR #479, opened 2019-01-13
PDO: add support for connection options in constructor.

f7h:
#476

### DIFF
--- a/adodb.inc.php
+++ b/adodb.inc.php
@@ -701,17 +701,21 @@ if (!defined('_ADODB_LAYER')) {
 			$this->password = 'not stored'; // not stored for security reasons
 		}
 		if ($argDatabaseName != "") {
-			$this->database = $argDatabaseName;
+			if (is_array($argDatabaseName)) {
+				$this->database = isset($argDatabaseName['database']) ? $argDatabaseName['database'] : '';
+			} else {
+				$this->database = $argDatabaseName;
+			}
 		}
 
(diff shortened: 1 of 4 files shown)


f7h:
1. yeah, I don't really like that myself, but I tried to follow your suggestion, see https://github.com/ADOdb/ADOdb/issues/476#issuecomment-453777872.
Maybe I can come up with another approach, but I don't promise anything.

2. As explained in the ticket, there are connection parameters that you cannot set with the PDO setAttribute method later. See also the test I included.

mnewnham:
Yes I did say that, didn't I. Apologies for not reading my own comments.  I've also checked out stackoverflow and I understand the issue. Rethinking setConnectionParameter, and having shamelessly copied an example from there, could I perhaps suggest this:

````php
$opt  = array(
    PDO::MYSQL_ATTR_FOUND_ROWS   => TRUE,
    PDO::ATTR_ERRMODE            => PDO::ERRMODE_EXCEPTION,
    PDO::ATTR_DEFAULT_FETCH_MODE => PDO::FETCH_ASSOC,
);

$db->setConnectionParameter('ADODB',$opt);
````
Then in the connect function,

````php
if (array_key_exists('ADODB',$this->connectionParameters)
{
    $connectOpts = $this->connectionParameters['ADODB'];
    unset($this->connectionParameters['ADODB']);
}
// continue on to connect.....
````

mnewnham:
See #650

Issue #292, opened 2016-11-12
ADODB_DataDict._GenFields: Unconditional remove of nullability and default on blobs


obmsch:
In 'adodb-datadict'  inside _GenFields:

line 729
if ($ty == 'X' || $ty == 'X2' || $ty == 'B') $fnotnull = false; // some blob types do not accept nulls [cf #291]

line 733-734
//some databases do not allow blobs to have defaults
if ($ty == 'X') $fdefault = false;

Thats completely ignoring the capabilites of any db to honor some dbs not implementing that feature. I am looking for a better approach(My knowledge of php and its inner workings is really limited. So this is more conceptional than a real implementation).

Add members (for capabilities) to the base class ADODB_DataDict like
$hasNullableBlobs = false;
$hasDefaultOnBlobs = false;

Change the code above to:
if (!$this->hasNullableBlobs && ($ty == 'X' || $ty == 'X2' || $ty == 'B')) $fnotnull = ...
and
if (!$this->hasDefaultOnBlobs &&   $ty == 'X') $fdefault = false;

Thats a noop in behaviour of ADOdb.

But lets say a db supports NULL in blobs, setting $hasNullableBlobs = true; in its datadict should do the trick.

obmsch:
Missed an outstanding pull request pr#118, adressing this issue.

Reconsidering my first thougts and the approach taken in the pull request, I would now recommend to handle this in a more general way in the generic datadict. There might be other field types beside blobs with restrictions too.

And if we agree on:

**Any given schema should be me mapped as precisely as possible to the underlying db.**

then

Either we drop those lines completely and let it to the dbs datadict to handle this, implementing its own _GenFields or take care of it in Add/AlterColumnSQL (caveat: to preserve current behaviour all dbs datadicts have to be adjusted, and at least me don't know what all the supported dbs would require then).

or

a) Extend the generic datadict with something like:

      HasNull($t) {
        return true;
      }
      HasDefault($t) {
        return true;
      }

b) Alter the corresponding lines:

    if (!this->HasNull($ty)) {
       $fnotnull = false;
    }

    if (!this->HasDefault($ty)) {
       $fdefault = false;
    }

c) To preserve current behaviour (and put it to the db to handle) extend all dbs datadicts with:

    HasNull($t) {
      case 'X':
      case 'X2':
      case 'B' : return false;
      default: return true;
    }
      
    HasDefault($t) {
      case 'X' : return false;
      default: return true;
    }

[...]

PR #480, opened 2019-01-13
Issue 478

obmsch:
Extend Control of BLOB data dictionary feature see #292
Add 'XL' to the blob checks in _GenFields.


dregad:
Merged in 127dd199cad830c88f4fcc39ad4cd8cc760ee5ba

PR #484, opened 2019-01-25
Fixed: Connecting to non-standard port with mysqlnd.

prasad83:
Connecting to MySQL server with port other than 3306 was not working when mysqlnd extension was enabled and mysqli type of driver was used. Having it part of hostname got it working.


jane-lyndon:
uncommenting the line above this change should fix this instead?

dregad:
@prasad83 as you have not responded to @mnewnham's questions, I'm going to close this as won't fix for now, but feel free to reopen / comment.

Issue #488, opened 2019-02-11
Improve/fix function docblocks to be IDE "friendly"


wonderboyjon:
While I'm hoping this request would aid other IDEs, I will say that I'm writing this request as someone who uses PHPStorm (in case that matters).

One of the main things the IDE does to aid development is highlight issues when it thinks that code is attempting to use the value returned from an object/function call in a way that it shouldn't be used, based on the expected return type(s) of that function. For example, trying to use a returned value as an array when the function says it only returns boolean values would be incorrect.

However, due to the way most of the function docblocks are currently written, the IDE flags up warnings all over the place and cannot provide useful hints to developers.

As an example of what I mean:
```
adodb.inc.php

       /**
	* Fetch a row, returning false if no more rows.
	* This is PEAR DB compat mode.
	*
	* @return false or array containing the current record
	*/
	function FetchRow() {
		if ($this->EOF) {
			return false;
		}
		$arr = $this->fields;
		$this->_currentRow++;
		if (!$this->_fetch()) {
			$this->EOF = true;
		}
		return $arr;
	}
```
When written like this, as far as the IDE is concerned, this function can only ever return `false`, and displays a warning when you try to treat the returned value as an array. 

The kind annotation I would expect it to have would be:

```
@return false|array The current record, or false if there is no current record.
```

The IDE then knows exactly which two types the function can return, and code issue highlighting and hinting works as expected.

Obviously a lot of dockblocks would need changing to accommodate this style, so maybe there's a better way to break this issue down, but it would be a big help and mean I no longer have to go in and modify the blocks manually every time there's an update.

dregad:
@wonderboyjon we are aware of this issue of course, it has been reported before (see #136, #138). 

The main problem is that, as you can imagine, this is no small undertaking, there are dozens of classes and hundreds of functions that need to be documented, not to mention file-level docblocks... And I don't have the time or motivation to go through it all myself. 

If you're willing to do the legwork and submit a pull request with phpdoc blocks fixes, we'll gladly review & merge them.

wonderboyjon:
I can certainly give it a go. Would your preference be one massive PR with "everything" done or break things down a bit (e.g. by class) and submit multiple PRs?

wonderboyjon:
I'll start with #490 and then hopefully do more once I'm happy that you're happy with the way it's being done.

dregad:
[...]

PR #490, opened 2019-02-12
Adding and improving function documentation for `ADODB_mysqli`

wonderboyjon:
Related to #488

### DIFF
--- a/drivers/adodb-mysqli.inc.php
+++ b/drivers/adodb-mysqli.inc.php
@@ -72,7 +72,14 @@ class ADODB_mysqli extends ADOConnection {
 	private $usePreparedStatement    = false;
 	private $useLastInsertStatement  = false;
 
-	function SetTransactionMode( $transaction_mode )
+    /**
+     * Set the transaction mode.
+     *
+     * @param string $transaction_mode The transaction mode to set.
+     *
+     * @return void
+     */
+    function SetTransactionMode($transaction_mode)
 	{
 		$this->_transmode = $transaction_mode;
 		if (empty($transaction_mode)) {
--- a/drivers/adodb-mysqli.inc.php
+++ b/drivers/adodb-mysqli.inc.php
@@ -83,14 +83,25 @@ class ADODB_mysqli extends ADOConnection {
 		$this->Execute("SET SESSION TRANSACTION ".$transaction_mode);
 	}
 
-	// returns true or false
-	// To add: parameter int $port,
-	//         parameter string $socket
-	function _connect($argHostname = NULL,
-				$argUsername = NULL,
-				$argPassword = NULL,
-				$argDatabasename = NULL, $persist=false)
-	{
+    /**
+     * Connect to a database.
+     * TODO add: parameter int $port, parameter string $socket
+     *
+     * @param string|null $argHostname (Optional) The host to connect to.
+     * @param string|null $argUsername (Optional) The username to connect as.
+     * @param string|null $argPassword (Optional) The password to connect with.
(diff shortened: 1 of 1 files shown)


drivers/adodb-mysqli.inc.php
```diff
@@ -83,14 +90,25 @@ function SetTransactionMode( $transaction_mode )
...
-	// To add: parameter int $port,
-	//         parameter string $socket
-	function _connect($argHostname = NULL,
-				$argUsername = NULL,
-				$argPassword = NULL,
-				$argDatabasename = NULL, $persist=false)
-	{
+    /**
+     * Connect to a database.
+     * TODO add: parameter int $port, parameter string $socket
```
dregad: Use `@todo` tag

drivers/adodb-mysqli.inc.php
```diff
@@ -153,28 +171,64 @@ function _connect($argHostname = NULL,
...
 
-	function IfNull( $field, $ifNull )
-	{
-		return " IFNULL($field, $ifNull) "; // if MySQL
-	}
-
-	// do not use $ADODB_COUNTRECS
-	function GetOne($sql,$inputarr=false)
+    /**
+     * Returns a stringing containing a MySQL IFNULL() function call for the provided values.
```
dregad: stringing -> string ?

drivers/adodb-mysqli.inc.php
```diff
@@ -153,28 +171,64 @@ function _connect($argHostname = NULL,
...
+     *
+     * @return string
+     */
+    function IfNull($field, $ifNull)
+    {
+        return " IFNULL($field, $ifNull) ";
+    }
+
+    /**
+     * Do not use $ADODB_COUNTRECS.
```
dregad: I know you just used whatever was there, but this comment is very confusing. Might as well remove it.

drivers/adodb-mysqli.inc.php
```diff
@@ -323,7 +424,13 @@ function CreateSequence($seqname='adodbseq',$startID=1)
 		return $this->Execute(sprintf($this->_genSeq2SQL,$seqname,$startID-1));
 	}
 
-	function GenID($seqname='adodbseq',$startID=1)
+    /**
+     * @param string $seqname
```
dregad: Missing description

drivers/adodb-mysqli.inc.php
```diff
@@ -946,7 +1158,7 @@ function SetCharSet($charset_name)
 }
 
 /*--------------------------------------------------------------------------------------
-	 Class Name: Recordset
+	 Class Name: ADORecordSet_mysqli
 --------------------------------------------------------------------------------------*/
```
dregad: This should be a PHPDoc block too; and consider adding one for class _ADODB_mysqli_ too.
```
/**
 * Class ADORecordSet_mysqli
 */
```

dregad:
As a side note, when submitting pull requests I strongly advise not to use your _master_ branch, but create a dedicated feature branch instead.

wonderboyjon:
Yeh I realised shortly afterwards that not creating my own branches was a bad idea. Will rectify that after this request.

I've gone through and made all the changes suggested. I tried to marry up the function calls to the website documentation where possible.

[Fields()](https://adodb.org/dokuwiki/doku.php?id=v5:reference:recordset:fields) was an odd one though, as the website documentation makes it clear that the function parameter is expected to be an `int`. However the code for that function in `ADORecordSet_mysqli` very clearly expects a `string`, so I have chosen to respect what the code does and documented it as a string rather than following the website documentation. If this is incorrect let me know.

PR #493, opened 2019-03-06
Update adodb-mysqli.inc.php

anoopnaik:
fix GenId() not returning next sequence value intermittently

### DIFF
--- a/drivers/adodb-mysqli.inc.php
+++ b/drivers/adodb-mysqli.inc.php
@@ -342,6 +342,12 @@ class ADODB_mysqli extends ADOConnection {
 
 		if ($rs) {
 			$this->genID = mysqli_insert_id($this->_connectionID);
+			if ($this->genID == 0)
+			{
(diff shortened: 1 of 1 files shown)


drivers/adodb-mysqli.inc.php
```diff
@@ -342,6 +342,12 @@ function GenID($seqname='adodbseq',$startID=1)
 
 		if ($rs) {
 			$this->genID = mysqli_insert_id($this->_connectionID);
+			if ($this->genID == 0)
+			{
+				$getnext = sprintf("select LAST_INSERT_ID() from %s;",$seqname);
+				$rs = @$this->Execute($getnext);
```
dregad: What happens if execute() fails ?

dregad: @anoopnaik you have not responded to my comment. 

Suppressing errors can cause problems, and while there may be valid reasons to do so, they need to be documented.

Issue #491, opened 2019-02-15
Warning: mysqli_free_result(): Couldn't fetch mysqli_result is shown in function _close for empty mysqli_result


aga112b:
Hi,
Unfortunately solution added by ivanbogomoloff in issue #463 causes that there is displayed warning:
`Warning: mysqli_free_result(): Couldn't fetch mysqli_result`
when the function ADORecordSet_mysqli->_close( ) is run.
I don't know why in function `_close`, `$this->_queryID` is instance of mysqli_result but couldn't be fetched. I found out that warning occurs when mysqli_result is empty (when I cast it to array and var_dump it looks:

```
array (size=5)
  'current_field' => null
  'field_count' => null
  'lengths' => null
  'num_rows' => null
  'type' => null
```
I've tried to find solution how to change mentioned code to not caused warnings but still resolve problem with displaying error

```
'LOGSQL Insert Failed: insert into adodb_logsql (created,sql0,sql1,params,tracer,timer) values( NOW(),?,?,?,?,?)
Commands out of sync; you can't run this command now'.
```
I found out that if in adodb-mysqli.inc.php in function `_close` we use
`if($this->_queryID instanceof mysqli_result && (!empty(implode('',(array)$this->_queryID)))) {`
instead
`if($this->_queryID instanceof mysqli_result) {`
then warning will not be displayed.

dregad:
@aga112b could you please provide a code snippet to reproduce the problem ?

aga112b:
when you set `$ADODB_PERF_MIN` to `0.0001`
and run code:
```
$sql = 'SELECT * FROM table ORDER BY name ASC, id ASC';
$result_array = $db->SelectLimit($sql, 1);
$result_array = $result_array->FetchRow();
```
where `$db` is a instance of `ADODB_mysqli`
then you should see warning 
`Warning: mysqli_free_result(): Couldn't fetch mysqli_result`

dregad:
Sorry, but I'm not able to reproduce the error with the provided information.

I'll close this and the associated PR #494 for now, but feel free to reopen if the problem persists and you can provide additional input.

dolphinotaku:
Hi,
I have the same issue. it is weird, it was not happening every time, 3 times throw warning out of 10 F5 

Error:
[...]

PR #494, opened 2019-03-13
warning for empty mysqli_result

aga112b:
Warning is shown in function _close for empty mysqli_result (see #491)


dregad:
Closed without merge, see https://github.com/ADOdb/ADOdb/issues/491#issuecomment-570801655

PR #496, opened 2019-03-21
PDO: add support for dblib

Markomas:
This includes first release for pdo_dblib driver.

Corectly supports meta table info (succesfully extracts columns, primary keys etc.). 
_query(), tries to convert strings to integers, for prepared statments (workaround, because PDO treats all variables as string).

### DIFF
--- a/drivers/adodb-pdo_dblib.inc.php
+++ b/drivers/adodb-pdo_dblib.inc.php
@@ -2,7 +2,6 @@
 
 class ADODB_pdo_dblib extends ADODB_pdo
 {
-
     var $hasTop = 'top';
     var $sysDate = 'convert(datetime,convert(char,GetDate(),102),102)';
     var $sysTimeStamp = 'GetDate()';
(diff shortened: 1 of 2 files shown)


drivers/adodb-pdo_dblib.inc.php
```diff
@@ -0,0 +1,176 @@
...
+
+    var $metaColumnsSQL = "SELECT c.NAME, OBJECT_NAME(c.id) as tbl_name, c.length, c.isnullable, c.status, ( CASE WHEN c.xusertype=61 THEN 0 ELSE c.xprec END), ( CASE WHEN c.xusertype=61 THEN 0 ELSE c.xscale END), ISNULL(i.is_primary_key, 0) as primary_key FROM   syscolumns c INNER JOIN systypes t ON t.xusertype=c.xusertype INNER JOIN sysobjects o ON o.id=c.id LEFT JOIN sys.index_columns ic ON ic.object_id = c.id AND c.colid = ic.column_id LEFT JOIN sys.indexes i ON i.object_id = ic.object_id AND i.index_id = ic.index_id WHERE c.id = OBJECT_ID('%s') ORDER by c.colid";
+
+
```
jane-lyndon: this has additional \r\n that not required

drivers/adodb-pdo_dblib.inc.php
```diff
@@ -0,0 +1,176 @@
...
+                $retarr[] = $fld;
+            } else {
+                $retarr[strtoupper($fld->name)] = $fld;
+            }
+            $rs->MoveNext();
+        }
+
+        $rs->Close();
+        return $retarr;
+
```
jane-lyndon: additional \r\n that is not required

drivers/adodb-pdo_dblib.inc.php
```diff
@@ -0,0 +1,176 @@
...
+            return PDO::PARAM_INT ;
+        }
+        return PDO::PARAM_STR;
+    }
+
+    function ServerInfo()
+    {
+        return ADOConnection::ServerInfo();
+    }
+
```
jane-lyndon: \r\n that is not required

drivers/adodb-pdo_dblib.inc.php
Markomas: Removed, thanks

drivers/adodb-pdo_dblib.inc.php
Markomas: Removed, thanks

drivers/adodb-pdo_dblib.inc.php
Markomas: Removed, thanks

PR #532, opened 2019-04-09
Remove unnecessary srand() calls

quamis:
srand() should not be called manually, unless one has a better way to
initialise the already-initialised random seend

the way the current srand() call is made is buggy, and will lead to
duplicate random seeds if called often enough. Because the current call
will loose the timestamp part, and will only use the first 6 chars from
the current microsecond, but the system might not have a high-enough
timer resolution, so the actual pool of random seeds gets even smaller

### DIFF
--- a/adodb.inc.php
+++ b/adodb.inc.php
@@ -237,11 +237,6 @@ if (!defined('_ADODB_LAYER')) {
 			}
 		}
 
-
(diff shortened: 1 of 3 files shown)


dregad:
Thanks for your contribution.

PR #545, opened 2019-08-14
Auto reconnect to database if connection is lost for mysqli and pdo

dirtyren:
Hello,

   I've added auto reconnect support for the mysqli and PDO drivers. I was having a lot of problems having mysql has gone away errors and this solves this problem.
   I think it be simple to implement for every other database.

### DIFF
--- a/adodb.inc.php
+++ b/adodb.inc.php
@@ -693,7 +693,7 @@ if (!defined('_ADODB_LAYER')) {
 			$this->user = $argUsername;
 		}
 		if ($argPassword != "") {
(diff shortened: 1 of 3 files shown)


Mike-Benoit:
Sending a "SELECT 1" query to check if the connection is up for every other query that is executed is a non-starter unfortunately, as that would be horrible for performance.

Instead it needs to detect the disconnection by an error/exception and only retry at that point. PR #111 has something similar to that functionality if you want to take a look.

PR #560, opened 2019-11-11
Support for pdo_mysql

pharazon:
Problem: If you instantiate pdo://mydb/ the $this->metaTablesSQL is empty for ADODB_pdo. I'd like to instantiate ADODB_pdo_mysql, but there seems to be no other way than by this fix.

### DIFF
--- a/adodb.inc.php
+++ b/adodb.inc.php
@@ -4866,6 +4866,11 @@ http://www.stanford.edu/dept/itss/docs/oracle/10g/server.101/b10759/statements_1
 				$class = $db;
 				break;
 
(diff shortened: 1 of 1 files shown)


dregad:
I'm not sure what you are trying to do, and why you want to access `$this->metaTablesSQL`. For PDO drivers, the DB-specific meta information is stored in `$this->_driver`, and you can access the schema's metadata via `$this->metaTables()`.

Please refer to the [documentation](https://adodb.org/dokuwiki/doku.php?id=v5:database:pdo#establishing_a_connection) for the right way to connect with PDO drivers.

Issue #565, opened 2019-11-28
Auto-increment feature in Oracle driver problematical with long table names


mnewnham:
Oracle does not provide a native method of implementing an auto-increment column, so ADOdb provides an alternative method, via an insert trigger and a sequence.

In order to distinguish the name of the trigger and the sequence, the strings //**TRIG_**// and //**SEQ_**// are prepended to the name of the table to obtain the name of the trigger and sequence respectively.

This approach provides uniqueness, but also provides an easy method of obtaining the trigger and sequence associated  with a table.

The feature is robust and and well-understood, but a problem arises as the length of the table name approaches 30 characters. A table name "EMPLOYEES", would have a trigger created, named "TRIG_EMPLOYEES", falling well within the maximum characters available. However a table named "EMPLOYEE_HOURLY_PAY_RATES", already 26 characters long, would exceed the maximum name size when prepended with "TRIG_". When an attempt is made to create the trigger an error **ORA-00972: identifier is too long** occurs

dregad:
@mnewnham This is [not necessary with Oracle 12cR2 and later](https://docs.oracle.com/en/database/oracle/oracle-database/12.2/sqlrf/Database-Object-Names-and-Qualifiers.html#GUID-75337742-67FD-4EC0-985F-741C93D918DA), as that release allows **128 chars** for identifiers (as long as [COMPATIBLE parameter](https://docs.oracle.com/en/database/oracle/oracle-database/12.2/refrn/COMPATIBLE.html) is set to 12.2.0 or higher).

mnewnham:
I know, but my government clients are short on money and resources (I'm still supporting Microsoft SQL Server 2008 as well). I'd like to get it in there as it's pretty trivial and I'm trying to get my companies version of ADOdb back in line with master. It's drifted as we weren't able to get 5.21 out.

PR #566, opened 2019-11-28
Provide a method of creating compact trigger/sequence names, see #565

mnewnham:
This enhancement provides a method of creating an emulated auto-increment column for a table with a name greater than 25 characters, built using createTableSql() and specifying an "AUTO" option on the required column.

Instead of using the table name, a CRC32 value of the tablename is created and used for the name of the sequence and trigger, i.e. instead of a trigger called "TRIG_EMPLOYEES" for a table called "EMPLOYEES", it create a trigger "TRIG_3129131776", and similarly, a trigger for a table named "EMPLOYEE_HOURLY_PAY_RATES" becomes "TRIG_3754194288". The name of the trigger and sequence can be easily obtained after the fact by finding the CRC32 value of the table name.

The feature is provided with a compatibility flag that must be enabled. Existing configurations that already use this feature are unchanged. To activate, set the dictionary class variable $dict->useCompactAutoIncrements = true when building the table, and the connection class variable $db->useCompactAutoIncrements = true when using the table in ADOdb.

The feature is global for the database and cannot be specified on a table level basis.

### DIFF
--- a/datadict/datadict-oci8.inc.php
+++ b/datadict/datadict-oci8.inc.php
@@ -25,8 +25,15 @@ class ADODB2_oci8 extends ADODB_DataDict {
 	var $alterCol = ' MODIFY ';
 	var $typeX = 'VARCHAR(4000)';
 	var $typeXL = 'CLOB';
+	
+	/*
+	* Legacy compatibility for sequence names for emulated auto-increments
+	* If set to true, creates sequences and triggers as TRIG_394545594
+	* instead of TRIG_possibly_too_long_tablename
+	*/
+	public $useCompactAutoIncrements = false;
 
-	function MetaType($t, $len=-1, $fieldobj=false)
+	function metaType($t, $len=-1, $fieldobj=false)
 	{
 		if (is_object($t)) {
 			$fieldobj = $t;
@@ -185,32 +192,52 @@ class ADODB2_oci8 extends ADODB_DataDict {
 		return $suffix;
 	}
 
-/*
-CREATE or replace TRIGGER jaddress_insert
-before insert on jaddress
-for each row
-begin
-select seqaddress.nextval into :new.A_ID from dual;
-end;
-*/
+	/**
+	* Creates an insert trigger to emulate an auto-increment column
+	* in a table
+	*
+	* @param	str		$tabname		The name of the table
+	* @param	str[]	$tableoptions   Optional configuration items
+	*
+	* @return	str[]	The SQL statements to create the trigger
+	*/
 	function _Triggers($tabname,$tableoptions)
 	{
+		
 		if (!$this->seqField) return array();
 
-		if ($this->schema) {
+		if ($this->schema) 
+		{
 			$t = strpos($tabname,'.');
-			if ($t !== false) $tab = substr($tabname,$t+1);
-			else $tab = $tabname;
+			if ($t !== false) 
+				$tab = substr($tabname,$t+1);
+			else 
+				$tab = $tabname;
+			
+			if ($this->connection->useCompactAutoIncrements)
+				$id = sprintf('%u',crc32(strtolower($tab)));
+			else
+				$id = $tab;
+			
 			$seqname = $this->schema.'.'.$this->seqPrefix.$tab;
 			$trigname = $this->schema.'.'.$this->trigPrefix.$this->seqPrefix.$tab;
-		} else {
-			$seqname = $this->seqPrefix.$tabname;
-			$trigname = $this->trigPrefix.$seqname;
+			
+		} 
+		else 
+		{
+			if ($this->connection->useCompactAutoIncrements)
+				$id = sprintf('%u',crc32(strtolower($tabname)));
+			else
+				$id = $tabname;
+			
+			$seqname = $this->seqPrefix.$id;
+			$trigname = $this->trigPrefix.$id;
 		}
 
 		if (strlen($seqname) > 30) {
 			$seqname = $this->seqPrefix.uniqid('');
 		} // end if
+		
 		if (strlen($trigname) > 30) {
 			$trigname = $this->trigPrefix.uniqid('');
 		} // end if
@@ -222,7 +249,7 @@ end;
 		if (isset($tableoptions['SEQUENCE_INCREMENT'])){$seqIncr = ' INCREMENT BY '.$tableoptions['SEQUENCE_INCREMENT'];}
 		$seqStart = '';
 		if (isset($tableoptions['SEQUENCE_START'])){$seqStart = ' START WITH '.$tableoptions['SEQUENCE_START'];}
-		$sql[] = "CREATE SEQUENCE $seqname $seqStart $seqIncr $seqCache";
+		$sql[] = "CREATE SEQUENCE $seqname MINVALUE 1 $seqStart $seqIncr $seqCache";
 		$sql[] = "CREATE OR REPLACE TRIGGER $trigname BEFORE insert ON $tabname FOR EACH ROW WHEN (NEW.$this->seqField IS NULL OR NEW.$this->seqField = 0) BEGIN select $seqname.nextval into :new.$this->seqField from dual; END;";
 
 		$this->seqField = false;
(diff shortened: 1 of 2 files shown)


datadict/datadict-oci8.inc.php
```diff
@@ -185,32 +192,52 @@ function _CreateSuffix($fname,&$ftype,$fnotnull,$fdefault,$fautoinc,$fconstraint
...
 			$t = strpos($tabname,'.');
-			if ($t !== false) $tab = substr($tabname,$t+1);
-			else $tab = $tabname;
+			if ($t !== false) 
+				$tab = substr($tabname,$t+1);
+			else 
+				$tab = $tabname;
+			
+			if ($this->connection->useCompactAutoIncrements)
+				$id = sprintf('%u',crc32(strtolower($tab)));
```
dregad: Suggest to use `hash("crc32b", $str)` instead to avoid potential issues with negative numbers (see warning in [PHP documentation](https://www.php.net/manual/en/function.crc32.php)).

However unlikely that may be, what would happen if there is a collision on the hash ?

datadict/datadict-oci8.inc.php
```diff
@@ -185,32 +192,52 @@ function _CreateSuffix($fname,&$ftype,$fnotnull,$fdefault,$fautoinc,$fconstraint
...
 			$trigname = $this->schema.'.'.$this->trigPrefix.$this->seqPrefix.$tab;
-		} else {
-			$seqname = $this->seqPrefix.$tabname;
-			$trigname = $this->trigPrefix.$seqname;
+			
+		} 
+		else 
+		{
+			if ($this->connection->useCompactAutoIncrements)
+				$id = sprintf('%u',crc32(strtolower($tabname)));
```
dregad: Same comment as above

drivers/adodb-oci8.inc.php
```diff
@@ -304,6 +321,43 @@ function IfNull( $field, $ifNull )
...
+		if ($this->schema) 
+		{
+			$t = strpos($tabname,'.');
+			if ($t !== false) 
+				$tab = substr($tabname,$t+1);
+			else 
+				$tab = $tabname;
+			
+			if ($this->useCompactAutoIncrements)
+				$tab = sprintf('%u',crc32(strtolower($tab)));
```
dregad: And again

drivers/adodb-oci8.inc.php
```diff
@@ -304,6 +321,43 @@ function IfNull( $field, $ifNull )
...
+				
+			$seqname = $this->schema.'.'.$this->seqPrefix.$tab;
+		} 
+		else 
+		{
+			if ($this->useCompactAutoIncrements)
+				$tabname = sprintf('%u',crc32(strtolower($tabname)));
```
dregad: And again

datadict/datadict-oci8.inc.php
```diff
@@ -25,8 +25,15 @@ class ADODB2_oci8 extends ADODB_DataDict {
 	var $alterCol = ' MODIFY ';
 	var $typeX = 'VARCHAR(4000)';
 	var $typeXL = 'CLOB';
+	
+	/*
+	* Legacy compatibility for sequence names for emulated auto-increments
+	* If set to true, creates sequences and triggers as TRIG_394545594
+	* instead of TRIG_possibly_too_long_tablename
+	*/
```
dregad: - This should be a PHPDoc block (use `/**)
- add a space before the `*` on each subsequent line to align properly.
- Missing period `.`at end of 1st line (*auto-increments*)
- Mention that the number is computed as CRC32 from the table name
- Add `@var bool $useCompactAutoIncrements`

datadict/datadict-oci8.inc.php
```diff
@@ -185,32 +192,52 @@ function _CreateSuffix($fname,&$ftype,$fnotnull,$fdefault,$fautoinc,$fconstraint
...
-*/
+	/**
+	* Creates an insert trigger to emulate an auto-increment column
+	* in a table
+	*
+	* @param	str		$tabname		The name of the table
+	* @param	str[]	$tableoptions   Optional configuration items
+	*
+	* @return	str[]	The SQL statements to create the trigger
+	*/
```
dregad: - add a space before the * on each line after the opening `/**` to align properly.
- use `string` not str for type
- use spaces for vertical alignment within comments

dregad:
Also, if you could avoid very long lines in commit message, and wrap the text at 72 chars, it would be nice - I work a lot from the command-line and it is not good friends with long lines :

image

datadict/datadict-oci8.inc.php
mnewnham: You don't get -ve numbers on 64 bit installs, and as far as I can see, you only get collisions on binary strings.

If I was starting from scratch  perhaps your hash idea would be better, but to be honest, I'm trying to backfill stuff from my production version into master. As you pointed out, with 12, the limitation goes away, so I would suspect it's a feature with a short lifespan.

mnewnham:
@dregad , While I was using subversion as my windows client I was happy, accepting that I was using the non-standard client, to make the effort to accommodate your somewhat arcane needs on this. Indeed, SVN allowed me to deal with this rather easily inside a windows client.

I have now transitioned to the official Git Desktop Client for Windows, and this offers me nothing for commit messages other than this 230px x 110px text-wrap box that uses a non-proportional font.  
2020-01-11 16_00_14-Window 

So I think you're going to need to make some adjustments on your end.

Also, if there are inconsistencies in commenting, I think our crucially short available time would be better allocated if you fixed them to your liking, then let me know what has changed so that I might adjust later commits when I remember, rather than kicking them back. I'm not looking to get my code marked for style points, but clearing issues as fast as possible with my somewhat limited time, and I'd rather focus on that. I know that this is much more important to you than me but you must understand my frustration that we have been bouncing this commit back and forwards for a week for reasons that you could have resolved yourself in 15 minutes

dregad:
Re: the commit message wrapping, sorry to be a pain... I don't mind doing the legwork if that's what it takes to accomodate my so-called _arcane needs_. In that case, please don't commit to _master_ directly, and let me know what's ready to merge via pull requests instead.

Just for the record, Github Desktop is hardly _the_ standard client, it's just one of many GUI's out there to abstract the command-line, which is where Git's roots are. They have a pending enhancement request to introduce wrapping, see https://github.com/desktop/desktop/issues/62.

As for fixing things directly, I'm used when reviewing PR's in other projects I'm working on, to just provide comments and let the submitter take care of fixing things as needed. If you prefer that I update the code directly, I can do that, no problem.

dregad:
Rebased, adjusted commit message (shortened 1st line and wrapped the rest to 72 chars)

Fixed PHPDoc in separate commit.

datadict/datadict-oci8.inc.php
dregad: OK fine

mnewnham:
> _Arcane Needs_

I thought you might like that one

> Github Client

Unfortunately, I have but a passing knowledge of  Git, its command structure and clients. As you know, I'm tied to subversion and that is unlikely to end. But if an upgrade to the client makes it easy to facilitate a more structured message, I'll certainly try to incorporate it. In the meantime, I'm happy to leave 99% of fixes ready for pull requests. There'll always be the odd one I'm going to commit directly.

dregad:
> I thought you might like that one

:grin: 

> if an upgrade to the client makes it easy to facilitate a more structured message, I'll certainly try to incorporate it.

If you don't mind trying another client, you may want to have a look at [SmartGit](https://www.syntevo.com/smartgit/). It's commercial software, but free for open-source. Of all the many GIT GUIs that I've tested it's most definitely the best one.

When commiting, the dialog box shows guidelines for line width (at 50 chars for the first line and 72 for the commit's body), and it highlights the line in red if it is too long, making it easy to add a hard newline, but still lets you commit as-is, e.g.

image
image

> I'm happy to leave 99% of fixes ready for pull requests. There'll always be the odd one I'm going to commit directly.

Thanks for your undestanding, accomodating my Arcane Style :wink:  and your cooperation.

PR #575, opened 2019-12-31
text driver review

peterdd:
- documented way (workaround) to make it working without changing ADOdb 5 core files
- fixed the example eval strings
- fixed some of the code style while going through

see also changes at https://adodb.org/dokuwiki/doku.php?id=v5:database:text

### DIFF
--- a/drivers/adodb-text.inc.php
+++ b/drivers/adodb-text.inc.php
@@ -9,8 +9,8 @@
 /*
 Setup:
 
- 	$db = NewADOConnection('text');
- 	$db->Connect($array,[$types],[$colnames]);
+ 	$db = ADONewConnection('text');
+ 	$db->connect($array, [$types], [$colnames]);
 
 	Parameter $array is the 2 dimensional array of data. The first row can contain the
 	column names. If column names is not defined in first row, you MUST define $colnames,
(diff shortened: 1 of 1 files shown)


drivers/adodb-text.inc.php
```diff
@@ -176,32 +179,37 @@ function Connect(&$array, $types = false, $colnames = false)
...
-	function _query($sql,$input_arr,$eval=false)
+	function _query($sql, $input_arr, $eval=false)
 	{
 		if ($this->_origarray === false) return false;
 
-		$eval = $this->evalAll;
+		if (!is_string($eval)) {
+			// This looks like a workaround for ADOdb 5 not having a 3rd execute() parameter by setting
+			// that evalAll object variable before calling execute()
+			$eval = $this->evalAll;
```
dregad: The _evalAll_ property is actually not defined anywhere.

The _ADOConnection_ class does define *$_evalAll*, but it is not used anywhere so maybe it should be removed from that class and defined in *ADODB_text* instead ? 

In any case, the property's name should be consistent.

peterdd: Yeah, I found this $_evalAll defined later too in the ADOConnection class and wanted ask that question about too. I'm fine with the removal of it from the ADOConnection class and keep that hacky stuff in the ADODB_text class.

peterdd: > and automatically clear it after query execution ?

Yes, I thought also about it, but was unsure what the original author had in mind with setting and use of the 'evalAll' variable.

dregad:
> > and automatically clear it after query execution ?
> 
> Yes, I thought also about it, but was unsure what the original author had in mind with setting and use of the 'evalAll' variable.

Considering that the original author is gone, that there is an obvious inconsistency in variable naming and that this is the only place that this property is being used, IMO the risk of such change is really low.

dregad:
@peterdd what is the status of this PR ?

peterdd:
Nothing new on this front. Generally I still wish to keep something text/proxy/csv driverlike in ADOdb, but more important is progress on PHP8 and xmlschema stuff. And bring the 5.20, 5.21, master branches back in sync if possible, so I do not mess up too easily when doing PR. I usually do/prefer doing PR against master as it is easier for me to stick to one branch. I thought master will be the 5.21, but it seems it is now 5.20,5.21,5.22 branch?

PR #591, opened 2020-01-08
Create a way to set the mysqli connection timeout

runia1:
### Description

We needed a way to set the mysqli connection timeout very low (1 second).  This is the way we did it.  It may be useful to implement this setting for the other drivers besides mysqli.


mnewnham:
This is already available using [setConnectionParameter](https://adodb.org/dokuwiki/doku.php?id=v5:reference:connection:setconnectionparameter)

````php
$db = newAdoConnection('mysqli');
$db->setConnectionParameter(MYSQLI_OPT_CONNECT_TIMEOUT,1);
$db->connect($host,$user,$pass,$db);
````

Thanks though, feel free to re-open if this is not what you were trying to achieve

runia1:
Actually that is awesome!  Thank you

PR #592, opened 2020-01-08
fix descr tag in session_schema2.xml

peterdd:


peterdd:
see also #595 and #594

dregad:
Fixed in 88106d48a4e6c0d212ba9f82f57a0ef91efb12dd.

PR #593, opened 2020-01-08
fix descr tag in session_schema.xml

peterdd:


dregad:
For the changelog, can you please open an issue briefly describing the reason for this change, and referencing this and #592.

I suppose this is related to #265 and the discussion in https://github.com/dregad/ADOdb/commit/4deada337f19e4e1ba0e3af7310f9d9a4993f4ab#commitcomment-36716892 ?

peterdd:
Yes, but I do care more about the xmlschema03 that is really used.
This file is about version 0.2 (xmlschema02) which dtd not even exists, but I fixed the xml file for completion.

Also see #595 and #594

dregad:
Fixed in 88106d48a4e6c0d212ba9f82f57a0ef91efb12dd.

PR #594, opened 2020-01-10
valid xmlschema03.dtd

peterdd:
Trying to make the xmlschema03.dtd a valid one following https://www.w3.org/TR/xml11/ , so it can be used for validation of xmlschema03 xml files.


dregad:
Fixed in 88106d48a4e6c0d212ba9f82f57a0ef91efb12dd

PR #605, opened 2020-01-26
Array to string conversion on exception with query parameters

### LINKED ISSUE #601: Array to string conversion on exception with query parameters
The ADODB_Exception constructor does not handle queries with parameter arrays. Patch attached.

[adodb-fix-exception-param.patch.txt](https://github.com/ADOdb/ADOdb/files/4076397/adodb-fix-exception-param.patch.txt)

hschletz:
Fixes #601

### DIFF
--- a/adodb-exceptions.inc.php
+++ b/adodb-exceptions.inc.php
@@ -29,6 +29,9 @@ var $database = '';
 
 	function __construct($dbms, $fn, $errno, $errmsg, $p1, $p2, $thisConnection)
 	{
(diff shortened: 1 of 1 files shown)


dregad:
Following discussion in https://github.com/ADOdb/ADOdb/issues/601#issuecomment-579771136, considering that the proposed patch is only addressing the symptoms of the problem instead of its root cause, this PR will be closed without merge.

dregad:
The actual fix for this was committed in 0e813bcc8da0985559ec9e9ce7b3ba882f381d9a.

Issue #604, opened 2020-01-26
MySQL Affected_Rows() returns number of rows selected instead of false for SELECT statements


mariobernheim:
Hi, 
```
<?php
$sql = "SELECT * from subscribers WHERE number='".$number."' AND service_id='".$service_id."' LIMIT 1";
$rs  = $Conn->execute($sql);
if ($Conn->Affected_Rows()) {
    $active = $rs->fields['active'];    
    
    echo return_json_status("3","Number $number is subscribed to $service_id and is set active=$active");
    die("");
}
?>
```

For some reason this works... however, from the ADOdb Documentation, its not supposed to:

`https://adodb.org/dokuwiki/doku.php?id=v5:reference:connection:affected_rows`

Database: MySQL 5.7
Driver: Mysqli
PHP Version: PHP 7.0.33
ADOdb Version: V5.18

mnewnham:
Database? Driver? PHP Version? ADOdb Version?

mariobernheim:
Database: MySQL 5.7
Driver: Mysqli
PHP Version: PHP 7.2
ADOdb Version:  v5.20.14

mariobernheim:
Database: MySQL 5.7
Driver: Mysqli
PHP Version: PHP 7.0.33
ADOdb Version: V5.18

mariobernheim:
For references... I just tested this these

PHP Version: 7.2.25
ADOdb Version: v5.20.14

I do SELECT * from table limit 1 and Affected_Rows(); returns 1

mnewnham:
This is due to a change in the behavior of the underlying mysqli_affected_rows(), which according to [The PHP documentation](https://www.php.net/manual/en/mysqli.affected-rows.php), now acts as mysqli_num_rows() for SELECT statements. In order to maintain cross-database compatibility, we need to suppress this behavior.

mariobernheim:
Thanks, so for all purposes, we should be able to use Affected_Rows(); on all SELECT's as well and can consider it reliable?

mnewnham:
No, you should not, because no other database supports it. The code will be changed so that it conforms to our standard, i.e. It will return false for SELECT statements

mariobernheim:
Should we use RecordCount(); instead?

mnewnham:
absolutely

peterdd:
[...]

PR #607, opened 2020-01-27
Retrieve SQL Server Affected Rows early before connection closes

### LINKED ISSUE #590: mssqlnative query are not correctly closed
In  drivers/adodb-mssqlnative.inc.php
method _close() contains a bug causing this method do nothing
`if(is_object($this->_queryID)) {`
is wrong because $this->_queryID is not an object but a resource (probable copy/paste from PDO version) 
it should be 
`if(is_resource($this->_queryID)) {`
It cause extra cursor left open, and if there's multiple big query on the same DB blocs through a getrow it can cause a deadlock, it's not common on normal operation but I've spend hours on a such bug due to a bugged query. Could be an issue on long worker too.

### LINKED ISSUE #606: Fix #590 causes SQL Server Affected_rows() to fail
Affected_rows() for SQL Server requires a query resource id available after the completion of a query, but fix #590 causes the query to be closed at an early stage, before ``execute()`` has ended. This means that the query resource is no longer available.

mnewnham:
Patch for #590 closes the connection early, making affected_rows() fail. In addition, the value returned for an invalid query (-1) does not conform to the ADOdb standard (false). Affected Rows is now retrieved immediately after query execution and stored for later use.

Fixes #606 

@dregad This needs to go into the new release, if you've already built it, I'm afraid we need to create a hotfix

### DIFF
--- a/drivers/adodb-mssqlnative.inc.php
+++ b/drivers/adodb-mssqlnative.inc.php
@@ -93,6 +93,11 @@ class ADODB_mssqlnative extends ADOConnection {
 
 	var $sequences = false;
 	var $mssql_version = '';
+	
+	/*
+	* Stores the affected rows after a query
+	*/
+	private $sqlServerAffectedRows = -1;
 
 	function __construct()
 	{
@@ -162,8 +167,13 @@ class ADODB_mssqlnative extends ADOConnection {
 
 	function _affectedrows()
 	{
-		if ($this->_queryID)
-		return sqlsrv_rows_affected($this->_queryID);
+		/*
+		* If a non-update statement, sql server returns
+		* -1, ADODb needs false
+		*/
+		if ($this->sqlServerAffectedRows == -1)
+			return false;
+		return $this->sqlServerAffectedRows;
 	}
 
 	function GenID($seq='adodbseq',$start=1) {
@@ -595,6 +605,12 @@ class ADODB_mssqlnative extends ADOConnection {
 
 		if ($this->debug) ADOConnection::outp("<hr>running query: ".var_export($sql,true)."<hr>input array: ".var_export($inputarr,true)."<hr>result: ".var_export($rez,true));
 
+		/*
+		* Store off the affected rows for retrieval later
+		*/
+		$this->sqlServerAffectedRows = sqlsrv_rows_affected($rez);
+		
+		
 		if(!$rez)
 			$rez = false;
 
--- a/drivers/adodb-mssqlnative.inc.php
+++ b/drivers/adodb-mssqlnative.inc.php
@@ -608,7 +608,8 @@ class ADODB_mssqlnative extends ADOConnection {
 		/*
 		* Store off the affected rows for retrieval later
 		*/
-		$this->sqlServerAffectedRows = sqlsrv_rows_affected($rez);
+		if (is_resource($rez))
+			$this->sqlServerAffectedRows = sqlsrv_rows_affected($rez);
 		
 		
 		if(!$rez)
--- a/drivers/adodb-mssqlnative.inc.php
+++ b/drivers/adodb-mssqlnative.inc.php
@@ -606,11 +606,21 @@ class ADODB_mssqlnative extends ADOConnection {
 		if ($this->debug) ADOConnection::outp("<hr>running query: ".var_export($sql,true)."<hr>input array: ".var_export($inputarr,true)."<hr>result: ".var_export($rez,true));
 
 		/*
-		* Store off the affected rows for retrieval later
+		* Store off the affected rows for retrieval later. There is no
+		* good way of getting a rowcount for a select statement
 		*/
-		if (is_resource($rez))
+		if (is_resource($rez) && !preg_match('/(SELECT COUNT)/',$sql))
+		{
+			//$recordCount = _adodb_getcount($this, $sql,$inputarr);
+			//print "RC=$recordCount"; exit;
+		}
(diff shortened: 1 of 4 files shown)


dregad:
> This needs to go into the new release, if you've already built it, I'm afraid we need to create a hotfix

I was planning to cut the release yesterday, but got sidetracked and did not have time to finalize it, so it can still go in as-is. But in anycase, a hotfix would be no problem either.

What is the relationship between this and #606 ? Seems to be the same or similar issue.

mnewnham:
It is, its just the pull request for you to work with, no?

dregad:
I thought so, but as you had not referenced issue #606 in this PR I just wanted to make sure.

dregad:
Considering that #590 was released in 5.20.16, I suppose that the regression is present there as well, so should I backport the fix and cut a 5.20.17 hotfix ?

drivers/adodb-mssqlnative.inc.php
```diff
@@ -595,6 +605,12 @@ function _query($sql,$inputarr=false)
 
 		if ($this->debug) ADOConnection::outp("<hr>running query: ".var_export($sql,true)."<hr>input array: ".var_export($inputarr,true)."<hr>result: ".var_export($rez,true));
 
+		/*
+		* Store off the affected rows for retrieval later
+		*/
+		$this->sqlServerAffectedRows = sqlsrv_rows_affected($rez);
```
dregad: According to [documentation](https://docs.microsoft.com/en-us/sql/connect/php/sqlsrv-query?view=sql-server-ver15#return-value), sqlsrv_query() returns _false_ if the statement cannot be created and/or executed.

I didn't test this, but I suppose that in such scenario, `sqlsrv_rows_affected($rez)` would throw some kind of error / warning.

Maybe better to do:
```php
if (!$rez) {
    $this->sqlServerAffectedRows = -1;
    return false;
} else {
    $this->sqlServerAffectedRows = sqlsrv_rows_affected($rez);
}
```

obmsch:
@dregad +1 to backport and for the additional check.

And for the sake of clarity the comment:
```
  * If a non-update statement, sql server returns
  * -1, ADODb needs false  
```
should be:
```
  * If sql server affected rows is -1 return false (ADOdb required)
```
My point is: There are other statements than update (delete, insert, ...) where
the affected rows are >= 0. Effectively (sql server) -1 matches false (ADOdb).

mnewnham:
>should be:
> If sql server affected rows is -1 return false (ADOdb required)

I disagree, the original statement not only explains why the value returned is -1, It explains what value ADOdb requires in a concise manner that is both grammatically and syntactically correct. Your revision eliminates the explanation that it is a non-update that is returning -1, as well as introducing a needlessly complex parenthetical statement.

In other news, I've added a test against the query result which improves the robustness of the check in case of a failed query.

dregad:
>  I've added a test against the query result which improves the robustness of the check in case of a failed query.

Thanks, although I'm not sure why an additional if statement is needed, since the conditions are mutually exclusive (we either get a resource, or false). But that's a nitpick and I can live with that.

I'm still concerned that by not resetting _sqlServerAffectedRows_ property after a failed query execution. there is a risk that an incorrect value could be returned by *Affected_Rows()*, for example in the following scenario: query 1 successfully updates 3 records, then update query 2 fails; in that case sqlServerAffectedRows == 3 but it should be -1 / false. Sorry I can't actually test, I don't have a MSSQL setup

As for the -1 vs false discussion, I think @obmsch's point was that your comment explicitly refers to *update* statements, implying that *delete/update* also return -1 which is incorrect. Anyway IMO the whole point is moot and the code block could even be removed because this is already taken care of by [Affected_Rows() function, line 1483](https://github.com/ADOdb/ADOdb/blob/517c8adc141d9375079729d0cf04deb8c6840113/adodb.inc.php#L1483) :

```php
$val = $this->_affectedrows();
return ($val < 0) ? false : $val;
```

mnewnham:
> update statements, implying that delete/update also return -1 which is incorrect. Anyway IMO the whole point is moot and the code block could even be removed
Which code block?

dregad:
> Which code block?

Apologies, I should have been explicit. I meant this
https://github.com/ADOdb/ADOdb/pull/607/files#diff-5f880e0824d7dd21bee7e18f6e54c4f8R170-R175

mnewnham:
Yes, I can see what you are getting at, which raises a couple of questions for me,

I think that the way Affected_Rows() handles the mssql data and actually returns the right value is more luck than skill. It looks like it was more designed to trap general errors. 

My first question on this is should we create a rule that the driver handle the query response correctly in` driver::_affectedRows()` and return the correct values (i.e. false or a number >= 0) to `adoconnection:affected_rows()` and not rely on the latter massaging any incorrect return values. Looking at #604, It's clear that we are going to have to make the mysql driver do the work to return the correct values.  This does also have the benefit of keeping driver specific code out of the core functions.

Secondly, looking at the top of the Affected_Rows() function, it looks as if the function will not return the correct value for mssql if performance monitoring is enabled.  I haven't tested it, but it looks as though it will return -1 rather than false

dregad:
@mnewnham since you closed this, can the [SQL-Server-Affected-Rows](https://github.com/ADOdb/ADOdb/tree/SQL-Server-Affected-Rows) be deleted ?

mnewnham:
I've deleted the branch, but kept a local copy. I've no idea where this stands but it is so far out of date it will need to be re-coded anyway

PR #632, opened 2020-09-16
Update pdo_sqlsrv driver based on dblib driver

Markomas:
Because pdo_sqlsrv was lacking some functionality (MetaColumns, MetaTables...), I updated it based on pdo_dblib driver

### DIFF
--- a/drivers/adodb-pdo_sqlsrv.inc.php
+++ b/drivers/adodb-pdo_sqlsrv.inc.php
@@ -5,12 +5,16 @@
  */
 class ADODB_pdo_sqlsrv extends ADODB_pdo
 {
+
 	var $hasTop = 'top';
 	var $sysDate = 'convert(datetime,convert(char,GetDate(),102),102)';
-	var $sysTimeStamp = 'GetDate()';
-	var $arrayClass = 'ADORecordSet_array_pdo_sqlsrv';
+	var $sysTimeStamp = 'GetDate()';var $metaDatabasesSQL = "select name from sysdatabases where name <> 'master'";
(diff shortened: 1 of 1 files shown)


dregad:
@mnewnham Please review and let me know if this can be merged and the target version (5.21 or 5.22 ?)

mnewnham:
The code has some changes to **_query()** that I need to run some physical tests on to understand. Could you please hold off merging for the moment. In any case it should be pulled intoor 5.22

mnewnham:
@Markomas , the _query() method you have implemented needs a little more work to handle bind parameters correclty. When portable SQL is written using the **param()** method, the inbound keys are strings, and SQL Server needs them to be integers. For example:

````php

$p1 = $db->param('p1');
$p2 = $db->param('p2');

$bind = array('p1'=>5, 'p2'=>'CHEESE');

$SQL = "SELECT * FROM foods WHERE food_group=$p1 AND food_name=$p2";

$db->execute($SQL,$bind);

````
Because the inbound parameters are associative, the query fails:

````
(pdo-sqlsrv): SELECT * FROM foods WHERE food_group=? AND food_name=? 
   [ (p1=>'5', p2=>'CHEESE')  ]
-----<hr>
-29: Tried to bind parameter number 65536.  SQL Server supports a maximum of 2100 parameters.
````
If you look at the **mssqlnative** driver, you can see that we handle this issue by simply converting the string using **array_values()**.   

Keep up the good work

dregad:
@Markomas are you planning to address Mark's review comments ?

dregad:
@Markomas you have not provided any feedback or updated the PR since

dregad:
@mnewnham unless you're OK with applying any required fixes yourself, I propose to close this without merge. Let me know.

mnewnham:
We talked about this driver elsewhere. I believe that we agreed, based on the fact that there are now official SQL Server for Linux drivers, to deprecate this immediately and remove it in future releases

mnewnham:
We talked about this driver elsewhere. I believe that we agreed, based on the fact that there are now official SQL Server for Linux drivers, to deprecate this immediately and remove it in future releases

dregad:
> We talked about this driver elsewhere. I believe that we agreed, based on the fact that there are now official SQL Server for Linux drivers, to deprecate this immediately and remove it in future releases

I vaguely remember that, but wasn't that discussion about the _mssql_(non-PDO) driver ? I could not find the original discussion. 

In any case, this PR is about the PDO driver. PHP documentation states:

- [PDO_DBLIB](https://www.php.net/manual/en/ref.pdo-dblib.php):
    > This extension is not available anymore on Windows. 
    > On Windows, you should use SqlSrv, an alternative driver for MS SQL is available from Microsoft: » http://msdn.microsoft.com/en-us/sqlserver/ff657782.aspx .
- [PDO_SQLSRV](https://www.php.net/manual/en/ref.pdo-sqlsrv.php)
    > The PDO_SQLSRV extension is only compatible with PHP running on Windows. For Linux, see ODBC and » Microsoft's SQL Server ODBC Driver for Linux. 

So I don't think PDO_SQLSRV driver should be deprecated at this time, since it's currently the recommended one to use on Windows with PDO.

mnewnham:
No, I'm sorry, I got this one backwards to the one I was thinking about. I didn't really understand where this driver came from. I don't think it was in 5.19, and then it appeared with a flurry of updates. 

You are absolutely correct in that pdo_sqlsrv should not be deprecated, but also in addition, we should not pull in any fixes to it based on pdo_dblib. The only source of upgrades to pdo_sqlsrv should be mssqlnative.

dregad:
OK so I'm closing this as _won't fix_ then.

PR #633, opened 2020-09-17
Update datadict-postgres.inc.php

jmruiz84:
ChangeTableSQL function override. Inheritance from ADODB DataDict does not work This new function has been verified in postgresql 12.2
#634

### DIFF
--- a/datadict/datadict-postgres.inc.php
+++ b/datadict/datadict-postgres.inc.php
@@ -539,8 +539,9 @@ CREATE [ UNIQUE ] INDEX index_name ON table
 		if ( empty($cols)) {
 			$sqlResult=$this->CreateTableSQL($tablename, $flds, $tableoptions);
 		} else {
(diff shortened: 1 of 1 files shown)


datadict/datadict-postgres.inc.php
```diff
@@ -516,4 +516,47 @@ function _GetSize($ftype, $ty, $fsize, $fprec, $options=false)
...
+				// genfields can return FALSE at times
+				if ($lines == null) 
+					$lines = array();
+				$alter = 'ALTER TABLE ' . $this->TableName($tablename);
+				foreach ( $cols as $id => $v ){
+					if ( !isset($lines[$id]) ){
+						$sqlResult[] = $alter . $this->dropCol . ' ' . $v->name;
+					}
+				}
+			}
```
jmruiz84, line 557: dropcol function defined in this class does not work, I manage the removal of the column manually

PR #639, opened 2020-10-23
do not use name quotes in pdo-dblib

pauljaks:
When using with MSSQL database adding quotes around table names fails (for example UPDATE "dbo.testTable" ).


dregad:
@mnewnham I let you review and decide what to do with this; set Milestone as needed.

mnewnham:
I feel this overlaps with something else, or a requested global feature change. I'm not marking duplicate yet until I've reviewed the code but please don't pull

dregad:
@mnewnham did you get a chance to review ?

mnewnham:
I'll look at it tomorrow

mnewnham:
I've reviewed this and here are my findings and some background:

1. pdo_dblib is the Linux only freetds connector to SQL Server databases. Support for Windows was removed from PHP in 5.3.
2. Since that removal, MS has introduced the Native Mode and PDO drivers for Linux, effectively obsoleting this driver.
3. If we want to retain this driver, I think we should:
    -   Rename the pdo_mssql driver to pdo_dblib and merge any new dblib changes it
    -  Deprecate and remove the mssql driver. 
4. Finally, [This discussion](https://github.com/FreeTDS/freetds/issues/321) at FreeTDS addresses the issue raised here. I see no reason to make any changes in our driver.

dregad:
Thanks for your research. So if I'm reading you right, this PR should be closed as "won't fix".

Considering that we have cross-platform Microsoft-supported drivers, I don't see much point in maintaining a linux-only one so we might as well deprecate/remove it and save ourselves the maintenance effort. That's a separate work item though, and should be tracked in its own issue/PR.

mnewnham:
Agreed

dregad:
Closing PR per reasons given above.

PR #655, opened 2020-12-21
Add support for bound variable statements in MySQL

mnewnham:
Adds support for true bound variables in prepared statements in the mysqli driver.. Bound variable statement  now use mysqli_statement_execute instead of an emulated form.

### DIFF
--- a/drivers/adodb-mysqli.inc.php
+++ b/drivers/adodb-mysqli.inc.php
@@ -76,6 +76,8 @@ class ADODB_mysqli extends ADOConnection {
 	 */
 	private $usePreparedStatement = false;
 	private $useLastInsertStatement = false;
+	private $usingBoundVariables = false;
+	private $statementAffectedRows = -1;
 
 	/**
 	 * Sets the isolation level of a transaction.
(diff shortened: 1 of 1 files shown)


drivers/adodb-mysqli.inc.php
```diff
@@ -1033,13 +1038,176 @@ function Prepare($sql)
...
+			* Place the field type list at the front of the
+			* parameter array. This is the mysql specific
+			* format
+			*/
+			$typeArray[0] = $typeString;
+		
+			$ret = $this->_execute($sql,$typeArray);
+			if (!$ret) {
+				return $ret;
+			}
```
dregad, line 1195: @mnewnham I came upon this as I was reviewing your patch for #806.

As the code stands, this if statement is useless and could be removed, because $ret is already being returned at the end of the function, just a few lines later. Maybe you meant to `return false;` ?

dregad: Mark's reply in https://github.com/ADOdb/ADOdb/pull/816#issuecomment-1076967876
> I did review that uneeded bit of code, you are correct in that it could be removed, so if its easiest you could just chop it out

dregad: Fixed in 88aa5d0f75d9888b6b6ce60f75cfdeb35427b136

PR #658, opened 2020-12-28
PHP8 xmlschema03 for 5.21

peterdd:
- fix for the PHP8 changes stated on https://www.php.net/manual/en/function.xml-parser-create

> 8.0.0 This function returns an XMLParser instance now;       previously, a resource was returned,  or false on failure.

- removes magic_* stuff from xmlschema03 php sources.


dregad:
@peterdd based on PR title, my understanding is that this is intended for _5.21_, despite being targeted at _master_; I'll cherry-pick.

peterdd:
Yes, the magic_* fixes are required for 5.20 and 5.21 branches too for PHP7.4 and PHP8 compatibility.

Issue #656, opened 2020-12-27
sqlite3 metaIndexes() reversed $primary param results, php notice, bug


peterdd:
There is something wrong with the metaIndexes() function for sqlite3:

```php
<?php
include 'ADOdb/adodb.inc.php';

/**
 * sessions2 table in mysql/mariadb10.4.14 as created by ADOdb/session/adodb-sessions2.php adodb_session_create_table()

MariaDB [adodbtest]> show create table sessions2;
CREATE TABLE `sessions2` (
  `SESSKEY` varchar(64) NOT NULL,
  `EXPIRY` datetime NOT NULL,
  `CREATED` datetime NOT NULL,
  `MODIFIED` datetime NOT NULL,
  `EXPIREREF` varchar(250) DEFAULT NULL,
  `SESSDATA` longtext NOT NULL,
  PRIMARY KEY (`SESSKEY`)
) ENGINE=InnoDB DEFAULT CHARSET=utf8mb4

*/

/**
 * sessions2 table in sqlite3 as created by ADOdb/session/adodb-sessions2.php adodb_session_create_table()
 *
sqlite> select * from sqlite_master;
table|sessions2|sessions2|2|CREATE TABLE sessions2 (
SESSKEY                  VARCHAR(64) NOT NULL,
EXPIRY                   DATETIME NOT NULL,
CREATED                  DATETIME NOT NULL,
MODIFIED                 DATETIME NOT NULL,
EXPIREREF                VARCHAR(250),
SESSDATA                 LONGTEXT NOT NULL,
                 PRIMARY KEY (SESSKEY)
)
index|sqlite_autoindex_sessions2_1|sessions2|3|

*/

$conf=array(
  array(
    'dbtype'=>'mysqli',
    'host'=>'localhost',
    'database'=>'adodbtest',
    'user'=>'adodb',
    'password'=>'adodb'
        ),
  array(
    'dbtype'=>'sqlite3',
    'host'=>'adodbtest_sessions2.sqlite',
    'database'=>null,
    'user'=>null,
    'password'=>null
  )
);
echo '<pre>';
foreach($conf as $c) {
  echo "\n\nTest ".$c['dbtype']."\n";
  testMetaIndexes($c);
}

function testMetaIndexes($c)
{
  echo "\n".$c['dbtype']." table sessions2\n";
  $db= NewADOconnection($c['dbtype']);
  $db->connect($c['host'], $c['user'], $c['password'], $c['database']);
[...]

PR #659, opened 2020-12-28
MetaIndexes does not retrieve primary key information correctly, see #656 

mnewnham:
1. The primary key argument to metaIndexes() was reversed in the sqlite3 driver
2. Information about the auto-created primary key was not correctly retrieved.


dregad:
This has been fixed in 45a1d6a567727a846fe2fa218c3f77dbc77464fc, see #656.

PR #660, opened 2020-12-28
metaIndexes does not return primary key correctly, see #656

mnewnham:
The primary key parameter was  reversed . In addition, the method tried to obtain primary key information  from the sqlite_master table, instead of the pragma


dregad:
Since this #656 was flagged for 5.21, I'm backporting this.

PR #668, opened 2021-01-04
fix Cannot pass parameter 1 by reference for text driver

peterdd:


dregad:
Fixed in c3eb0b37cdb95a1be2e5932ada6be778ecf4f229

PR #670, opened 2021-01-08
Remove duplicated code from method qstr

stodorovic:
Methods addq and qstr use same code. Call addq from qstr.

### DIFF
--- a/adodb.inc.php
+++ b/adodb.inc.php
@@ -178,7 +178,6 @@ if (!defined('_ADODB_LAYER')) {
 	define('DB_AUTOQUERY_UPDATE', 2);
 
 
-	
 	function ADODB_Setup() {
 	GLOBAL
 		$ADODB_vers,		// database version
@@ -1456,8 +1455,8 @@ if (!defined('_ADODB_LAYER')) {
 	 *
 	 * @param string $table
 	 * @param string $id
-	 
-	 * @return mixed The last inserted ID. All databases support this, but be 
+	 *
+	 * @return mixed The last inserted ID. All databases support this, but be
 	 *               aware of possible problems in multiuser environments.
 	 *               Heavily test this before deploying.
 	 */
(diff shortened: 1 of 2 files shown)


mnewnham:
From [the documentation](https://adodb.org/dokuwiki/doku.php?id=v5:reference:connection:addq)

**The function addQ() takes an input string, and applies database specific string quoting. Unlike the method qStr, no pre or post quoting is applied.**

e.g
````php
$string = "Seamus O'Reilly";

print $db->addQ($string);

Seamus O\'Reilly

print $db->qStr($string);

'Seamus O\'Reilly'

````

So if they are in fact the same, that is incorrect. The 2 functions should return different values as shown above.

stodorovic:
They are different functions, but there is "same code" in both methods. It's following code:

https://github.com/ADOdb/ADOdb/blob/97066862ee9d7ae4ada1a0a530330f88cff26e8c/adodb.inc.php#L2943-L2967

Only differences are:

`return  "'".str_replace("'",$this->replaceQuote,$s)."'";`
vs
`return  str_replace("'",$this->replaceQuote,$s);`

`return "'$s'";`
vs
`return $s; `

`return "'".str_replace("\\'",$this->replaceQuote,$s)."'";`
vs
`return str_replace("\\'",$this->replaceQuote,$s); `

So, _qstr_ returns:
`return "'".$this->addq($s,$magic_quotes)."'";`

This PR simplifies current code in function qstr.

mnewnham:
Ok, I see what you are doing there. This may be problematical as you are removing the core system code and replacing it with a call to a driver specific method. So while what you have done may be correct for the mssql driver, we don't know that this is true for every other one. Can I suggest a safer approach would be to extend the qstr method into the driver and place your changes there.

dregad:
@stodorovic are you planning to address the review comments ?

dregad:
@stodorovic as you have not provided any feedback or updated the PR, I suppose you are not interested in pursuing this. I am therefore closing this pull request for now, feel free to reopen it if you intend to follow Mark's suggestion.

PR #675, opened 2021-01-25
Remove magic quotes

### LINKED ISSUE #674: Remove all magic quotes related code
Let's finally get rid of this.

For backwards compatibility purposes we'll just keep the `$magic_quote` parameter for public methods (e.g. qStr(), addQ(), AutoExecute(), GetUpdateSQL(), GetInsertSQL() ), but it no longer used and is defaulted to _false_, to avoid changing the functions' signature.

Following up on / related to #619, #643, #644, #657, #658.

dregad:
Fixes #674

### DIFF
--- a/adodb-csvlib.inc.php
+++ b/adodb-csvlib.inc.php
@@ -245,16 +245,8 @@ $ADODB_INCLUDED_CSV = 1;
 
 		fclose($fp);
 		@$arr = unserialize($text);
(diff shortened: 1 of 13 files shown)


dregad:
@mnewnham @peterdd your review and feedback would be appreciated before I merge this.

dregad:
Without feedback, I assume everyone is OK with this.

PR #676, opened 2021-01-25
Allow use of weighted server groups in the Memcached connection library

mnewnham:
The memcached caching service now supports the use of **Weighted Server Groups**,
This allows improved control over the **memcached** service when used with multiple memcache servers. 
This feature is not available for Windows clients. For more information, see https://adodb.org/dokuwiki/doku.php?id=v5:userguide:memcached

The memcached caching service now supports the use of **Sever Options**,
This allows the **memcached** service to be configured with available options, such as Hashing method. 
This feature is not available for Windows clients. For more information, see https://adodb.org/dokuwiki/doku.php?id=v5:userguide:memcached


adodb-memcache.lib.inc.php
```diff
@@ -21,200 +17,415 @@
...
+	public function connect(&$err)
+	{
+		/*
+		* do we have memcache or memcached? see the note
+		* at adodb.org on memcache
+		*/
+		if (class_exists('Memcache')) 
+			$this->libraryFlag = self::MCLIB;
+		elseif (class_exists('Memcached'))
+			$this->libraryFlag = self::MCLIB;
```
dregad, line 136: @mnewnham shouldn't this be `self::MCLIBD` ?

PR #691, opened 2021-03-09
Cosmetics datadict postgres

peterdd:
Some changes following PSR-2 to make review easier with existing/coming PR this month that have this cosmetics yet:

* proper function cases
* remove space between function name and following '('
* fix some indentation where really wrong
* add { and } to single instruction ifs to avoid future potential bugs
* { on new line for functions
* { on same  line as if, foreach, or switch statements
* add single space between function parameters after ,

### DIFF
--- a/datadict/datadict-postgres.inc.php
+++ b/datadict/datadict-postgres.inc.php
@@ -28,7 +28,7 @@ class ADODB2_postgres extends ADODB_DataDict
 	public $blobAllowsDefaultValue = true;
 	public $blobAllowsNotNull = true;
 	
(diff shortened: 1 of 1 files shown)


dregad:
> Some changes following PSR-2

For the record, [PSR-2 is deprecated](https://www.php-fig.org/psr/psr-2/) and replaced by PSR-12.

peterdd:
I know, I just changed thing that is same on PSR-2 and PSR-2 and improves readability.

PSR-2 is just an orientation for us as we do not follow it bit by bit. We use tabs for line indenting, and I prefer it that way. :-)

Issue #185, opened 2016-01-04
mssqlnative: Query not returning ID


jorgemfm:
Version 5.20.2
mssqlnative Driver

If a textfield in inserted empty, the query does not return the ID:

Ex:
$data['field1'] =1;
$data['field2'] = '';
$conn->AutoExecute($table, $data, 'INSERT');
$conn->Insert_ID(); //does not return ID

$data['field1'] =1;
$data['field2'] = 'string of text';
$conn->AutoExecute($table, $data, 'INSERT');
$conn->Insert_ID(); // returns ID

The problem lies in preg_match on the _query method, but i'm not very good at regex so I was unable to solve it.

On version 5.19 works OK but this versions has this problem: failure on Insert_ID() if the insert statement contains a semicolon in a value string

mnewnham:
based on your above example, do you mean:

``if the insert statement contains a null in a value string``

or should your example 1 say

``$data['field2'] = 'some text with a ; in it';``

jorgemfm:
No, it's really like I said, if I pass an empty string like this: 

$data['field2'] = ''; 

The ID is never returns because its never entrer the condition:

jorgemfm:
No, it's really like I said, if I pass an empty string like this: 

$data['field2'] = '';

The ID is never returned because it never enters the condition:

if(preg_match('......'){
....
}

on the _query() method

dregad:
See #96 for the original issue introducing the new regex. Looking at test code provided by @stekule, it appears the empty string was not considered; I believe the simplest fix would be to add an alternative to match the empty string, i.e. 

```
/^\W*insert\s(?:\'\'|(?:(?:\'\')*\'[^\']+\'(?:\'\')*)|[^;\'])*;?$/i
                ^^^^^ new bit
```

The current regex is not very efficient, as it goes through the matching group for every single char after the initial *insert*. and needs over 400 steps [1] to match a simple statement like `INSERT INTO schema.my_table ( c1, c2, c3 ) VALUES ( '', 'x;', 12 )`.

Leveraging the insert statement's syntax, we can modify the regex to only parse the values. Also, moving the complex group after simpler ones is better. Something like this reduces the number of steps to 72 [1] for the same SQL statement:
```
/^\W*insert[\s\w()",.]+values\s*\((?:[^;\']|\'\'|(?:(?:\'\')*\'[^\']+\'(?:\'\')*))*;?$/i
```

[...]

PR #699, opened 2021-03-13
mssql: enable _insertID() again

### LINKED ISSUE #692: mssqlnative: _insertid() doesn't work anymore
In versions prior to 5.21 the `lastInsertID` was set/retrieved inside
`_query()` for insert statements and `_insertId()` just returned that value.
This worked because appending `; select SCOPE_IDENTITY()` to the original
sql makes it a batch, which surely has the same scope.

I guess using a new query in `_insertid()` to get that information doesn't
work (always `null`) because its a different scope from the insert query.

I run into this while testing MantisBT with ADOdb 5.21, every op which
inserts a record in table1 and needs the id of this record for an insert
in table2 fails.

I suggest to revert (works for me in the test), and look at other problems
with `_query()` (affected rows, triggers, unset `lastInsertID`) later.

I could offer a pull request, but not sure which branch to target.

dregad:
In ADOdb 5.21.0, for performance reasons [1] the _query() method no
longer retrieves the last inserted Id.

This introduced a regression, as it was only possible to retrieve the
Insert Id for "normal" queries, but not parameterized ones as
Insert_ID() always returned null in this case.

To maintain the driver's performance, we now allow the client to
decide whether insert queries should execute the extra SCOPE_IDENTITY()
call, by providing a new enableLastInsertID() method.

Fixes #692

[1]: https://github.com/ADOdb/ADOdb/issues/185#issuecomment-169057606

### DIFF
--- a/adodb.inc.php
+++ b/adodb.inc.php
@@ -496,7 +496,8 @@ if (!defined('_ADODB_LAYER')) {
 	//
 	var $_oldRaiseFn =  false;
 	var $_transOK = null;
-	var $_connectionID	= false;	/// The returned link identifier whenever a successful database connection is made.
+	/** @var resource Identifier for the native database connection */
+	var $_connectionID = false;
 	var $_errorMsg = false;		/// A variable which was used to keep the returned last error message.  The value will
 								/// then returned by the errorMsg() function
 	var $_errorCode = false;	/// Last error code, not guaranteed to be used - only by oci8
--- a/adodb.inc.php
+++ b/adodb.inc.php
@@ -1467,6 +1467,15 @@ if (!defined('_ADODB_LAYER')) {
 	}
 
 	/**
+	 * Enable or disable the Last Insert Id functionality.
+	 *
+	 * If the Driver supports it, this function allows setting {@see $hasInsertID}.
+	 *
+	 * @param bool $enable False to disable
+	 */
+	public function enableLastInsertID($enable = true) {}
+
+	/**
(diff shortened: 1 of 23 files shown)


dregad:
@mnewnham I expected some reaction from you, considering that this reverts the change in default behavior for InsertId that you implemented in 5.21.0, as I explained in https://github.com/ADOdb/ADOdb/issues/692#issuecomment-799221092. 

After merging this, code will have to call `$db->enableLastInsertId(false);` to disable execution of the extra scope_identity() query and get best performance.

May I assume that you're OK with this ? I will of course update the docs in wiki accordingly.

mnewnham:
I was hands-off on this one, I thought you and your mantis guys had it well under control . Is there some specific issue you want me to comment on?
 Also remember I'm not on the git product family every day like yourself and notifications via email (business email regards github notifications as spam) and phone notifications (android keeps disabling them) means that I sometimes miss these things.

dregad:
> Is there some specific issue you want me to comment on?

I was just giving you a heads up to make sure you're aware of this change and its consequences, since you previously asked _that the existing 5.21 remains the no-config default_ [[1]] and the PR actually does the opposite... This may have an impact in your code (i.e. degraded performance) unless you adapt it to call `$db->enableLastInsertId(false);` as needed.

[1]: https://github.com/ADOdb/ADOdb/issues/692#issuecomment-797156377

mnewnham:
Ok, missed that one.

Issue #702, opened 2021-03-13
sqlite: Add support for correct insertion/updating of blob data


mnewnham:
Add the function to enable insertion of binary data into blob fields

PR #703, opened 2021-03-14
Add support for blob handling in SQLite driver. see #702

mnewnham:
Add the updateBlob() and updateBlobFile Methods

### DIFF
--- a/drivers/adodb-sqlite3.inc.php
+++ b/drivers/adodb-sqlite3.inc.php
@@ -646,7 +646,9 @@ class ADODB_sqlite3 extends ADOConnection {
 		$stmt = $this->_connectionID->prepare($sql);
 
 		// Set the first bind value equal to value we want to update
-		$bindOk = $stmt->bindValue(1, $val, SQLITE3_BLOB);
+		if (!$stmt->bindValue(1, $val, SQLITE3_BLOB)) {
+			return false;
+		}
 
 		// Build as many keys as available
 		$bindIndex = 2;
@@ -660,7 +662,11 @@ class ADODB_sqlite3 extends ADOConnection {
 			} else {
 				$type = SQLITE3_TEXT;
 			}
-			$stmt->bindValue($bindIndex, $bindValue, $type);
+
+			if (!$stmt->bindValue($bindIndex, $bindValue, $type)) {
+				return false;
+			}
+
 			$bindIndex++;
 		}
 
@@ -688,6 +694,9 @@ class ADODB_sqlite3 extends ADOConnection {
 
 		// Read file information
 		$fileContents = file_get_contents($path);
+		if ($fileContents === false)
+			// Distinguish between an empty file and failure
+			return false;
 
 		return $this->updateBlob($table, $column, $fileContents, $where, $blobtype);
 	}


drivers/adodb-sqlite3.inc.php
```diff
@@ -596,6 +596,119 @@ function year($fld)
...
+	* be an array (array params) or a string that we will
+	* do our best to unpack and turn ito a prepared statement
+	*
+	* @param string $table
+	* @param string $column
+	* @param string $val
+	* @param mixed  $where
+	* @param string $blobtype ignored
+	*
+	* @return success
```
dregad, line 613: I guess you mean *bool* here

drivers/adodb-sqlite3.inc.php
```diff
@@ -596,6 +596,119 @@ function year($fld)
...
+	/**
+	* SQLite update for blob from a file
+	*
+	* @param string $table
+	* @param string $column
+	* @param string $val filename
+	* @param mixed  $where
+	* @param string $blobtype ignored
+	*
+	* @return success
```
dregad, line 701: bool

drivers/adodb-sqlite3.inc.php
```diff
@@ -596,6 +596,119 @@ function year($fld)
 		$x = "strftime('%Y',$fld)";
 		return $x;
 	}
+	
+	/**
+	* SQLite update for blob 
+	*
+	* SQLite must be a fully prepared statement, so where can either
```
dregad, line 603: Not sure what you mean by _SQLite must be a fully prepared statement_

drivers/adodb-sqlite3.inc.php
```diff
@@ -596,6 +596,119 @@ function year($fld)
...
+		{
+			
+			if (is_integer($v) || is_bool($v) || is_float($v))
+				$bindOk = $stmt->bindValue($bindIndex, $bindValue, SQLITE3_NUM);
+					
+			else if(is_object($v))
+				/*
+				* Assume a blob
+				*/
+				$bindOk = $stmt->bindValue($bindIndex, $bindValue, SQLITE3_NUM);
```
dregad, line 677: If it's assumed to be a blob, shouldn't the type be `SQLITE3_BLOB` ?

drivers/adodb-sqlite3.inc.php
```diff
@@ -596,6 +596,119 @@ function year($fld)
...
+	*
+	* SQLite must be a fully prepared statement, so where can either
+	* be an array (array params) or a string that we will
+	* do our best to unpack and turn ito a prepared statement
+	*
+	* @param string $table
+	* @param string $column
+	* @param string $val
+	* @param mixed  $where
+	* @param string $blobtype ignored
```
dregad, line 611: Why add the parameter, if it's never used ?

drivers/adodb-sqlite3.inc.php
```diff
@@ -596,6 +596,119 @@ function year($fld)
...
+	* SQLite update for blob 
+	*
+	* SQLite must be a fully prepared statement, so where can either
+	* be an array (array params) or a string that we will
+	* do our best to unpack and turn ito a prepared statement
+	*
+	* @param string $table
+	* @param string $column
+	* @param string $val
+	* @param mixed  $where
```
dregad, line 610: The description of _where_ in the above paragraph should be moved here

drivers/adodb-sqlite3.inc.php
```diff
@@ -596,6 +596,119 @@ function year($fld)
...
+				
+			$bindIndex++;
+		}
+		
+		/*
+		* Now execute the update. NB this is SQLite execute,
+		* not ADOdb
+		*/
+		$ok = $stmt->execute();
+		return is_object($ok) ? true : false;
```
dregad, line 689: The `? true : false` bit is unnecessary, is_object already returns bool - just `return is_object($ok)`

drivers/adodb-sqlite3.inc.php
```diff
@@ -596,6 +596,119 @@ function year($fld)
...
+	* @param string $blobtype ignored
+	*
+	* @return success 
+	*/
+	function updateBlobFile($table,$column,$val,$where,$blobtype='BLOB')
+	{
+		$fileContents = file_get_contents($val);
+		/*
+		* Read file information
+		*/
```
dregad, line 708: Comment should go before the file_get_contents() statement

dregad (changes requested):

drivers/adodb-sqlite3.inc.php
mnewnham: We often write code that is "partially" prepared - for example the function I'm addressing looks like this in other drivers:
````php
UPDATE sessions2 SET sessdata=? WHERE sesskey='dsdfsfdsfsfsfsdfsd'", 
````
This isn't acceptable in SQLite for updating blob data. You have to bind **every** variable, i.e. make it "fully" prepared:
````php
UPDATE sessions2 SET sessdata=? WHERE sesskey=?", 

Thats what I'm addressing

File: drivers/adodb-sqlite3.inc.php (hunk 5 above)
Comment (mnewnham) [reply drivers/adodb-sqlite3.inc.php:611]: I'd make joke about making sure that method signatures matches the parent but I'm guessing you didn't realize that this is a core function and that the $blobtype parameter is used elsewhere.

File: drivers/adodb-sqlite3.inc.php (hunk 3 above)
Comment (dregad) [reply drivers/adodb-sqlite3.inc.php:603]: Ah OK I get it, didn't realize this (I don't use sqlite much). For the interest of others who like me are not aware of such technicalities, maybe improve the comment like

`SQLite must be a fully prepared statement, i.e. having all of its variables bound` or something like that

File: drivers/adodb-sqlite3.inc.php (hunk 5 above)
Comment (dregad) [reply drivers/adodb-sqlite3.inc.php:611]: Duh. That's embarrassing :flushed:  That will teach me to respond quickly without looking at the big picture. ignore me...

File: drivers/adodb-sqlite3.inc.php (hunk 6 above)
Comment (mnewnham) [reply drivers/adodb-sqlite3.inc.php:611]: Can I suggest that when you request a change from a programmer that is a little ambiguous, for example here where I think you are asking for a change of format as well as change of location for the text block, you provide a short example of what it should look like or a reference to a standard, e.g. "place your description here, per PSR3" or whatever.  Adding a reference link provides positive reinforcement. This reduces the kind of back-and-forth conversations that can delay commits. 

Personally, just to reiterate, I have no issue if you tweak this stuff yourself and tell me afterwards (if at all).

**Review by dregad — CHANGES_REQUESTED** [review]

File: drivers/adodb-sqlite3.inc.php [hunk 9]
```diff
@@ -596,6 +596,125 @@ function year($fld)
...
+				$bindOk = $stmt->bindValue($bindIndex, $bindValue, SQLITE3_NUM);
+					
+			else if(is_object($v))
+				/*
+				* Assume a blob, this should never appear in
+				* the binding for a where statement anyway
+				*/
+				$bindOk = $stmt->bindValue($bindIndex, $bindValue, SQLITE3_BLOB);
+			else
+				$bindOk = $stmt->bindValue($bindIndex, $bindValue, SQLITE3_TEXT);
```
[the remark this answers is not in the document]
dregad: @mnewnham in case you missed the notification, I'm waiting for your feedback on this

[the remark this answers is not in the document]
mnewnham: Have added some error trapping. updated from master

PR #706, opened 2021-03-14
Ensure  bind parameters are presented in a numeric array see #705

### LINKED ISSUE #705: PDO: Bind parameters fail if sent in associative array
The ADOdb method **param()** can be used to create database-specific bind parameters for use in portable SQL. The resulting values can then be sent to the execute method in either a numeric or associative arrays. example:

````php
$p1 = $db->param('p0');
$bind = array('p0'=>'somevalue');
$db->execute($somesql,$bind);
````
The PDO driver however requires the bind array to be presented in a numeric array and fails otherwise

mnewnham:
Fixes #705

### DIFF
--- a/drivers/adodb-pdo.inc.php
+++ b/drivers/adodb-pdo.inc.php
@@ -565,6 +565,10 @@ class ADODB_pdo extends ADOConnection {
 				$this->_driver->debug = $this->debug;
 			}
 			if ($inputarr) {
+				/*
+				* inputarr must be numeric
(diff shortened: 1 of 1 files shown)


dregad:
Just realized looking at the related issue, that you lined it up for the 5.21.1 milestone, but I merged to master. 

I will backport the change, but maybe next time you can select `hotfix/5.21` as target branch so it goes straight where intended.

Issue #707, opened 2021-03-18
extractSchema() with postgresql fails


peterdd:
Using the extractSchema() of adodb-xmlschema03.inc.php with a postgresql database is not working:

Testscript with existing adodbtest database:

```php
<?php
$adodbdir='ADOdb/';

include_once $adodbdir.'adodb.inc.php';
include_once $adodbdir.'adodb-xmlschema03.inc.php';

/*
$driver='mysqli';
$host='127.0.0.1';
$user='';
$password='';
$database='adodbtest';
*/

$driver='postgres';
$host='127.0.0.1';
$user='';
$password='';
$database='adodbtest';

$db = ADONewConnection($driver);
$db->debug=true;
$db->connect($host, $user, $password, $database);
$schema = new adoSchema($db);

$xml = $schema->extractSchema(true);

echo $xml;
```
```
PHP Warning:  pg_query(): Query failed: ERROR:  argument of WHERE must be type boolean, not type integer
LINE 1: SELECT * FROM flyspray_history WHERE -1
                                             ^ in /srv/www/htdocs/ADOdb/drivers/adodb-postgres7.inc.php on line 259
-1: ERROR:  argument of WHERE must be type boolean, not type integer
LINE 1: SELECT * FROM flyspray_history WHERE -1
                                             ^
	ADOConnection._Execute(SELECT * FROM flyspray_history WHERE -1, false)% line 1383, file: /srv/www/htdocs/ADOdb/adodb.inc.php
ADOConnection.Execute(SELECT * FROM flyspray_history WHERE -1)% line 2095, file: /srv/www/htdocs/ADOdb/adodb-xmlschema03.inc.php
adoSchema.extractSchema(true)% line   25, file: /srv/www/htdocs/extractsch.php
```

Fix is probably rewriting the query so the WHERE clause is always a boolean false if we are just interested in field definition, not data, right?

```
adodbtest=# create table adodbtest(id int);
CREATE TABLE
adodbtest=# insert into adodbtest values (10),(20),(30);
INSERT 0 3
adodbtest=# select * from adodbtest where -1;
ERROR:  argument of WHERE must be type boolean, not type integer
LINE 1: select * from adodbtest where -1;
                                      ^
adodbtest=# select * from adodbtest where 0=1;
 id 
----
(0 rows)
```

PR #708, opened 2021-03-18
fix extractSchema() xmlschema03 for postgresql

peterdd:
and probably more database types

fix for #707


dregad:
Fixed in 97442e1a

PR #717, opened 2021-04-05
Fixed function declaration to match parent

edgardmessias:
Fixed message "Declaration of ADODB_pdo::metaIndexes($table, $normalize = true) should be compatible with ADOConnection::MetaIndexes($table, $primary = false, $owner = false)"


dregad:
Thank you

nickano:
I would be super grateful if this can be hotfixed into 5.21.1 or similar?  Getting warnings on every DB connection when using PDO atm

thanks for fixing btw!

dregad:
@nickano sorry I missed your message. Will backport it now.

Issue #710, opened 2021-03-25
firebird: Update ADOdb to support latest Firebird drivers 


mnewnham:
Test and update the **firebird** ADOdb driver to support the [Firebird SQL Version 1.1.1](https://github.com/FirebirdSQL/php-firebird/releases/tag/v1.1.1) PHP driver and test with PHP7.4

PR #720, opened 2021-04-08
New release of Firebird Driver see #710

mnewnham:
This new release of the driver for Firebird is designed for use with the latest PHP7 drivers.

### DIFF
--- a/drivers/adodb-firebird.inc.php
+++ b/drivers/adodb-firebird.inc.php
@@ -299,7 +299,7 @@ class ADODB_firebird extends ADOConnection {
 	*
 	* @return array|bool An array of the indexes, or false if the query to get the indexes failed.
 	*/
-	function metaIndexes($table, $primary = false, $owner = false)
+	public function metaIndexes($table, $primary = false, $owner = false)
 	{
 		// save old fetch mode
 		global $ADODB_FETCH_MODE;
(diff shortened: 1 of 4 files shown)


drivers/adodb-firebird.inc.php
```diff
@@ -255,37 +355,65 @@ function &MetaIndexes ($table, $primary = FALSE, $owner=false)
...
-		$seqname = strtoupper($seqname);
-		return $this->Execute("DROP GENERATOR $seqname");
-	}
-
-	function GenID($seqname='adodbseq',$startID=1)
+		$sql = sprintf($this->_genSeqSQL,$seqname,$startID);
+		$ok = $this->Execute($sql);
+		
+		if (!$ok) 
+			return false;
```
dregad, line 398: Aren't we missing a `return $ok;` statement here ? The PHPDoc says the function returns a recordset.

drivers/adodb-firebird.inc.php
```diff
@@ -255,37 +355,65 @@ function &MetaIndexes ($table, $primary = FALSE, $owner=false)
...
+	
+	/**
+	 * A portable method of creating sequence numbers.
+	 *
+	 * @link https://adodb.org/dokuwiki/doku.php?id=v5:reference:connection:genid
+	 *
+	 * @param string $seqname (Optional) The name of the sequence to use.
+	 * @param int $startID (Optional) The point to start at in the sequence.
+	 *
+	 * @return bool|int|string
```
dregad, line 409: Looks like we're only returning _int_ here

drivers/adodb-firebird.inc.php
```diff
@@ -307,23 +435,36 @@ function SelectDB($dbName)
...
+	/**
+	 * Prepares an SQL statement and returns a handle to use.
+	 * This is not used by bound parameters anymore
+	 *
+	 * @link https://adodb.org/dokuwiki/doku.php?id=v5:reference:connection:prepare
+	 * @todo update this function to handle prepared statements correctly
+	 *
+	 * @param string $sql The SQL to prepare.
+	 *
+	 * @return string The original SQL that was provided.
```
dregad, line 465: The function actually returns an array

drivers/adodb-firebird.inc.php
```diff
@@ -610,105 +816,106 @@ function UpdateBlobFile($table,$column,$path,$where,$blobtype='BLOB')
...
+	* Returns a portably-formatted date string from a timestamp database column.
+	*
+	* @link https://adodb.org/dokuwiki/doku.php?id=v5:reference:connection:sqldate
+	*
+	* Firebird does not support an AM/PM format, so the A indicator always shows AM
+	*
+	* @param string $fmt The date format to use.
+	* @param string|bool $col (Optional) The table column to date format, or if false, use NOW().
+	*
+	* @return bool|string The SQL DATE_FORMAT() string, or false if the provided date format was empty.
```
dregad, line 874: Unless I missed something, the function never returns _false_

dregad (commented):
I noted a few minor issues / inconsistencies in PHPDoc

mnewnham:
Ok, thanks for the review. It may be that some of these existed already and I didn't notice. I'll check them later

drivers/adodb-firebird.inc.php
mnewnham: should return a recordset

drivers/adodb-firebird.inc.php
mnewnham: Updated doc

drivers/adodb-firebird.inc.php
mnewnham: Returns an array, or false if the preparation fails

drivers/adodb-firebird.inc.php
mnewnham: The docblock for this is misleading for this method is misleading in both the prototype and a number of the drivers

PR #723, opened 2021-04-10
Fix #715 by removing subqueries without using regex

### LINKED ISSUE #715: PageExecute bug - pdo error: [1248: Every derived table must have its own alias]
Database: mysql
Driver: pdo
```php
$dbconn->PageExecute("
    select a.*, count(distinct notes.id) as notes_cnt, count(distinct images.id) as images_cnt
                    from locations a
                        left join locations_notes notes
                            on notes.locations_id = a.id
                        left join locations_images images
                            on images.locations_id = a.id
                    where a.status = 'active'
                    group by a.id
", 25, 1);
```

"count" query is transformed into 
```sql
SELECT COUNT(*) FROM (
    select a.*, count(distinct notes.id) as notes_cnt, count(distinct images.id) as images_cnt
                    from locations a
                        left join locations_notes notes
                            on notes.locations_id = a.id
                        left join locations_images images
                            on images.locations_id = a.id
                    where a.status = 'active'
                    group by a.id
)
```
and throws Uncaught ADODB_Exception: pdo error: [1248: Every derived table must have its own alias] in EXECUTE("...")
as you can see - no alias is added

dregad:
Alternative to #718

### DIFF
--- a/adodb-lib.inc.php
+++ b/adodb-lib.inc.php
@@ -426,12 +426,32 @@ function _adodb_getcount(&$zthis, $sql,$inputarr=false,$secs2cache=0)
 		}
 
 	} else {
-		// now replace SELECT ... FROM with SELECT COUNT(*) FROM
-		if ( strpos($sql, '_ADODB_COUNT') !== FALSE ) {
-			$rewritesql = preg_replace('/^\s*?SELECT\s+_ADODB_COUNT(.*)_ADODB_COUNT\s/is','SELECT COUNT(*) ',$sql);
-		} else {
-			$rewritesql = preg_replace('/^\s*SELECT\s.*\s+FROM\s/Uis','SELECT COUNT(*) FROM ',$sql);
+		// Replace 'SELECT ... FROM' with 'SELECT COUNT(*) FROM'
+		// Parse the query one char at a time starting after the SELECT
+		// to find the FROM clause's position, ignoring any sub-queries.
+		$start = stripos($sql, 'SELECT') + 7;
+		if ($start === false) {
+			// Not a SELECT statement - probably should trigger an exception here
+			return 0;
 		}
+		$len = strlen($sql);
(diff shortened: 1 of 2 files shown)


dregad:
@mnewnham as mentioned in https://github.com/ADOdb/ADOdb/pull/718#issuecomment-817215690, it would be nice if you could take a look at this.

mnewnham (approved):
This seems all good. I was unable to break it with
 - Subselects within the 'select from'
 - Subselects within the 'where'
 - Unmatched parentheses inside strings within the statements

dregad:
Great, thanks for testing.

Mike-Benoit:
While I haven't tested this yet, I disagree with removing the _ADODB_COUNT functionally to override the built-in removal of sub-selects. Without a full SQL parser (which will never happen of course), I fail to see how we can handle every possible SQL query that anyone could ever possibly write across all drivers. It seems like pretty cheap insurance to keep around, and even a potential optimization for queries that may be executed often too.

Based on a very quick read of the code, it would appear that the following real-world query (where _ADODB_COUNT is currently used) could possibly fail:
```
select
_ADODB_COUNT
a.*,
(
CASE WHEN EXISTS
( select 1 from table1 as y where y.accrual_policy_account_id = a.id and y.deleted = 0)
THEN 1
ELSE
CASE WHEN EXISTS
( select 1 from table2 as z where z.accrual_policy_account_id = a.id and z.deleted = 0)
THEN 1
ELSE 0
END
END
) as in_use
_ADODB_COUNT
FROM table3 as a ...;
```

dregad:
@Mike-Benoit given the above SQL statement (with or without the `_ADODB_COUNT` markers), the new parser code in _adodb_getcount() correctly transforms it to `SELECT COUNT(*) FROM table3 as a`, which is the expected result.

PR #730, opened 2021-05-02
mysql-affected-rows incorrectly returns rowcount for select statement…

mnewnham:
…s, see #604

The mysql affected_rows works as a rowcount() function when called against a select statement. It should return false for ADOdb statements.


dregad:
This duplicates #729. A separate PR is not necessary, master branch will be updated when hotfix/5.21 is merged.

Issue #687, opened 2021-03-08
Caching in FieldTypesArray() causes problems


dregad:
The static variable added in #367 to cache field information introduces a regression, [as mentioned](https://github.com/ADOdb/ADOdb/pull/367#issuecomment-346954017) by @Mike-Benoit.

This [affects MantisBT](https://mantisbt.org/bugs/view.php?id=28015#c65202), when running on PostgreSQL.

## Reproducing the issue

The following code snippet consistently reproduces the problem (tables are from a fresh install of MantisBT, but results are the same with any 2 distinct SQL statements):

``` php
<?php
include 'adodb.inc.php';

$db = ADONewConnection('pgsql');
$db->connect('localhost', $u, $p, 'bugtracker');

q('SELECT basename, priority, protected FROM m_plugin');
q('SELECT config_id, value FROM m_config');

function q($sql) {
    global $db;
    $rs = $db->execute($sql);
    echo "--------- $sql\n";
    print_r($rs->fieldTypesArray());
}
```

### Actual Output
The 2nd call to _fieldTypesArray()_ returns information about the 1st query's fields.
```
--------- SELECT basename, priority, protected FROM m_plugin
Array
(
    [0] => ADOFieldObject Object
        (
            [name] => basename
            [max_length] => -1
            [type] => varchar
        )

    [1] => ADOFieldObject Object
        (
            [name] => priority
            [max_length] => 4
            [type] => int4
        )

    [2] => ADOFieldObject Object
        (
            [name] => protected
            [max_length] => 1
            [type] => bool
        )

)
--------- SELECT config_id, value FROM m_config
Array
(
    [0] => ADOFieldObject Object
        (
            [name] => basename
            [max_length] => -1
            [type] => varchar
        )

    [1] => ADOFieldObject Object
        (
            [name] => priority
            [max_length] => 4
            [type] => int4
        )

    [2] => ADOFieldObject Object
        (
            [name] => protected
            [max_length] => 1
            [type] => bool
        )

)
```

### Expected output for the 2nd query
```
Array
(
    [0] => ADOFieldObject Object
        (
            [name] => config_id
            [max_length] => -1
            [type] => varchar
        )

    [1] => ADOFieldObject Object
        (
            [name] => value
            [max_length] => -1
            [type] => text
        )

)
```

joebordes:
[...]

PR #731, opened 2021-05-04
fieldTypesArray caches with class variable instead of static, see #687

mnewnham:


dregad:
- [ ]  Note to self: compare this with my local, work-in-progress branch before merging.

dregad:
This is pretty much the same as what I had coded on my side, so I'm closing this.

PR #742, opened 2021-07-25
Empty string will not work if TCP is not available

mason-chase:


mnewnham:
Thank you for your submission. Could you please open an issue, with a more detailed description of the problem that your request addresses, for example any  error messages, PHP Version, platform etc, then link your pull request to that

Briefly reviewing your code, I notice that the line changed is tagged thus: 
````php
var $socket = ''; //Default to empty string to fix HHVM bug
````
1. I have at this point no clue what HHVM is, but before we pull this, we would need to understand whether this is still an issue or not.
2. The value you have chosen for this ``/var/run/mysqld/mysql.sock`` appears to be in conflict with the default ``/tmp/mysql.ini`` as [described here](https://mariadb.com/kb/en/local-socket-communication-path-in-myini/)
3. The value in this variable is also used in Windows to describe the name of the named pipe used for connection listening.  This might break with your suggested change,

Regards

Mark

peterdd:
https://en.wikipedia.org/wiki/HHVM

dregad:
> I have at this point no clue what HHVM is, but before we pull this, we would need to understand whether this is still an issue or not.

Considering that HHVM does not and will not support PHP 7 and later, I would say it is not, and this comment can probably be deleted together with any HHVM-specific code. Maybe @Mike-Benoit will care to comment, as he was the one who introduced this compatibility code back in 2014.

@mason-chase are you planning to address Mark's comments ?

dregad:
@mason-chase Since you have not given any feedback, I'm closing this PR. 

Feel free to reopen it with the requested information and fix the merge conflicts.

Issue #733, opened 2021-05-31
pdo:  Throws Notice, Returns 0 when rows affected, on Affected_Rows() call


ooglek:
https://github.com/ADOdb/ADOdb/blob/199391f639a0b3346a80d66c00c87230fa3e6a07/drivers/adodb-pdo.inc.php#L611

Worked in 5.20.18, does not work in 5.21.0.

Error in PHP when attempting to call `$adodb->affected_rows()` after an `$adodb->execute()` call using the pdo_mysql driver.

> PHP Notice:  Undefined property: ADODB_pdo_mysql::$_stmt in /usr/local/apps/include/vendor/adodb/adodb-php/drivers/adodb-pdo.inc.php on line 611

pdo_mysql does NOT have an `_affectedrows()` method, so it attempts to return a value from the `_stmt` but for some reason it doesn't exist:

https://github.com/ADOdb/ADOdb/blob/199391f639a0b3346a80d66c00c87230fa3e6a07/drivers/adodb-pdo.inc.php#L606-L612

I also get a warning when initializing ADOdb.

> PHP Warning:  Declaration of ADODB_pdo::metaIndexes($table, $normalize = true) should be compatible with ADOConnection::MetaIndexes($table, $primary = false, $owner = false) in /usr/home/USER/dev/trunk4/include/vendor/adodb/adodb-php/drivers/adodb-pdo.inc.php on line 0

I can recreate this using `psysh` --

```
>>> $db->execute("select * from numbers limit 3");
=> ADORecordSet_pdo {#2689
     +bind: false,
     +databaseType: "pdo",
     +dataProvider: "pdo",
[... cut ...]
     +sql: "select * from numbers limit 3",
     +connection: ADODB_pdo {#2586
       +databaseType: "pdo",
       +dataProvider: "pdo",
       +fmtDate: "'Y-m-d'",
       +fmtTimeStamp: "'Y-m-d H:i:s'",
       +replaceQuote: "''",
       +hasAffectedRows: true,
       +_bindInputArray: true,
       +_genIDSQL: "UPDATE %s SET id=LAST_INSERT_ID(id+1);",
       +_genSeqSQL: "CREATE TABLE  if NOT EXISTS %s (id int not null)",
       +_dropSeqSQL: "drop table %s",
       +_autocommit: true,
       +_lastAffectedRows: 0,
       +_errormsg: false,
       +_errorno: false,
       +dsnType: "mysql",
       +stmt: false,
[...]

PR #744, opened 2021-08-02
First attempt at a refactor of method_exists

ooglek:
Apologies for the tabs/vs/spaces issue, I can revert and put back, just not sure what the coding standards are (I'll go look).

First attempt to refactor `method_exists` away since it is broken given its intent.

Background: Issue #733 was caused by the lack of `metaIndexes()` being defined in drivers/adodb-pdo_mysql.inc.php -- this should have been fine if `method_exists` worked as written, but since both `ADOConnection` and `ADODB_pdo` classes define `metaIndexes()`, `method_exists` will always return true, even if the PDO-derived driver also defines the method.

I'm not sure why it is implemented in `ADODB_pdo` as it is. It seems like the underlying PDO driver should use inheritance to override the parent methods, as the parent method would be called if the driver did not define the method. But I didn't look deeply here...

I've written a small helper method that, if the "Late Static Binding" class name defines the method in question, return it, otherwise return false.

This simplifies a bunch of code. 

However, for the most part, it seems strange that those methods exist at all, but for a few, as inheritance overrides should work in most of the cases I've modified.


mnewnham:
Obvious first question. Am I misunderstanding something? You have this method callled ``_callChildMethodIfExists``, but everywhere you use it, you call it as an external function, rather than as a method, e.g.

````php
 public function metaProcedures($procedureNamePattern = null, $catalog  = null, $schemaPattern  = null) {
        return _callChildMethodIfExists(__FUNCTION__, func_get_args());
    }
````

Instead of
````php
 public function metaProcedures($procedureNamePattern = null, $catalog  = null, $schemaPattern  = null) {
        return $this->_callChildMethodIfExists(__FUNCTION__, func_get_args());
    }
````

**Comment by mnewnham** [comment]
So I did a quick search and replace to fix the above issue and it seems to work pretty well. I also tried it with a couple of other databases and that worked out ok.

**Comment by mnewnham** [comment]
Here's another idea. Instead of a whole complex dynamic testing of whether a child method exists, why not just fill in the missing methods so that we **always** know that a method is going to be there, even if it is empty and is simply returning false?

**Comment by ooglek** [comment]
That's totally fine! It just needs to be done. 

Probably makes sense to write a few tests around it, and make an interface that forces them to exist.

The challenge is that a few of the parent methods do things, and I don't know if the child method needs to implement those things too, or do its own thing AND THEN call the parent method too.

I also don't know what might break.

**Comment by ooglek** [comment]
If the decision has been made to go the inheritance route and get rid of all of this method_exists gunk, sweet! I'll close this out.

**Comment by dregad** [comment]
> why not just fill in the missing methods so that we always know that a method is going to be there, even if it is empty and is simply returning false?

@mnewnham just to make sure I understand your suggestion correctly, you are saying we should define every ADODB_pdo method currently relying on method_exists(), in each of the child driver classes ? If so, wouldn't that result in duplicating code so the driver calls the base ADOConnection method ?

**Comment by dregad** [comment]
@ooglek FYI I force-pushed your *hotfix/5.21* branch after fixing the whitespace and a couple of other issues, and rebasing the branch on top of latest master, so the code could actually be reviewed.

**Comment by dregad** [comment]
I didn't actually test the code, but just looking at it I am concerned that *_callChildMethodIfExists()* method's returning _false_ will cause unexpected behavior when _false_ is a legit return value for the child method, which is the case at least for the transaction methods.

In this case, ADODB_pdo will call the parent method which is semantically incorrect, and will at least be a problem with commitTrans() since the stub returns true. 

So assuming driver::commitTrans() returns false because the transaction could not be committed, ADODB_pdo will call ADOConnection::commitTrans() which will happily return **true** and the client will think the transaction was committed successfully.

So I think we need to either make the _callChildMethodIfExists() function throw an exception when the child function is not found, or change its name and purpose to *_childMethodExists()* and let the caller execute call_user_func_array() if it returns true or do whatever is appropriate if not.

Anyway before doing anything, let's wait and hear what @mnewnham answers to my earlier question about inheritance.

**Comment by mnewnham** [comment]
The problem I see with the PDO driver, and I don't really understand why it was done like this, is that we cannot really use the inheritance from the base class to seamlessly use the parent. So instead of this:

````php
class PDO_mysql extends ADOdb_connection
{
     function metaIndexes() {
     /*
       * Do something PDO specific
       * /
     }
}

$db = new PDO_mysql;

$result = $db->metaIndexes(); // Get something PDO specific

$result = $db->metaTables(); // No Generic method defined , call parent
````
It actually works like this:

````php
class PDO extends ADOdb_connection
{
  function connect(){
      $myDriver = new ADOdb_PDO_mysql;
      $myDriver->connect;
      return $myDriver;
 }       
 function metaIndexes() {
     /*
     * See if driver method exists
     */
     if (method_exists($myDriver::metaIndexes)
                /*
                * Call the child class
                */
       else
              /*
              * If there is no PDO level code then call the parent class
              */
          else
             /*
            * Run the parent level code
            */
    }
}

class ADOdb_PDO_mysql extends ADOdb_pdo
{
   function metaIndexes() {
       /*
         * Do something PDO_mysql specific
      * /
    }
}

$db = new PDO
$db->$connect('mysql');

$result = $db->metaIndexes(); // Get something PDO specific

$result = $db->metaTables(); // No Generic method defined , call parent
````

So in order to ensure that the PDO driver does not crash when we call a method that does not exist at the child level, we must do one of 2 things:
1. Call a function as suggested by @ooglek for every defined method if the ADOdb driver. 
     - Upsides: Only needs work at the PDO level
     - Downsides: Incorrect returns for specific methods need to be customized, overhead of reading object internals with every call
2. Fill in every method of every driver with at least a prototype
    - Upsides: Creates a true child/parent object relationship, fast
    - Downsides: Tedious to achieve.

My opinion is that both of these methods suck. I only suggested the second because it is, no matter how awful, a valid alternative. What I would like to happen is:
3. A magically awesome PHP coder appears and says, "look, here is how you can replace an instantiated class with another, simple, really".
But I don't think it's going to happen. It's just a question of choosing the least worst option.

**Comment by ooglek** [comment]
Isn't it the whole point of inheritance to do the highest level at which it is defined?

If I instantiate pdo_mysql, there are three places a method could be defined:
1. in pdo_mysql
2. in pdo
3. in generic

It is in pdo_mysql that you decide HOW you are going to implement a method. If the pdo version is fine, do NOT define it in the pdo_mysql assuming it is a subclass of pdo. 

You can optimize at each level, allowing new subclasses to implement the methods as needed for the inherited class.

There shouldn't be any checking to see if it exists in the class -- it's gonna exist in the class in generic, unless you are implementing something in pdo that doesn't exist in generic.

We come back to the humble `interface` -- the place that requires all things branching from that to have a set implementation.

I can attempt to resubmit, unless this has been solved. I haven't attempted to update the ADODB driver since I backed out of the current version, so mostly I've forgotten about this (sorry!)

**Comment by ooglek** [comment]
Basically what **SHOULD** be done is

1. Implement every method in ADOConnection. Have it default to a certain behavior.
2. Override every method needed in the PDO driver to ensure proper function.
3. Override every method needed in the PDO_MYSQL driver (et al) to ensure proper function.

How close is that?

metaTables()
1. ADOConnection line 2999: `function MetaTables($ttype=false,$showSchema=false,$mask=false)`
2. PDO line 298: `function MetaTables($ttype=false,$showSchema=false,$mask=false)` (but this should be deleted)
3. PDO_MYSQL line 142: `function MetaTables($ttype=false, $showSchema=false, $mask=false)`

metaIndexes()
1. ADOConnection line 3135: `function MetaIndexes($table, $primary = false, $owner = false)` (returns false always)
2. PDO line 308: `public function metaIndexes($table,$normalize=true)` (should be deleted, calls child, silly)
3. PDO_MYSQL line 70: `function metaIndexes($table, $primary = false, $owner = false)`

So, I guess what I should do is delete the PDO implementation of metaIndexes() and MetaTables() and remove the child crap and see what happens. 😄

**Comment by ooglek** [comment]
I've removed a bunch of unnecessary methods in the `PDO` driver. I can test the `pdo_mysql` driver with these changes, but I don't have access to all the other `PDO_` derived drivers to test that they also continue to function.

**Comment by Mike-Benoit** [comment]
Have you benchmarked the following reflection code:
```
$refl = new \ReflectionMethod($this->_driver, $func);
		if ($refl->getDeclaringClass()->getName() == get_class($this->_driver)) {
```

It might be much improved in newer PHP versions, or I could be perpetuating an urban myth ( https://stackoverflow.com/questions/3937671/is-it-bad-to-use-reflection-in-production-code ) but my understanding Reflection related code in PHP is quite slow in comparison to other approaches.

PR #746, opened 2021-08-14
Fix incorrect handling of $ADODB_QUOTE_FIELDNAMES

### LINKED ISSUE #721: Mysql autoExecute() update doesnt work anymore in v5.21.0
When update sql  is  generated  it  puts  brackets  instead  native  quotes

see     
 vendor\adodb\adodb-php\adodb-lib.inc.php       line  681

```
if ((strpos($upperfname,' ') !== false) || ($ADODB_QUOTE_FIELDNAMES)) {
	switch ($ADODB_QUOTE_FIELDNAMES) {
		case 'BRACKETS':
			$fnameq = $zthis->leftBracket.$upperfname.$zthis->rightBracket;break;
		case 'LOWER':
			$fnameq = $zthis->nameQuote.strtolower($field->name).$zthis->nameQuote;break;
		case 'NATIVE':
			$fnameq = $zthis->nameQuote.$field->name.$zthis->nameQuote;break;
		case 'UPPER':
```

The  ADODB_QUOTE_FIELDNAMES variable  is  **true**   so BRACKETS is  selected

Anyway  it  was  wrong  before    but BRACKETS case wasn't there,  so always  LOWER was  selected  so  it  worked

Thanks

dregad:
This fixes #721.

@leon-mbs if you are able to test and confirm this indeed resolves your problem, your feedback will be appreciated

### DIFF
--- a/adodb-active-record.inc.php
+++ b/adodb-active-record.inc.php
@@ -19,6 +19,8 @@
  * @copyright 2014 Damien Regad, Mark Newnham and the ADOdb community
  */
 
+include_once(ADODB_DIR.'/adodb-lib.inc.php');
+
 global $_ADODB_ACTIVE_DBS;
 global $ADODB_ACTIVE_CACHESECS; // set to true to enable caching of metadata such as field info
 global $ACTIVE_RECORD_SAFETY; // set to false to disable safety checks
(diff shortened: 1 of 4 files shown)


leon-mbs:
It seems  work
I will  use it in my project  for hard  testing

PR #753, opened 2021-08-22
perf: fix invalid SQL statement

peterdd:


dregad:
Out of curiosity, did you actually hit that bug ? 

According to git blame, this SQL statement has been like this since the oldest recorded commit in the AODdb repository (3b2d2a3d4848869ed34741987e25999a5049b37d)...

peterdd:
No, I just loosely eye scanned throught code if I see obvious typos when I saw a typo on adodb.org and while fixing it looked into the real code.

PR #758, opened 2021-09-27
Update MetaIndexes() for reserved words

Unifex:
Update MetaIndexes() so it will play nice with table names that are also reserved words.


Unifex:
Found another one.  Will submit both in a single request.

PR #759, opened 2021-09-27
mysql: Handle tables that are reserved words

Unifex:
Wrapping table names represented by `%s` in backticks so tables names that are also reserved words don't cause issues.

These 2 locations have impacted our project.  There are more in this file but they have not impacted us yet so I have left them untouched.

I also updated the call to `$this->Execute()` in the first case.  PHP is case insensitive to this, but not all IDE's are so this should help those.


dregad:
> There are more in this file but they have not impacted us yet so I have left them untouched.

By all means, feel free to fix them too

dregad (approved):

PR #760, opened 2021-09-27
is_array() is expensive

Unifex:
The call to `pg_fetch_array()` returns `false` in all cases where we wouldn't need or want to potentially `_fixblobs()` and return `true`.

A straight up boolean check is all that is needed here.

### DIFF
--- a/drivers/adodb-postgres64.inc.php
+++ b/drivers/adodb-postgres64.inc.php
@@ -1009,6 +1009,10 @@ class ADORecordSet_postgres64 extends ADORecordSet{
 
 	function _fixblobs()
 	{
+		// Check prerequisites and bail early if we do not have what we need.
+		if (!isset($this->_blobArr)) return false;
+		if ($this->fields === false) return false;
+
 		if ($this->fetchMode == PGSQL_NUM || $this->fetchMode == PGSQL_BOTH) {
 			foreach($this->_blobArr as $k => $v) {
 				$this->fields[$k] = ADORecordSet_postgres64::_decode($this->fields[$k]);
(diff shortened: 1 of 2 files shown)


dregad (changes requested):
Thanks for your contribution.

I have no objections with the change, but considering that this usage of is_array() is also present in the _ADORecordSet_assoc_postgres7_ as well as in the parent _ADORecordSet_postgres64_ classes' _fetch() and MoveNext() methods, I believe it should be changed there as well.

Unifex:
I'll take a run at that this weekend.  Thanks for spotting the other spots.

Unifex:
Hello @dregad, sorry for not getting to this earlier.  Life happens unfortunately.  Upside though, I got some time from my employer to finish this up.  Looking at the other places this would have been updated resulted in more refactoring than anticipated but I think it makes things much more readable and removes a lot of repeated code.

I was also thinking that the request to `$this->fields = @pg_fetch_array($this->_queryID,$this->_currentRow,$this->fetchMode);` could also be moved into `_fixblobs()` but at this point that method would be doing a lot more than what its name suggests.  If there is confidence that third party code isn't reliant on it it could be renamed to `_prepfields()` and the `pg_fetch_array()` line moved there too.

Unifex:
@dregad, I went ahead and pushed an extra branch to explore the `_prepfields()` approach should you want to go down that route.  It is on #767.

dregad:
> sorry for not getting to this earlier. Life happens unfortunately.

No worries... Trust me, I know what it's like :wink: 

As a matter of fact, I got delayed as well and did not release 5.22.0 last week-end as planned either... I'll review your PRs, hopefully by tomorrow.

dregad:
Closed in favor of #767

Issue #747, opened 2021-08-14
Move caching to properties of ADORecordSet class


dregad:
Follow-up on https://github.com/ADOdb/ADOdb/issues/687#issuecomment-798937424

> Adapt mssqlnative and oci8 drivers to store their cache in a property of the parent class.

PR #773, opened 2021-12-05
standardize caching of properties of ADORecordSet class #747

mnewnham:
The method of retrieval of the field properties of a recordset now has a standardized method, ``adoRecordset::setFieldObjectsCache()``, which is called immediately after initial recordset retrieval via ``init()``. The objects retrieved are stored in a parent level variable, ``adoRecordset::fieldObjectsCache``.  Calls to ``adoRecordSet::fetchfield()`` now simply read this cache instead of re-querying the database. 

 - The cache persists for the life of the recordset
 - This methodology is usable for all drivers
 - Initially covers the following drivers: db2, oci8, mssqlnative, pgsql, firebird

### DIFF
--- a/adodb.inc.php
+++ b/adodb.inc.php
@@ -3899,6 +3899,14 @@ class ADORecordSet implements IteratorAggregate {
 	 */
 	protected $fieldObjectsCache;
 
+	/*
+	* Indicates if the fieldObjects have been retrieved. Prevents
+	* multiple attempts even if the feature is unavailable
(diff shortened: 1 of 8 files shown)


dregad (changes requested):
Hello Mark, I reviewed and performed limited testing on this, see my review comments.

Be careful if you're going to add new commits, as I have rebased the branch on top of latest master, and added one commit to fix whitespace issues.

PR #774, opened 2021-12-07
Fix php syntax errors found with phpstan

dianne-tennent:
Working on an open source project that uses ADOdb.
Found these errors when analysing with phpstan in
preparation for PHP8 upgrade.


mnewnham:
Thanks

PR #782, opened 2021-12-29
Add PostgreSQL standard datatypes for UUID, NUMERIC, DECIMAL, FLOAT4 …

Mike-Benoit:
Add a few more standard PostgreSQL datatypes for UUID, NUMERIC, DECIMAL, FLOA4 and FLOAT8.

I know ADODB supports custom types now, however since these are standard and have been in PostgreSQL for decades, they should probably be standard in ADODB too.


dregad:
OK for me. @mnewnham any objections against merging this ?

mnewnham:
As we are talking about actual types then this is fine

Issue #772, opened 2021-11-27
core: parameters of ADOrecordset __construct() are incompatible with drivers and other classes


mnewnham:
ADOrecordset::__construct() parameters are defined as ``mixed $queryId``, while:

1. the constructors of all drivers that extends this class are defined as ``mixed $queryId, optional mixed $mode``
2. The constructor of ADOrecordset_array() which extends ADOrecordset is defined as ``optional int``

The parent and all classes that extend ADOrecordset should accept parameters of ``mixed $queryId, option mixed $mode``

mnewnham:
See #[57186f90](https://github.com/ADOdb/ADOdb/commit/57186f905c9a1e49a010daf582baa9a07be826bb) for fix info. (Labelled incorrectly)

PR #785, opened 2021-12-29
Fix PHP FATAL: Uncaught ArgumentCountError: Too few arguments

Mike-Benoit:
When calling _rs2serialize() in adodb-csvlib.inc.php directly, it tries to instantiate a new 'ADORecordSet_array' object, which now requires the first parameter to be specified as of commit: 57186f905c9a1e49a010daf582baa9a07be826bb -- Prior to that it just defaulted to "1". 

Not sure why the default was removed, but we could put it back instead, or make this proposed change. Not sure if "1" or "-1" makes any real difference, but I seem to recall seeing "-1" used in other places.

### DIFF
--- a/adodb-csvlib.inc.php
+++ b/adodb-csvlib.inc.php
@@ -74,7 +74,7 @@ $ADODB_INCLUDED_CSV = 1;
 
 		$savefetch = isset($rs->adodbFetchMode) ? $rs->adodbFetchMode : $rs->fetchMode;
 		$class = $rs->connection->arrayClass;
-		$rs2 = new $class();
+		$rs2 = new $class( -1 );
 		$rs2->timeCreated = $rs->timeCreated; # memcache fix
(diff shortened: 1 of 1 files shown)


dregad:
> When calling _rs2serialize() in adodb-csvlib.inc.php directly

You probably know this, but just for the record, ADOdb library functions prefixed by `_` are for internal use and should not be called by userland code.

Anyway, considering the error occurs also when calling ADOConnection::cacheExecute(), this is a valid bug (regression introduced by 57186f905c9a1e49a010daf582baa9a07be826bb).

> Not sure why the default was removed

Standardizing ADORecordSet constructor signatures, see #772

dregad:
Fixed in a61b2ce4a4570fed450810edafcefdd47c616c9b

dregad:
Removing the Milestone, which is not needed because technically this change is part of #772.

PR #787, opened 2022-01-02
Need support in adodb.inc.php to properly handle new types (ie: UUID, FLOAT4, ...)

Mike-Benoit:
Need support in adodb.inc.php to properly handle UUID, JSON, JSONB, FLOAT4, FLOAT8 types.

This is a missing part that goes with #782

### DIFF
--- a/adodb.inc.php
+++ b/adodb.inc.php
@@ -4876,7 +4876,10 @@ class ADORecordSet implements IteratorAggregate {
 			'BPCHAR' => 'C',
 			'CHARACTER' => 'C',
 			'INTERVAL' => 'C',  # Postgres
-			'MACADDR' => 'C', # postgres
+			'MACADDR' => 'C', # Postgres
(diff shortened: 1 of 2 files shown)


mnewnham:
I'm not sure I understand why we need to push postgres specific types into adorecordset::metatypes() when adorecordset_recordset_postgres64::metatypes () exists

Mike-Benoit:
This list already has a few PGSQL specific types, though most of what is added in this MR isn't actually specific to PGSQL, and because the default metatype is numeric in ADODB, it errors on the side of things not working and would potentially catch people off guard. These types are pretty universal and will be used more often in the future, so I think it makes sense to to support them out of the box. (ie: UUID, JSON). 

JSONB is one that could probably be removed, as currently that is specific to just PGSQL, but I figured since its handled exactly like JSON it wouldn't hurt to have it here.

Let me know what you think.

PR #802, opened 2022-02-12
oci8: implement session_mode parameter

### LINKED ISSUE #801: [oci8] Oracle driver should support the session mode argument
Hello,
I need it to support external authentication OCI_CRED_EXT

dregad:
Fixes #801

### DIFF
--- a/docs/changelog.md
+++ b/docs/changelog.md
@@ -16,6 +16,12 @@ Older changelogs:
 
 ## [5.23.0] - Unreleased
 
+### Added
+
+- oci8: support session_mode parameter
+  [#801](https://github.com/ADOdb/ADOdb/issues/801)
+
+
 ## [5.22.1] - Unreleased
 
 ### Removed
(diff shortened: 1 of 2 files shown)


dregad:
Please note that the code is untested, as I do not have an Oracle setup available at the moment.

dregad:
@eminemma your feedback on this would be appreciated. Did you get a chance to test the feature, is everything working as expected ? Thanks in advance.

eminemma:
I tried it but it didn't work as expected
I change the line 659 from adodb.inc.php
Actual
$this->connectionParameters[] = array($parameter=>$value);
Fix
$this->connectionParameters[$parameter] = $value;

dregad:
Thanks for testing @eminemma 

You're right, it does not work but your proposed fix is not correct as it would break other drivers relying on _connectionParameters_ property to be a set of parameter/value pairs.

I'll push a fix shortly.

eminemma:
I tried it with the fix and it works correctly, sorry for my English I don't speak that much and it's hard for me to interpret, I use a lot of google to understand

dregad:
Thanks for the feedback. Don't worry about your English, so far it's been (mostly) good enough to understand. What's your native language ?

eminemma:
my native language is spanish, i'm from Latin America

dregad:
Ah in that case I can't help much, my Spanish is just about good enough to order a :beer:

Issue #814, opened 2022-03-20
mysqli remove legacy $optionFlags property


dregad:
mysqli $optionFlags property is ignored by _connect(). It is reset to an empty array, so the subsequent foreach loop never does anything.

https://github.com/ADOdb/ADOdb/blob/d99127b8811a9f949c00ab425065e380dce6bdc1/drivers/adodb-mysqli.inc.php#L182-L186

This change was introduced in 5.21.0-beta.1, commit 0e3d95cc2c38906a39172fcbdf1f4f28c8c955bd, as a fix for #166, but it seems unrelated to the issue being patched. 

@mnewnham can you please have a look ?

mnewnham:
Will do today

mnewnham:
It would appear that the ``optionFlags`` array is a MySQL only option that has been replaced by ``setConnectionParameter()``. Given that it has been non-functional for a substantial period of time, I would suggest scraping all reference to it out of the driver.

dregad:
Thanks for looking into it. I'm fine with your recommended approach.

PR #815, opened 2022-03-20
[mysqli] Remove mysqli driver  legacy option flags parameter, see #814

mnewnham:
Obsolete feature replaced by use of standard setConnectionParameter() methodology


dregad:
Thanks for the PR. I had actually already fixed that locally yesterday, but forgot to push - sorry for the miscommunication and wasting your time.

Issue #807, opened 2022-03-07
memcached weighted servers not working (never connects)


setcooki:
the example in https://adodb.org/dokuwiki/doku.php?id=v5:userguide:memcached does not work because of this line =

https://github.com/ADOdb/ADOdb/blob/ded096e9d97a0c06e82b062cfa432c55152293fc/adodb-memcache.lib.inc.php#L189

this will produce: 

```
    [serverControllers:ADODB_Cache_MemCache:private] => Array
        (
            [0] => Array
                (
                    [host] => Array
                        (
                            [host] => 192.168.86.91
                            [port] => 11211
                            [weight] => 70
                        )

                    [port] => 11211
                    [weight] => 0
                    [key] => 
                )

            [1] => Array
                (
                    [host] => Array
                        (
                            [host] => 192.168.86.92
                            [port] => 11211
                            [weight] => 30
                        )

                    [port] => 11211
                    [weight] => 0
                    [key] => 
                )

        )
```

Also this line > https://github.com/ADOdb/ADOdb/blob/ded096e9d97a0c06e82b062cfa432c55152293fc/adodb-memcache.lib.inc.php#L234 makes absolutely no sense since Memcached::getServerList() always returns an array and if no server is present it returns an empty array

dregad:
@mnewnham I actually started working on this yesterday, but got sidetracked and could not finish as planned. I just finalized the fixes and submitted PR #816.

Since you assigned yourself to this, maybe you'd like to review and comment.

mnewnham:
couple of items of note:

1. The code associated with the 'key' parameter is completely missing That portion of the code is associated with the new development path I was working on with the replacement caching and session modules and there has clearly been some mistake on my part that has allowed this to remain in the code base. I would recommend we pull the code associated with this feature from the current release. 
2. I don't quite understand what this is:
````php
			/** @noinspection PhpExpressionResultUnusedInspection */
```` 
Is this some tool you are using that is injecting this into the code? What does it mean?

dregad:
See my answer in https://github.com/ADOdb/ADOdb/pull/816#issuecomment-1074935563

PR #816, opened 2022-03-21
Memcache bug fixes and improvements

### LINKED ISSUE #823: memcache: memCacheCompress option ignored by memcached
When using memcached and the $memCacheCompress option is specified, ADODB_Cache_MemCache::connect() adds it as a connection option.

https://github.com/ADOdb/ADOdb/blob/d99127b8811a9f949c00ab425065e380dce6bdc1/adodb-memcache.lib.inc.php#L160-L169

This does not work correctly, for 2 reasons
1. using `==` instead of `=`, so the option is never actually assigned and Memcached always uses the default value for OPT_COMPRESSION (true)
2. the option is only assigned when `$this->compress` is true, so it's effectively not possible to disable caching.

### LINKED ISSUE #824: memcache: default port should be used in server controller template
In 5.22.0, it became possible to define memcache servers as array, specifying host, port and weight.

The ADODB_Cache_MemCache class defines a *$serverControllerTemplate* which hardcodes the default port to 11211. 

If a different value is defined globally in ADOConnection::$memCachePort it will be ignored.

dregad:
- Fixes the 2 issues in #807
- Fixes #823
- Fixes #824
- Improve PHPDoc
- Remove references to future _key_ option
- Code cleanup

### DIFF
--- a/adodb-memcache.lib.inc.php
+++ b/adodb-memcache.lib.inc.php
@@ -84,7 +84,6 @@ class ADODB_Cache_MemCache
 		'host' => '',
 		'port' => 11211,
 		'weight' => 0,
-		'key' => ''
 	);
 
 	/**
@@ -194,8 +193,7 @@ class ADODB_Cache_MemCache
 				if ($this->isLibMemcached()) {
 					$connector['weight'] = $connector['weight'] ? (int)$connector['weight'] : 0;
 				} else {
-					// Cannot use a key or weight in memcache, simply discard
-					$connector['key'] = '';
+					// Cannot use weight in memcache, simply discard
 					$connector['weight'] = 0;
 				}
 
--- a/adodb-memcache.lib.inc.php
+++ b/adodb-memcache.lib.inc.php
@@ -160,10 +160,8 @@ class ADODB_Cache_MemCache
 		}
 
 		// Convert simple compression flag for memcached
-		if ($this->isLibMemcached() && $this->compress) {
-			// Value of Memcached::OPT_COMPRESSION = 2;
-			/** @noinspection PhpExpressionResultUnusedInspection */
-			$this->options[2] == 1;
+		if ($this->isLibMemcached()) {
+			$this->options[Memcached::OPT_COMPRESSION] = $this->compress;
 		}
 
 		// Are there any options available for memcached
--- a/adodb-memcache.lib.inc.php
+++ b/adodb-memcache.lib.inc.php
@@ -17,6 +17,8 @@
  *
  * @copyright 2000-2013 John Lim
  * @copyright 2014 Damien Regad, Mark Newnham and the ADOdb community
+ *
+ * @noinspection PhpUnused
  */
 
 // security - hide paths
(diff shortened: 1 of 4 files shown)


dregad:
@mnewnham in reply to your note https://github.com/ADOdb/ADOdb/issues/807#issuecomment-1074640604

> 1. The code associated with the 'key' parameter is completely missing That portion of the code is associated with the new development path I was working on with the replacement caching and session modules and there has clearly been some mistake on my part that has allowed this to remain in the code base. I would recommend we pull the code associated with this feature from the current release. 

I'm sorry, what you want to do is not very clear to me. 

The PR is based on v5.22.0, which *is* the current release, and I don't see any code related to a _key_ parameter in there so I'm not sure needs to be pulled.

Do you mean the _key_ element in the $serverControllerTemplate property ? That one actually does not make much sense to me, as Memcached::addserver() does not expect it (only host, port and weight).

Please explain, or if it's easier for you, just commit whatever adjustments are needed to the *dregad:i807-memcached* branch.


> 2. I don't quite understand what this is:
>````php
>			/** @noinspection PhpExpressionResultUnusedInspection */
>```` 
>Is this some tool you are using that is injecting this into the code? What does it mean?

Yes indeed, this tells the PHPStorm IDE, which performs static analysis on the code, to ignore specific inspections.

But now that you mention it, I think squelched that one when I probably should not have, as I just noticed that the code does a comparison `==` instead of an assignment `=`.

https://github.com/ADOdb/ADOdb/blob/2fbb30196ca893ed9b1536b9f3f794bddfee8adc/adodb-memcache.lib.inc.php#L166

mnewnham:
###  Server Key Code
I'll expand out on my comment about the code missing: 

In order for the 'key' setting to function, the memcacached driver needs to call the ``getByKey()`` and ``setByKey()``.  I added these into my preview feature in the **master-future-cache** branch. in the  ``addins/cache/plugins/memcached/ADOCacheMethods.php`` program. To use this also needs changes to methods like ``cacheExecute()``. 

Until we can get all this in place, having the key option in place is misleading which is why I suggest removing the references.
### PHPStorm
Re your phpstorm comments, no harm will be caused if I change the code around them in the normal course of program changes?
### Code Review
I did review that uneeded bit of code, you are correct in that it could be removed, so if its easiest you could just chop it out

dregad:
> I did review that uneeded bit of code, you are correct in that it could be removed, so if its easiest you could just chop it out

OK, I removed the code in 88aa5d0f75d9888b6b6ce60f75cfdeb35427b136

> Re your phpstorm comments, no harm will be caused if I change the code around them in the normal course of program changes?

If you delete or change the statement immediately following a `@noinspection` hint, the best and safest would be to also delete the comment block

That being said, they are just comments, so the worst that can happen in such case, is that a static analysis warning may be unduly suppressed, thus potentially preventing early catching of a bug. No big deal. 

> Server Key Code

I'll rebase the branch to fix merge conflict before reviewing and responding to that.

mnewnham (approved):
I've checked these against both the memcache and memcached drivers and they look fine.

dregad:
Awesome, many thanks for testing !

Issue #820, opened 2022-03-27
Affected_Rows() not returning correct value


beastybeast:
Affected_Rows() is always returning false in my case.

Looks like the same problem as here: https://github.com/ADOdb/ADOdb/issues/800

I upgraded from 

$ADODB_vers = 'v5.20.14  06-Jan-2019';

to

$ADODB_vers = 'v5.22.0  2022-02-08';

And the following code fails 

```
$query = "INSERT INTO `sessions` (`id`, `tag`, `value`, `data`) VALUES (?, 'srpvalues', 'data', ?) ON DUPLICATE KEY UPDATE data = ?";
  $result = $db->Execute($query, array($data->email, $srpvalue, $srpvalue))
  or die(makereply($id, $pcrypt_errorstringarray[4].substr($query, 0, 200)." => ".$db->ErrorMsg(), 4));
/*
  trigger_error("DB: ".print_r($db, true));
  trigger_error("RESULT: ".print_r($result, true));
  trigger_error("AFFECTED: ".print_r($db->Affected_Rows(), true));
  trigger_error("RECORD: ".print_r($result->RecordCount(), true));  
*/
  if($db->Affected_Rows() == 0)
    die(makereply($id, $pcrypt_errorstringarray[4].substr($query, 0, 200)." => ".$db->ErrorMsg(), 4));
```
 if($db->Affected_Rows() == 0) is always true.

If I revert back to the old version everything works fine again?

NB: mysqli used

beastybeast:
Looks as if the error is still there in version 5.22.1 ?

Just downloaded latest stable release from https://adodb.org to test is out

dregad:
@beastybeast I just tried to reproduce this with 5.22.1, as follows

```php
$db = NewADOConnection('mysqli');
$db->connect($host, $user, $password, $database);
$sql = "insert into test (id, data) values (?, ?) on duplicate key update data = ?";
$r = $db->execute($sql, [1, "data",	"update" . time()]);
var_dump($db->affected_rows());
```

Displays `int(2)`, as expected (same results as with 5.20.14). Am I missing something ? Can you provide more information ?

beastybeast:
I am using PHP 8.1 on a Centos 7 platform (Koozali)

This is how I open the DB connection

$db = ADONewConnection(pcrypt::db_system);
[...]

PR #821, opened 2022-03-28
[mysqli] Ensure affected rows returns correct value see #820

mnewnham:
Back port fix from master

### DIFF
--- a/docs/changelog.md
+++ b/docs/changelog.md
@@ -25,8 +25,10 @@ Older changelogs:
 
 - mysqli: regression on setConnectionParameter()
   [#803](https://github.com/ADOdb/ADOdb/issues/803)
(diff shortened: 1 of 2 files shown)


dregad:
@mnewnham You did not pick up the following change from the original commit f79488b72345b53b6bb184bfc3985df7604769fe as part of the back-port, was that intentional ?

https://github.com/ADOdb/ADOdb/blob/f79488b72345b53b6bb184bfc3985df7604769fe/drivers/adodb-mysqli.inc.php#L1259-L1262

beastybeast:
Looks as if the error is still there in version 5.22.1 ?

Just downloaded latest stable release from https://adodb.org/ to test is out

PR #826, opened 2022-04-02
PDO drivers refactoring with preload, new drivers for firebird and DB2

dregad:
This PR's purpose is to facilitate review of @mnewnham's recent work on PDO drivers.

Includes updated mysql, sqlite, pgsql, sqlsrv and oci drivers, plus new drivers for firebird and ibm db2.

The only change required for legacy code is to update the *'ADONewConnection()* call, as follows:

before
```php
$db = ADONewConnection('PDO');
$db->connect('mysql:<parameters>');
```
after
```php
$db = ADONewConnection('PDO\mysql');
```

All of the driver files are loaded as necessary by ADOLoadCode(). This allows the class to load properly i.e. 'ADORecordset_pdo_mysql -> ADORecordset_pdo -> ADORecordset', rather than the current method, where 'ADORecordset' loads 'ADORecordset_pdo_mysql'


mnewnham:
See more about this new driver [At the new documentation page](https://adodb.org/dokuwiki/doku.php?id=v5:database:pdo_new)

drivers/adodb-pdo.inc.php
```diff
@@ -64,79 +64,97 @@ function adodb_pdo_type($t)
 
 /*----------------------------------------------------------------------------*/
 
-
-class ADODB_pdo extends ADOConnection {
+class ADODB_pdo extends ADOConnection
```
dregad, line 67: Shouldn't this be an *abstract* class now ? If I'm not mistaken after the refactoring it's only ever used as a base class for the PDO drivers.

As a nice side-effect, this would ensure legacy `ADONewConnection('pdo')` calls are no longer possible, as a _PHP Fatal error:  Uncaught Error: Cannot instantiate abstract class ADODB_pdo_ exception would be thrown in this case.

adodb.inc.php
```diff
@@ -5585,7 +5631,15 @@ function ADONewConnection($db='') {
...
-			$cls = 'ADODB_'.$db;
+			if ($pdoExtension)
+			{
+				$cls = 'ADODB_pdo_' . $pdoExtension;
+				$db = 'PDO';
+			}
+			else
+			{
+				$cls = 'ADODB_'.$db;
+			}
```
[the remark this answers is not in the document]
dregad: Also, what about performing early detection of legacy ADONewConnection('pdo') calls (i.e. when `$db == 'pdo' && $pdoExtension == ''`), either here or maybe even better when we split and process the pdo driver string above, and throw an error instead of delaying this until ADODB_pdo::_connect()

drivers/adodb-pdo.inc.php
```diff
@@ -64,79 +64,97 @@ function adodb_pdo_type($t)
...
+		)
+		* So its easy to determine that the driver we are using is mysql
+		*/
+		$driverClass      = array_pop($driverClassArray);
+	
+		$at = strpos($argDSN,':');
+		if ($at > 0)
+		{
+		
+			$this->dsnType = substr($argDSN,0,$at);
```
dregad, line 142: Probably better to use explode() - more efficient than strpos+substr, and avoids error in case DSN begins with a `:`
```php
$dsnSplit = explode(':', $argDSN, 2);
if (count($dsnSplit) > 1) {
    $this->dsnType;
```

adodb.inc.php
```diff
@@ -5363,7 +5373,8 @@ function ADOLoadDB($dbType) {
 	/**
 	 * Load the code for a specific database driver. Private function. Do not use.
 	 */
-	function ADOLoadCode($dbType) {
+	function ADOLoadCode($dbType,$pdoExtension=false) {
```
dregad, line 5376: Why add a $pdoExtension parameter ? Wouldn't it be simpler to just build and pass the appropriate `pdo_XXX` string from ADONewConnection() ?

drivers/adodb-pdo_ibm.inc.php
```diff
@@ -0,0 +1,609 @@
...
+ * and the GNU Lesser General Public Licence (LGPL) v2.1 or, at your option,
+ * any later version. This means you can use it in proprietary products.
+ * See the LICENSE.md file distributed with this source code for details.
+ * @license BSD-3-Clause
+ * @license LGPL-2.1-or-later
+ *
+ * @copyright 2022 Damien Regad, Mark Newnham and the ADOdb community
+ */
+
+final class ADODB_pdo_ibm extends ADODB_pdo {
```
dregad, line 21: Why `final` ?

drivers/adodb-pdo_informix.inc.php
```diff
@@ -0,0 +1,597 @@
...
+ * and the GNU Lesser General Public Licence (LGPL) v2.1 or, at your option,
+ * any later version. This means you can use it in proprietary products.
+ * See the LICENSE.md file distributed with this source code for details.
+ * @license BSD-3-Clause
+ * @license LGPL-2.1-or-later
+ *
+ * @copyright 2022 Damien Regad, Mark Newnham and the ADOdb community
+ */
+
+final class ADODB_pdo_informix extends ADODB_pdo {
```
dregad, line 22: Why `final` ?

drivers/adodb-pdo_mysql.inc.php
```diff
@@ -19,7 +19,7 @@
  * @copyright 2014 Damien Regad, Mark Newnham and the ADOdb community
  */
 
-class ADODB_pdo_mysql extends ADODB_pdo {
+final class ADODB_pdo_mysql extends ADODB_pdo {
```
dregad, line 22: Why `final` ?

drivers/adodb-pdo_oci.inc.php
```diff
@@ -19,7 +19,7 @@
  * @copyright 2014 Damien Regad, Mark Newnham and the ADOdb community
  */
 
-class ADODB_pdo_oci extends ADODB_pdo_base {
+final class ADODB_pdo_oci extends ADODB_pdo {
```
dregad, line 22: Why `final` ?

drivers/adodb-pdo_pgsql.inc.php
```diff
@@ -19,7 +19,7 @@
  * @copyright 2014 Damien Regad, Mark Newnham and the ADOdb community
  */
 
-class ADODB_pdo_pgsql extends ADODB_pdo {
+final class ADODB_pdo_pgsql extends ADODB_pdo {
```
dregad, line 22: Why `final` ?

drivers/adodb-pdo_sqlite.inc.php
```diff
@@ -15,41 +15,37 @@
...
  *
- * @copyright 2000-2013 John Lim
- * @copyright 2014 Damien Regad, Mark Newnham and the ADOdb community
+ * @copyright 2022 Damien Regad, Mark Newnham and the ADOdb community
  * @author Diogo Toscano <diogo@scriptcase.net>
  * @author Sid Dunayer <sdunayer@interserv.com>
  */
 
-class ADODB_pdo_sqlite extends ADODB_pdo {
+final class ADODB_pdo_sqlite extends ADODB_pdo {
```
dregad, line 23: Why `final` ?

drivers/adodb-pdo_sqlsrv.inc.php
```diff
@@ -20,48 +20,522 @@
  * @author Ned Andre
  */
 
-class ADODB_pdo_sqlsrv extends ADODB_pdo
+final class ADODB_pdo_sqlsrv extends ADODB_pdo
```
dregad, line 23: Why `final` ?

adodb.inc.php
```diff
@@ -5444,6 +5468,28 @@ function ADONewConnection($db='') {
 			define('ADODB_ASSOC_CASE', ADODB_ASSOC_CASE_NATIVE);
 		}
 
+		/*
+		* If we have a pdo and drivern in format \\, we
+		* need to remove the driver portion of the file
+		*/
+		$pdoExtension = '';
+		$pdoSplit 	  = explode('\\',$db);
```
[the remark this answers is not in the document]
mnewnham: I think that as it stands, provision of the DSN on the newADOcommand line is no longer supported. Not to say that the processing of the parameter should be super robust.  I did note it in the Quick upgrade notes.

adodb.inc.php
```diff
@@ -5444,6 +5468,28 @@ function ADONewConnection($db='') {
...
+		
+		if (count($pdoSplit) > 1)
```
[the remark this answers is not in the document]
mnewnham: Maybe it would be better if I split the entirety of the statement

````php
$split = preg_split('/^(pdo\\)/',$driver);

File: drivers/adodb-pdo_mysql.inc.php (hunk 7 above)
Comment (mnewnham) [reply drivers/adodb-pdo_mysql.inc.php:22]: Thats a very open question - we use them a lot because we've thousands of classes in our application in its a just a useful way to track the usage so it's become a bit of a habit.  

Not sure why we wouldn't start setting these.  The way that we use classes is not very intuitive ( and not  well documented)

Comment (dregad) [reply drivers/adodb-pdo_mysql.inc.php:22]: Personally I see no added value in making our classes final... In fact, we even advertise customizing ADOdb by extending drivers [in the documentation](https://adodb.org/dokuwiki/doku.php?id=v5:userguide:learn_extensions:new_driver)... Why should we prevent someone from doing this with the PDO drivers ?

File: adodb.inc.php (hunk 12 above)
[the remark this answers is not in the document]
Comment (dregad) [reply adodb.inc.php:5476]: > provision of the DSN on the newADOcommand line is no longer supported 
> I did note it in the Quick upgrade notes.

I missed that bit.
Is there a particular reason for dropping this functionality ? It could be fairly easily maintained by adapting the PDO-specific logic to use the $pdoExtension variable:
```diff
diff --git a/adodb.inc.php b/adodb.inc.php
index c5e30f09..6ff88c0b 100644
--- a/adodb.inc.php
+++ b/adodb.inc.php
@@ -5547,17 +5547,11 @@ class ADORecordSet implements IteratorAggregate {
                $dsna['host'] = '';
            }
 
-           if (strncmp($origdsn,'pdo',3) == 0) {
-               $sch = explode('_',$dsna['scheme']);
-               if (sizeof($sch)>1) {
-                   $dsna['host'] = isset($dsna['host']) ? rawurldecode($dsna['host']) : '';
-                   if ($sch[1] == 'sqlite') {
-                       $dsna['host'] = rawurlencode($sch[1].':'.rawurldecode($dsna['host']));
-                   } else {
-                       $dsna['host'] = rawurlencode($sch[1].':host='.rawurldecode($dsna['host']));
-                   }
-                   $dsna['scheme'] = 'pdo';
-               }
+           if ($pdoExtension) {
+               $host = $pdoExtension
+                   . ($pdoExtension == 'sqlite' ? ':' : ':host=')
+                   . rawurldecode($dsna['host'] ?? '');
+               $dsna['host'] = rawurlencode($host);
            }
 
            $db = @$dsna['scheme'];
```

adodb.inc.php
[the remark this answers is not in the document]
dregad: That would work better indeed.

I would suggest adjusting the regex: `/^pdo[_\\]/` 
- removing the unnecessary capturing group, 
- adding `_` as accepted delimiter for backwards-compatibility, following my earlier suggestion (https://github.com/ADOdb/ADOdb/pull/826#discussion_r841080490)

adodb.inc.php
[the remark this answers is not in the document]
mnewnham: Two things on future issues, 

1. The ''_'' character can't be used if using off-the-shelf PSR style autoloaders, we'll probably need a project in the future to remove them from classnames. Might as well start somewhere.
2.   To be honest, I don't particularly like fact that you can assign a DSN through here (and not any other type of parameter). It feels very unstructured, and I'd certainly not do it if I did the original design.

mnewnham:
@dregad  I have brought this repo up to date, could you take a look again please

faisalliaqat1 (commented):
just test coments functonility

Issue #829, opened 2022-04-14
mssql: Automatic conversion of false to array is deprecated in PHP 8.1


fluchette:
Hi,

I get the following message with the mssqlnative driver in PHP 8.1 :

PHP Deprecated: Automatic conversion of false to array is deprecated adodb-mssqlnative.inc.php on line 154

Thanks

dregad:
I just found out that the same issue had been [reported on StackOverflow](https://stackoverflow.com/questions/71035322/php-deprecated-automatic-conversion-of-false-to-array-is-deprecated-adodb-mssql) a couple months ago.

PR #830, opened 2022-04-15
Fix undeclared variable, see #829

mnewnham:


dregad:
@mnewnham Looking at the ServerInfo() method, I'm wondering if we really need to cache the server information (is sqlsrv_server_info() an expensive call ?) If not, I would suggest simplifying the code by removing the static variable.

If the cache needs to be kept, then simply removing the assignment to _false_ when declaring the variable and adjusting the test to compare against _null_ in the subsequent statement would be a better fix IMO.
```php
static $arr;
if ($arr !== null)
```

Alternatively, we could also initialize the array with a single statement
```php
$arr = array(
	'description' => $arrServerInfo['SQLServerName'] . ' connected to ' . $arrServerInfo['CurrentDatabase'],
	'version' => $arrServerInfo['SQLServerVersion'],
);
```

Also, you referred to the wrong issue in the commit and PR description, it should have been #829

dregad:
One more thing, I do not understand why we're saving, setting and restoring the fetch mode, this seems useless... Any idea ?

dregad:
Code could be simplified as follows:
```php
function serverInfo() {
	$info = sqlsrv_server_info($this->_connectionID);
	return array(
		'description' => $info['SQLServerName'] . ' connected to ' . $info['CurrentDatabase'],
		'version' => $info['SQLServerVersion'],
	);
}
```

mnewnham:
1. I think it was static because we were using it a lot more to differentiate between SQL server versions in sequence management where versions prior to server 2012 had different management. This isn't really important anymore. Probably defaulting to server 2012+ for sequence management but have prior version available would be better, we could also remove static in the function
2. As we insist on fetch mode being ASSOC for this to work correctly would suggest that saving off any prior mode and restoring afterwards would be appropriate unless I'm misunderstanding your meaning.
3. I spotted that incorrect link, but I've no idea how to fix it.

dregad:
> As we insist on fetch mode being ASSOC for this to work correctly would suggest that saving off any prior mode and restoring afterwards would be appropriate unless I'm misunderstanding your meaning.

It makes no sense to do this, as sqlsrv_server_info() does not know about ADOConnection::$fetchMode... 

Having looked at the Git history for the ServerInfo() method, it looks like this is a [leftover from 2007 code](https://github.com/ADOdb/ADOdb/blob/748a67d70f6e2d924a0fcccac9c4b405172e9efc/phplens/adodb/drivers/adodb-mssql.inc.php#L123), when the server information was retrieved by calling the [sp_server_info](https://docs.microsoft.com/en-us/sql/relational-databases/system-stored-procedures/sp-server-info-transact-sql) procedure.

To minimize the risk of regression, I'll merge your PR in the _hotfix/5.22_ branch, and apply the simplified code from https://github.com/ADOdb/ADOdb/pull/830#issuecomment-1100602929 in _master_.

PR #834, opened 2022-05-06
Fix #833 -- PostgreSQL: On PHP 8.1, Affected_Rows() always returns false.

### LINKED ISSUE #833: PostgreSQL: On PHP 8.1, Affected_Rows() always returns false.
On PHP 8.1, Affected_Rows() always returns false when using PostgreSQL.

```
	function _affectedrows()
	{
		if (!is_resource($this->_resultid) || get_resource_type($this->_resultid) !== 'pgsql result') return false;
		return pg_affected_rows($this->_resultid);
	}
```

I'm pretty sure this is because of that get_resource_type check -- On PHP 8.1.0 and above, the 'pgsql result' resource has been replaced by the PgSql\Result class:

https://www.php.net/manual/en/class.pgsql-result.php

I also see a bunch of other similar checks throughout adodb-postgres64.inc.php, which will also all need changing.

ReimuHakurei:
Fixes #833 by replacing the previous resource type checks with a simple check for false.

On all versions of PHP, any error that would've made the types of these values not be the resource (pre-8.1) or object (8.1+) would result in the result being false, so this should accomplish the same thing in a way that works on all PHP versions.

### DIFF
--- a/drivers/adodb-postgres64.inc.php
+++ b/drivers/adodb-postgres64.inc.php
@@ -155,7 +155,7 @@ class ADODB_postgres64 extends ADOConnection{
 	 */
 	protected function _insertID($table = '', $column = '')
 	{
-		if (!is_resource($this->_resultid) || get_resource_type($this->_resultid) !== 'pgsql result') return false;
+		if ($this->_resultid === false) return false;
 		$oid = pg_last_oid($this->_resultid);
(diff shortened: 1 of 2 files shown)


dregad (approved):
Looks good. 

You just missed one occurrence in adodb-postgres7.inc.php, I'll fix that.

dregad:
Thanks @ReimuHakurei for your contribution !

Issue #842, opened 2022-06-21
MySQL 8 creates connection object even if client connection fails


mnewnham:
If a connection to a MySQL version 8 fails, a mysqli connection object similar to below is created.
````php
mysqli Object
(
    [client_info] => mysqlnd 8.0.18
    [client_version] => 80018
    [connect_errno] => 1045
    [connect_error] => Access denied for user 'some_user'@'somehost.com' (using password: YES)
    [errno] => 1045
    [error] => Access denied for user 'some_user'@'somehost.com' (using password: YES)
)
````
This causes an error in **ADOConnection_mysqli::_errorMsg()** which expects the connection object to be empty if there is no connection.

PR #843, opened 2022-06-21
[mysql] MySQL 8 returns object if client connection fails see #842

mnewnham:
Earlier versions left the connection object empty if the connection failed

### DIFF
--- a/drivers/adodb-mysqli.inc.php
+++ b/drivers/adodb-mysqli.inc.php
@@ -1392,10 +1392,12 @@ class ADODB_mysqli extends ADOConnection {
 	 */
 	function ErrorNo()
 	{
-		if (empty($this->_connectionID))
-			return @mysqli_connect_errno();
-		else
-			return @mysqli_errno($this->_connectionID);
+		if (empty($this->_connectionID)) {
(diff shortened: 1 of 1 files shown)


dregad (changes requested):
There's one case that is not covered by the proposed code: if the connection has been closed (with mysqli_close()), ErrorMsg() will throw *PHP warning:  Uncaught Error: mysqli object is already closed* when mysqli_error() is called in the else block. 
EDIT: with PHP < 8.0 the warning won't actually be thrown due to the `@` operator, but the function will return *false*; with PHP 8.0 or later, this is a fatal error (exception).

Considering that _connectionID is either a mysqli object or false/null, the is_object() check is unnecessary. I would just replace mysqli_error()  instead:

```php
if (empty($this->_connectionID)) {
	$this->_errorMsg = mysqli_connect_error();
} else {
	$this->_errorMsg = $this->_connectionID->error ?? $this->_connectionID->connect_error;
}
return $this->_errorMsg;
```

The same fix should be applied to the ErrorNo() method as well.

Note: I believe the `@` operator is unnecessary on mysqli_connect_error() functions, I suggest we remove it.

PR #855, opened 2022-08-21
PHP 8.1 - replace strftime() with date() - sqlite 

fmancardi:
Because strftime() has been deprecated in PHP 8.1 and these functions will be removed in PHP 9.0, I've replaced it with date().

In adodb-time.inc.php  gmstrftime() is used but I was no able to find the way to fix it.

### DIFF
--- a/drivers/adodb-sqlite.inc.php
+++ b/drivers/adodb-sqlite.inc.php
@@ -388,7 +388,7 @@ class ADODB_sqlite extends ADOConnection {
 	 */
 	function month($fld)
 	{
-		$x = "strftime('%m',$fld)";
+		$x = "date('m',$fld)";
 
 		return $x;
 	}
@@ -400,7 +400,7 @@ class ADODB_sqlite extends ADOConnection {
 	 * @return	str				The SQL Statement
 	 */
 	function day($fld) {
-		$x = "strftime('%d',$fld)";
+		$x = "date('d',$fld)";
 		return $x;
 	}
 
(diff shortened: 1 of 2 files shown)


dregad:
@fmancardi thanks for your contribution, but unless I'm missing something here, I believe this change must not be applied.

The ADOConnection's day()/month()/year() functions are used to simplify the creation of an SQL statement to obtain just the formatted day/month/year of a specified date/time field, see https://adodb.org/dokuwiki/doku.php?id=v5:reference:connection:month.

In the SQLite drivers, they are referencing [SQLite's internal strftime() function](https://www.sqlite.org/lang_datefunc.html), **not PHP's**. To my knowledge, there is no change or deprecation planned on SQLite for this.

> In adodb-time.inc.php gmstrftime() is used but I was no able to find the way to fix it.

I don't actually use these routines myself, but AFAICT adodb_strftime() falls back to using adodb_date() after converting the format string when the timestamp is greater than 0x7FFFFFFF, so we could probably just remove that block (and maybe deprecate adodb_strftime() and adodb_gmstrftime() too)

https://github.com/ADOdb/ADOdb/blob/d7005067be28ccbcd15c1541c29e32249da26371/adodb-time.inc.php#L1370-L1374

Come to think of it, this 0x7FFFFFFF limit does not make much sense in the days of 64-bit integers...  And the whole adodb-time library is probably obsolete too.

mnewnham:
Hey @fmancardi , don't worry about it. We very much appreciate contributions, thank you for your effort and hope that you're interested enough in the project to help us again. If you've any ideas, you should raise an issue before you do your coding.

Issue #865, opened 2022-09-05
datadict: different results when column is modified between alterColumnSql() and changeTableSQL()


mnewnham:
If the same simple change to the schema of a table is applied to a column using the above 2 methods, the output is completely different. This is because ''changeTableSql()'' does not call the database specific method, but instead uses the generic base method. So using the following change to a table:
````php
'regiondescription C(100) NOTNULL DEFAULT "BILL"';

````
When passed to ''alterColumnSql()'', the result is
````php
Array
(
    [0] => ALTER TABLE region ALTER COLUMN regiondescription VARCHAR(100) NOT NULL
    [1] => ALTER TABLE region ADD CONSTRAINT DF__region__regiondescription__51a89880 DEFAULT 'BILL' FOR regiondescription
)
````
When passed to ''changeTableSql()''
````php
Array
(
    [0] => ALTER TABLE region ALTER COLUMN regiondescription VARCHAR(100) DEFAULT 'BILL' NOT NULL
)
````
In addition, ''changeTableSql()'' now calls the driver specific ''dropColumnSql()'' instead of using the generic method in case of column drops.

This issue was previously reported by @boshrin on  #383 For Postgres/xmlSchema but covers all drivers using the datadict feature, thx for the code snippet.

PR #866, opened 2022-09-05
datadict: Call driver specific column change methods in alterTableSql()

mnewnham:
The alterTableSql uses core methods instead of driver specific ones, see #865 , #383. Change method to create create change SQL statements.


dregad (approved):
I'll target this at hotfix/5.22

tuxedo-psc:
Hi, 

this change removes creating new fields as stated in the comment. 

Is this intentional ?

PR #871, opened 2022-09-08
Refactoring adodb_strip_order_by(), add test case

### LINKED ISSUE #869: adodb_strip_order_by() throws deprecated warnings on PHP 8.1
```
Deprecated:  strpos(): Passing null to parameter #1 ($haystack) of type string is deprecated in ./adodb/adodb-lib.inc.php on line 39
Deprecated:  str_replace(): Passing null to parameter #1 ($search) of type array|string is deprecated in ./adodb/adodb-lib.inc.php on line 55
```

To reproduce, call pageExecute with a SELECT statement without an ORDER BY clause:
```php
$db->pageExecute('SELECT * FROM table');
```

### LINKED ISSUE #870: adodb_strip_order_by() shouldn't strip clause from subqueries
The fix for #549 is working fine for queries with sub-queries where the outer SELECT has an ORDER BY clause.

However it does not produce the expected result when the given SQL contains a sub-query with an ORDER BY clause and the outer SELECT does not, e.g.
```
SELECT * FROM table WHERE name IN (SELECT TOP 5 name FROM table_b ORDER by name)

Expected :'SELECT * FROM table WHERE name IN (SELECT TOP 5 name FROM table_b ORDER by name)'
Actual   :'SELECT * FROM table WHERE name IN (SELECT TOP 5 name FROM table_b)'
```

dregad:
- Fixes #869 
- Fixes #870

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -16,3 +16,7 @@
 phpdoc.xml
 docs/api
 docs/cache
+
+# PHPUnit
+/.phpunit.cache
+.phpunit.result.cache
--- a/adodb-lib.inc.php
+++ b/adodb-lib.inc.php
@@ -27,34 +27,22 @@ if (!defined('ADODB_DIR')) die();
 global $ADODB_INCLUDED_LIB;
 $ADODB_INCLUDED_LIB = 1;
 
+/**
+ * Strip the ORDER BY clause from the outer SELECT.
+ *
+ * @param string $sql
+ *
+ * @return string
+ */
 function adodb_strip_order_by($sql)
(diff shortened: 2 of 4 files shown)


dregad:
@mnewnham let me know if you can think of any additional, meaningful test cases for adodb_strip_order_by() or if I missed something in the refactoring.

mnewnham (approved):
I've tried to think of something more complex but I can't so I think we're good with this

dregad:
Thanks for the feedback.

PR #876, opened 2022-09-16
mysqli: store error msg/code after query execution

### LINKED ISSUE #872: mysqli: regression on errorMsg()/errorCode()
I see a difference in how errors are reported for duplicate entries in mysqli (error number 1062). Up to version (including) 5.21.4  all duplicate entries outputted an error (ErrorNo() and ErrorMsg() are correctly populated. For newer versions (5.22.*) duplicate entries are not outputted as errors anymore (ErrorNo() is 0 and ErrorMsg() is "").

dregad:
ADOConnection's errorNo() and errorMsg() methods were returning 0 / '' when executing an SQL statement with bound parameters, since the introduction of true prepared statements (in #655).

This is because they are getting error information from the mysqli connection object, but when using prepared statements, it must be retrieved from the mysqli_stmt object.

Fixes #872

### DIFF
--- a/drivers/adodb-mysqli.inc.php
+++ b/drivers/adodb-mysqli.inc.php
@@ -1311,6 +1311,10 @@ class ADODB_mysqli extends ADOConnection {
 
 				$ret = mysqli_stmt_execute($stmt);
 
+				// Store error code and message
+				$this->_errorCode = $stmt->errno;
+				$this->_errorMsg = $stmt->error;
+
 				/*
 				* Did we throw an error?
 				*/
@@ -1336,6 +1340,10 @@ class ADODB_mysqli extends ADOConnection {
 			*/
 			$this->usePreparedStatement   = false;
 			$this->useLastInsertStatement = false;
+
+			// Reset error code and message
+			$this->_errorCode = 0;
+			$this->_errorMsg = '';
 		}
 
 		/*
@@ -1377,10 +1385,12 @@ class ADODB_mysqli extends ADOConnection {
 	 */
 	function ErrorMsg()
 	{
-		if (empty($this->_connectionID)) {
-			$this->_errorMsg = mysqli_connect_error();
-		} else {
-			$this->_errorMsg = $this->_connectionID->error ?? $this->_connectionID->connect_error;
+		if (!$this->_errorMsg) {
+			if (empty($this->_connectionID)) {
+				$this->_errorMsg = mysqli_connect_error();
+			} else {
+				$this->_errorMsg = $this->_connectionID->error ?? $this->_connectionID->connect_error;
+			}
 		}
 		return $this->_errorMsg;
 	}
@@ -1392,10 +1402,12 @@ class ADODB_mysqli extends ADOConnection {
 	 */
 	function ErrorNo()
 	{
-		if (empty($this->_connectionID)) {
-			$this->_errorCode = mysqli_connect_errno();
-		} else {
-			$this->_errorCode = $this->_connectionID->errno ?? $this->_connectionID->connect_errno;
+		if (!$this->_errorCode) {
+			if (empty($this->_connectionID)) {
+				$this->_errorCode = mysqli_connect_errno();
+			} else {
+				$this->_errorCode = $this->_connectionID->errno ?? $this->_connectionID->connect_errno;
+			}
 		}
 		return $this->_errorCode;
 	}


dregad:
@mnewnham I'd like to pick your brains please. While working on this, I was puzzled by the _query() method's logic when the *$sql* parameter is an array.

First of all, I'm not really sure in which context this would actually happen; [documentation](https://adodb.org/dokuwiki/doku.php?id=v5:reference:connection:execute#parameter_1) mentions: _A array of SQL Statements. This is only available to internal functions._ Which internal functions ?

Second and more importantly, I am concerned that the code is faulty, but not knowing how it is/should be used I can't really test. Consider lines 1210 onwards:

https://github.com/ADOdb/ADOdb/blob/6f373af66d79468fae0f0f179dfa0c59a7c3cc58/drivers/adodb-mysqli.inc.php#L1210-L1215
https://github.com/ADOdb/ADOdb/blob/6f373af66d79468fae0f0f179dfa0c59a7c3cc58/drivers/adodb-mysqli.inc.php#L1229-L1231

If *$sql* is indeed an array of SQL statements (i.e. strings), then *$stmt* contains a string (with the 2nd statement, which is kind of strange), but mysqli_stmt_bind_param() and mysqli_stmt_execute() expect a *mysqli_stmt* object so PHP throws a warning
*Uncaught TypeError: mysqli_stmt_execute(): Argument #1 ($statement) must be of type mysqli_stmt, string given*

If *$sql[1]* is indeed a mysqli_stmt object, then the documentation is incorrect. Looking at ADODB_mysqli::Prepare(), there is some dead code that does `return array($sql,$stmt);` which would explain the *$sql[1]*, but going as far back as 2005 that has never been active and the method always returns the original SQL.

Could this be dead code that should be removed ? Or am I missing something ? Let me know your thoughts.

mnewnham:
I'll have a look at it when I get home later

mnewnham:
#### is_array($sql)
I remember this code when I was building the documentation and when adding the true bound statement and being puzzled myself about it.

 I **think** that someone had a go at building bind support when the mysqli driver first came out and ultimately abandoned it, by copying the oci8 driver (which **_can_** use $sql as an array for 2d bound statements) (There's an explanation in the docblock above _query() in oci8 to explain).  
We do, however handle multi-dimensional inserts in the new code by passing the multi-dimensional info in inputarr[0].

Therefore I think that the first bit of code is dead and could be safely removed

As to the whole "internal functions" thing, I think that is misleading, it's more connected with the Oracle stuff (which I dont really understand) . Looking at adodb-oci8.inc.php at around line 780, you can see that there is a whole bunch of 2d bind arrays _**or**_ it seems you can pass in multiple SQL statements separated by a ":" This is a rabbit hole I do not want to go down

dregad:
Many thanks for looking into it.

> Therefore I think that the first bit of code is dead and could be safely removed

So we reach the same conclusion :smiley: 
I'll open a separate issue to track removal of this dead code and target at 5.23

> As to the whole "internal functions" thing, I think that is misleading

Should we just remove that from the documentation ? 

> Looking at adodb-oci8.inc.php at around line 780, you can see that there is a whole bunch of 2d bind arrays or it seems you can pass in multiple SQL statements separated by a ":"

Is this the correct line reference ? That points me to ADODB_oci8::SelectLimit() method, and the only reference to `:` is when processing the query parameters :thinking:

mnewnham:
Sorry, must have been falling asleep or something. What I meant we this piece of code at line 1033

````php
if ($array2d && $this->_bindInputArray) {
    if (is_string($sql)) {
        $stmt = $this->Prepare($sql);
    } else {
        $stmt = $sql;
    }

    foreach($inputarr as $arr) {
        $ret = $this->_Execute($stmt,$arr);
        if (!$ret) {
            return $ret;
        }
    }
    return $ret;
} else {
    $sqlarr = explode(':', $sql); <------------------------- MULTIPLE SQL STATEMENTS SEP BY ":"
    $sql = '';
    $lastnomatch = -2;
````

dregad:
I thought that was what you meant, but was not sure... Thanks for confirming.

As I understand it, this block is about processing query parameters (with `:xxx` syntax in Oracle), not multiple SQL statements.

neurotools:
When we can see this issue closed and 5.22.4 tagged? as 5.22.3 is faulty, many of us are waiting for the next release.

Thanks a lot for your work!

PR #883, opened 2022-09-26
silence deprecation warnings on oci connect

jamadam:
This silences following warnings.

```
Deprecated: oci_pconnect(): Passing null to parameter #4 ($encoding) of type string is deprecated
```


dregad:
Thanks @jamadam for your contribution.

Note: this does not apply to the 5.22 branch, because ADODB_oci8::_connect() was refactored in master only (see commit 5a78dcc1ed18850fe19064ca38802352406f54b8).

PR #884, opened 2022-09-26
revert primary_key default to unset

jamadam:
Introduction of the default value broke oci8 behavior. This PR seems to fix it.

The default "False" avoids to meet following `isset` condition,

https://github.com/ADOdb/ADOdb/blob/e3e3d14051dcd4d9203a12de87848a43a0687da4/adodb-active-record.inc.php#L451

And end up with "No primary key found for table" error.

### DIFF
--- a/adodb.inc.php
+++ b/adodb.inc.php
@@ -247,7 +247,7 @@ if (!defined('_ADODB_LAYER')) {
 		/**
 		 * @var bool True if field is a primary key
 		 */
-		public $primary_key = false;
+		public $primary_key;
 
 		/**
 		 * @var bool True if field is unique key


mnewnham:
Thanks for your help on this. I agree that this is a problem, however I think that reverting  the ``ADOFieldObject{}`` class is probably the wrong approach, as that would ultimately weaken that definition, as well as putting that variable definition in conflict with all the others in that class. 

If you are able, I think it would be a better to refactor the active record code to make that match the ADOFieldObject definition.

dregad:
@jamadam are you planning to revise your pull request following the review ?

Issue #891, opened 2022-10-07
[session] PHP 7 does not permit non-static methods to be called statically


clarumedia:
https://github.com/ADOdb/ADOdb/blob/8dafad709ca8951b1b665f2576d4680f7b4ef9b1/session/adodb-session.php#L580

8192:Non-static method ADODB_Session::close() should not be called statically

How about changing all the methods in the class to be:

```public static function```

In my tests, even the private methods need to be called statically and publicly.....

dregad:
As mentioned in the [documentation](https://adodb.org/dokuwiki/doku.php?id=v5:session:session_index), _the original session management routines that use adodb-session.inc.php are deprecated as of ADOdb version 5.20, and will be removed in ADOdb Version 6.0.0._

@clarumedia You should be using adodb-session2.php. Does the problem also exist there ?

@mnewnham Is it really worth patching deprecated code ?

clarumedia:
Fair enough. I will move forwards. Thank you

dregad:
Thanks for your understanding.

PR #894, opened 2022-10-20
Ensure methods called statically have static definitions see #891

mnewnham:


dregad:
As mentioned in https://github.com/ADOdb/ADOdb/issues/891#issuecomment-1285102324 adodb-session.php is deprecated. Is it worth patching ?

PR #902, opened 2022-11-24
Update adodb.inc.php

PJFonseca:
This will avoid messages being sent to the output when debug is false, this is how I avoid seeing errors in PROD.

If this is a bad way and there is another option, please comment on it. 

I have changed my local files to work like this. :)


dregad:
Please refer to the [documentation](https://adodb.org/dokuwiki/doku.php?id=v5:reference:logging#overriding_the_default_logging_function), for instructions on how to properly override default output behavior, without changes to ADOdb code.

PR #921, opened 2023-01-17
Add Bug Report issue template

dregad:
This is leveraging Github's [Issue templates](https://docs.github.com/en/communities/using-templates-to-encourage-useful-issues-and-pull-requests/about-issue-and-pull-request-templates) feature. which should hopefully help improving the quality of the bug reports submitted by the community, as I find we frequently need to request the same additional information like ADOdb, PHP and DB type and version over and over again.

Let me know if you feel anything should be added or changed.

Note that we can also create a similar templates for *feature requests* and even a "custom" type, but we don't get that many of those so I'm not sure it's worth it. It's also possible for pull requests as well. Let me know if you think it any of these would be useful.

### DIFF
--- a/.github/ISSUE_TEMPLATE/bug_report.md
+++ b/.github/ISSUE_TEMPLATE/bug_report.md
@@ -2,7 +2,7 @@
 name: Bug report
 about: Submit a new bug report to help us improve ADOdb
 title: ''
-labels: ''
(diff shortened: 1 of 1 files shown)


dregad:
@mnewnham no comments on this ?

mnewnham (approved):

PR #924, opened 2023-01-22
Remove obsolete error reporting operator(@)

ioigoume:
Remove obsolete error reporting operator (@):

> Prior to PHP 8.0.0, the [error_reporting()](https://www.php.net/manual/en/function.error-reporting.php) called inside the custom error handler always returned 0 if the error was suppressed by the @ operator. As of PHP 8.0.0, it returns the value E_ERROR | E_CORE_ERROR | E_COMPILE_ERROR | E_USER_ERROR | E_RECOVERABLE_ERROR | E_PARSE.

## Reference
[Error Reporting Operator Manual](https://www.php.net/manual/en/language.operators.errorcontrol.php)

### DIFF
--- a/adodb-lib.inc.php
+++ b/adodb-lib.inc.php
@@ -1350,7 +1350,14 @@ function _adodb_backtrace($printOrArr=true, $maximumDepth=9999, $elementsToIgnor
 			}
 		}
 		$s .= $arr['function'] . '(' . implode(', ', $args) . ')';
-		$s .= @sprintf($fmt, $arr['line'], $arr['file'], basename($arr['file']));
+
+		// Disable error reporting directly from the library. Report according
+		// to ADOdb debug configuration
+		error_reporting(0);
+		ini_set('display_errors',0);
+		$s .= sprintf($fmt, $arr['line'], $arr['file'], basename($arr['file']));
+		error_reporting(E_ALL);
+		ini_set('display_errors',1);
 		$s .= "\n";
 	}
 	if ($html) {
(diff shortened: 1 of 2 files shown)


dregad:
While I'm generally in favor of removing usage of `@`, I'd like to understand the rationale for your proposed change. 

ADOdb does not rely on an error handler so I'm guessing it's somehow related to how you're using the library, and possibly within your own code that a change is actually needed.

As for the proposed patch, 
1. I'm not even sure why we have `@sprintf` in _adodb_backtrace() in the first place - I'm thinking we could probably just remove the `@` here instead. 
2. I don't think it's the right approach to suppress errors when executing a query. And even if it were, you should save the previous value and restore it afterwards, because nothing tells you that *display_errors* was set to `1` or _error_reporting_ to `E_ALL` before

dregad:
> Should i remove it then?

Better start by explaining why this is causing a problem in your code, and why the change is needed.

dregad:
> My goal is to restore the behavior by using the error_level function instead of the @ operator.

You're missing the point.

In my opinion it is conceptually wrong to squelch the warnings, i.e. it was not a good idea to put the `@` there in the first place. I believe that perpetuating that behavior is not the right approach.

If ADOdb were rewritten from scratch, it would systematically throw exceptions, and it would be up to the code using the library to handle them (or not).

This is why I'm asking you *why this is causing a problem in your code*, i.e. what kind of warnings/errors do you get on PHP 8.x that you don't in earlier versions, so we can address the root cause and not the symptoms.

dregad:
@ioigoume are you still working on this ?

dregad:
@ioigoume since you have not answered my questions or updated the code, I assume you are not interested in pursuing this so I'll close the PR. Feel free to reopen if needed.

PR #926, opened 2023-01-31
Fixed dynamic properties deprecation in PHP 8.2

meirzamoodle:


drivers/adodb-postgres64.inc.php
```diff
@@ -91,6 +91,15 @@ class ADODB_postgres64 extends ADOConnection{
 	var $version;
 	var $_nestedSQL = false;
 
+	/** @var array an array with client, protocol and server values. */
+	var $version;
+
+	/** @var bool */
+	var $_nestedSQL = false;
```
matt-land, line 98: duplicate, declared on line 92

adodb-xmlschema03.inc.php
```diff
@@ -1401,6 +1406,9 @@ class adoSchema {
 	 */
 	var $existingData;
 
+	/**  @var dbTable A table object. */
+	var $obj;
```
matt-land, line 1410: I didn't find any references to this in the class

drivers/adodb-sybase.inc.php
```diff
@@ -47,6 +47,12 @@ class ADODB_sybase extends ADOConnection {
 
 	var $port;
 
+	/** @var mixed */
```
matt-land, line 50: I only see this used as a bool (vs mixed), though its only read from and never set so I'm not sure

matt-land (changes requested):
All comments that reference "databaseName" could also say that "databaseName" is deprecated, preferring the value "database"

The change to drivers/adodb-postgress64.inc.php is the most pressing.

drivers/adodb-postgres64.inc.php
meirzamoodle: Hi @matt-land. I just updated it to remove those lines. Thanks.

matt-land (approved):

dregad:
All issues raised in this PR have been addressed in separate commits.

meirzamoodle:
> Out of curiosity, did you use some static analysis tool to detect and fix these issues ? If so it would be interesting to know which one.

Hi, @dregad. I'm using PHPStan to scan the all files.

dregad:
Thanks for the feedback. Was it the pro version that proposed the fixes, or did you code them by hand ?

meirzamoodle:
> Thanks for the feedback. Was it the pro version that proposed the fixes, or did you code them by hand ?

Sorry for the late reply. PHPStan is free.

PR #927, opened 2023-02-01
Use globally-defined ADODB_FETCH_MODE if it is false on the object in getAssoc

letynsoft:
In my Ubuntu 22.10, I've come to an issue with the getAssoc function. 
It was a version 5.21.4, but i believe the current implementation did not fix the underlying issue.

I think that in case the connection fetchMode is `false`, the global `$ADODB_FETCH_MODE` should be used instead.

### DIFF
--- a/adodb.inc.php
+++ b/adodb.inc.php
@@ -4237,6 +4237,7 @@ class ADORecordSet implements IteratorAggregate {
 	 */
 	function getAssoc($force_array = false, $first2cols = false)
 	{
+		global $ADODB_FETCH_MODE;
 		/*
 		* Insufficient rows to show data
 		*/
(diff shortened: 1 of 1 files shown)


dregad:
@mnewnham do you remember the reason for using the Connection's fetch mode here, instead of the RecordSet's ?

I'm thinking that since we now have $adodbFetchMode defined in the Recordset, it would make sense to use that and not rely on the Connection, what do you think ?

See also #958, which cleans up initialization of RecordSet $fetchMode.

dregad:
@mnewnham it would be great to have your feedback on this.

mnewnham:
If I remember correctly:
We use the fetchMode in getAssoc() rather than ADODB_FETCH_MODE because we use this method chained from other internals.  example
1. We start with a procedure that sets the $ADODB_FETCH_MODE to FETCH_ASSOC
2. We then call an internal function **MetaColumns()**. This requires that the fetch mode is FETCH_NUM
3. We save the current fetch mode in the MetaColumns method and switch $fetchMode to FETCH_NUM
4. We execute getAssoc with the fetchMode of FETCH_NUM even though ADODB_FETCH_MODE is FETCH_ASSOC
5. We return the results of getAssoc to MetaColumns, process it out, then switch the fetchMode back to FETCH_ASSOC

We have also talked about deprecating globals like ADODB_FETCH_MODE and bringing them into class variables because it prevents us controlling 2 simultaneous connections properly. For example if I open an Oracle connection in FETCH_NUM mode and a MySQL connection in FETCH_ASSOC mode, I have to reset the global before every call to the database.

PR #930, opened 2023-02-22
Suppress a PHP Notice

Unifex:
When running our behat tests we are seeing a PHP Notice that we tracked down to this location.

It appears that `$col_names` is an array and it is not zero-indexed. We appear to have a `$row` of the form:
```
0: "behat_usr_fir_ix"
1: "f"
2: "0"
```
In this case `$row[2]` is just a "0" so the first `$col` is "0" and there is no `$col_names[0]`, hence the notice.

I admit that I'm not up to speed with what is happening here, and I don't know how or why this is coming about in the first place. But this fix doesn't appear to be breaking anything else and the scenario that this showed up in was one that starts with an empty database, builds the tables, does a *lot* of CRUD, and uninstalls the tables at the end.

### DIFF
--- a/drivers/adodb-postgres64.inc.php
+++ b/drivers/adodb-postgres64.inc.php
@@ -691,7 +691,9 @@ class ADODB_postgres64 extends ADOConnection{
 		while ($row = $rs->FetchRow()) {
 			$columns = array();
 			foreach (explode(' ', $row[2]) as $col) {
-				$columns[] = $col_names[$col];
+				if (isset($col_names[$col])) {
+					$columns[] = $col_names[$col];
+				}
 			}
 
 			$indexes[$row[0]] = array(


mnewnham:
What database do you have on the backend?

dregad:
@Unifex Not really sure what a *behat test* is, but based on this

> We appear to have a `$row` of the form:
> 
> ```
> 0: "behat_usr_fir_ix"
> 1: "f"
> 2: "0"
> ```
> In this case $row[2] is just a "0" 

$row[2] is in `pg_catalog.pg_index.indkey` which, according to [PostgreSQL documentation](https://www.postgresql.org/docs/current/catalog-pg-index.html) is *an array of indnatts values that indicate which table columns this index indexes. [...] **A zero in this array indicates that the corresponding index attribute is an expression over the table columns, rather than a simple column reference**.*

So if the index is calculated then of course there would be no matching entry in the MetaColumns data ($col_names array).

Can you please confirm this is the case for your *behat_usr_fir_ix* index ?

It would be helpful, to reproduce the problem, if you could post the SQL to create the offending table and index.

dregad:
@Unifex any chance you can provide feedback on this by tomorrow night ? I'm almost ready to cut 5.22.5 and this could go in if I have confirmation of the behavior's root cause

dregad:
OK, I was able to replicate the problem with the following

```sql
CREATE TABLE test (
    id integer NOT NULL,
    name character varying(100)
);
ALTER TABLE ONLY test
    ADD CONSTRAINT test_pkey PRIMARY KEY (id);

CREATE UNIQUE INDEX name_idx ON test USING btree (name);
CREATE INDEX test_lower_name_idx ON test USING btree (lower((name)::text));
CREATE INDEX test_2colcalc_idx ON test USING btree (id, lower((name)::text));
```

```php
$db = ADONewConnection('pgsql');
$db->connect('localhost', $username, $password, $database);
var_export($db->metaindexes('test'));

// Warning: Undefined array key 0 in .../adodb/drivers/adodb-postgres64.inc.php on line 694
```

That said, I don't like the proposed fix, because MetaIndexed() entirely skips the index column if it does not match the table column, which could be confusing; IMO it would be more appropriate to set it to NULL so the caller knows that the index has an expression-based column.

dregad:
Closing this as won't fix - follow-up in #940.

PR #917, opened 2023-01-07
pdo: Creation of dynamic property deprecated warning in PHP 8.2

reportico-web:
Fatal Error: drivers/adodb-pdo.inc.php Line 605 - Creation of dynamic property ADODB_pdo::$_stmt is deprecated
Fatal Error: drivers/adodb-pdo.inc.php Line 788 - Creation of dynamic property ADORecordSet_pdo::$adodbFetchMode is deprecated

PR #944, opened 2023-03-29
Fix php81 deprecation warning

asabirov:
On PHP 8.2 with php_pdo extensions in the errror log, you might see these warnings:

Deprecated in .../vendor/adodb/adodb-php/drivers/adodb-pdo.inc.php line 788: Creation of dynamic property ADORecordSet_pdo::$adodbFetchMode is deprecated

and 

Deprecated in /data/vendor/adodb/adodb-php/drivers/adodb-pdo.inc.php line 605: Creation of dynamic property ADODB_pdo::$_stmt is deprecated

That PR fixes that.

### DIFF
--- a/drivers/adodb-pdo.inc.php
+++ b/drivers/adodb-pdo.inc.php
@@ -853,6 +853,7 @@ class ADORecordSet_pdo extends ADORecordSet {
 
 	/** @var PDOStatement */
 	var $_queryID;
+	var $adodbFetchMode;
 
 	function __construct($id,$mode=false)
 	{
--- a/drivers/adodb-pdo.inc.php
+++ b/drivers/adodb-pdo.inc.php
@@ -87,6 +87,7 @@ class ADODB_pdo extends ADOConnection {
 	var $_errorno = false;
 
 	var $_stmt = false;
+	var $_stmt;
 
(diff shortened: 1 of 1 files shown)


dregad:
@asabirov Thanks for your contribution. 

Did you check whether these warnings are still present with the latest *master* or *hotfix/5.22* branches ?
I believe they have been fixed already, see 6486f117d9cba8d532ef6adfb92f3ccd515c805b (#917).

ADOdb 5.22.5 including these fixes should be released soon.

dregad:
> There exists also `#[\AllowDynamicProperties]` as alternative that can be set in front of a class definition.

Yes I know, but that amounts to hiding the problem instead of fixing it, so I prefer not to do that if it can be avoided.

dregad:
5.22.5 is out now. 

@asabirov you have not provided any feedback so I'm assuming the problem is fixed. Feel free to reopen if not.

DanielRuf:
We can still see related warnings like the following:

> Deprecated: Creation of dynamic property PpGridAreaRecord::$show_empty_section is deprecated in vendor\adodb\adodb-php\adodb-active-record.inc.php on line 115

PR #950, opened 2023-04-12
Remove legacy work arounds for for PGSQL v7.1 (Nested SQL), and bytea_output

Mike-Benoit:
Remove legacy work arounds for PGSQL v7.1 (Nested SQL), and PostgreSQL 9.0 to 9.2 bytea_output escaping work-around. 

PostgreSQL 10 was end-of-life as of November 2022, and these cause additional SQL queries which hurts performance.


dregad:
Thanks for your contribution

PR #952, opened 2023-04-12
Minor fixes to avoid PHP warnings in v8.2+

Mike-Benoit:
Minor fixes to avoid PHP warnings in v8.2+.

See comments in code for further details if its not a simple object property initialization.

### DIFF
--- a/adodb.inc.php
+++ b/adodb.inc.php
@@ -4023,10 +4023,8 @@ class ADORecordSet implements IteratorAggregate {
 	var $_atLastPage = false;	/** Added by Iván Oliva to implement recordset pagination */
 	var $_lastPageNo = -1;
 	var $_maxRecordCount = 0;
-	var $rowsPerPage;
 	var $datetime = false;
 
-	var $oldProvider;
 
 	public $customActualTypes;
 	public $customMetaTypes;
@@ -4067,10 +4065,7 @@ class ADORecordSet implements IteratorAggregate {
 	}
 
 	function __destruct() {
-		//If the _queryID is bogus, calling Close() may cause an error in the database driver.
-		if($this->_queryID != -1) {
-			$this->Close();
-		}
+		$this->Close();
 	}
 
 	#[\ReturnTypeWillChange]
@@ -4796,7 +4791,8 @@ class ADORecordSet implements IteratorAggregate {
 		// free connection object - this seems to globally free the object
 		// and not merely the reference, so don't do this...
 		// $this->connection = false;
-		if (!$this->_closed) {
+		// Also ensure the _queryID is not bogus, as calling Close() may cause an error in the database driver.
+		if (!$this->_closed && $this->_queryID != -1) {
 			$this->_closed = true;
 			return $this->_close();
 		} else
--- a/adodb.inc.php
+++ b/adodb.inc.php
@@ -4023,7 +4023,10 @@ class ADORecordSet implements IteratorAggregate {
 	var $_atLastPage = false;	/** Added by Iván Oliva to implement recordset pagination */
 	var $_lastPageNo = -1;
 	var $_maxRecordCount = 0;
+	var $rowsPerPage;
 	var $datetime = false;
+
+	var $oldProvider;
 
 	public $customActualTypes;
 	public $customMetaTypes;
@@ -4064,7 +4067,10 @@ class ADORecordSet implements IteratorAggregate {
 	}
 
 	function __destruct() {
-		$this->Close();
+		//If the _queryID is bogus, calling Close() may cause an error in the database driver.
+		if($this->_queryID != -1) {
+			$this->Close();
+		}
 	}
 
 	#[\ReturnTypeWillChange]
--- a/drivers/adodb-postgres64.inc.php
+++ b/drivers/adodb-postgres64.inc.php
@@ -272,7 +272,7 @@ class ADODB_postgres64 extends ADOConnection{
 		if ($this->_connectionID) {
 			return "'" . pg_escape_string($this->_connectionID, $s) . "'";
 		} else {
-			//If there is no database connection, then pg_escape_string() will not work in PHP 8.3+.
+			//If there is no database connection, then pg_escape_string() will not work in PHP 8.1+.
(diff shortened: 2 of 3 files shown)


adodb.inc.php
```diff
@@ -4023,8 +4023,11 @@ class ADORecordSet implements IteratorAggregate {
 	var $_atLastPage = false;	/** Added by Iván Oliva to implement recordset pagination */
 	var $_lastPageNo = -1;
 	var $_maxRecordCount = 0;
+	var $rowsPerPage;
 	var $datetime = false;
 
+	var $oldProvider;
+
```
dregad, line 4030: The only reference to this _$oldProvider_ property in the whole code base is in adodb-csvlib.inc.php.

Do you actually need/use this ? If not, I would propose to drop the offending assignment instead of adding a new, useless property in the base class. 

What do you think ?

adodb.inc.php
```diff
@@ -4064,7 +4067,10 @@ function __construct($queryID,$mode=false) {
 	}
 
 	function __destruct() {
-		$this->Close();
+		//If the _queryID is bogus, calling Close() may cause an error in the database driver.
+		if($this->_queryID != -1) {
+			$this->Close();
+		}
```
dregad, line 4073: Wouldn't it be more appropriate to add this fix in ADORecordSet::Close() ?

drivers/adodb-postgres64.inc.php
```diff
@@ -272,7 +272,9 @@ function qStr($s, $magic_quotes=false)
 		if ($this->_connectionID) {
 			return "'" . pg_escape_string($this->_connectionID, $s) . "'";
 		} else {
-			return "'" . pg_escape_string($s) . "'";
+			//If there is no database connection, then pg_escape_string() will not work in PHP 8.3+.
+			//  This can occur when using setSessionVariables() in the load balancer, so fall back to emulated escaping in these cases.
+			return parent::qStr( $s );
```
dregad, line 277: I'm OK with this, but can you confirm where you got the information about it no longer working in 8.3 ?

The [pg_escape_string() documentation](https://www.php.net/manual/en/function.pg-escape-string.php) states that _As of PHP 8.1.0, using the default connection is deprecated_. This comes from the [PHP 8.1 deprecations RFC](https://wiki.php.net/rfc/deprecations_php_8_1#calling_overloaded_pgsql_functions_without_the_connection_argument).

Deprecated features are usually removed in the next major release, in that case that would be PHP 9.0.

dregad (changes requested):

adodb.inc.php
Mike-Benoit: Its been a decade since I've even looked at this code, but you are correct, it seems that $oldProvider is not used anywhere and should probably be removed entirely.

adodb.inc.php
Mike-Benoit: Agreed, that would make more sense.

drivers/adodb-postgres64.inc.php
Mike-Benoit: I would need to retest this. I seem to recall it triggering a fatal error, but it may have just been a deprecation warning. I can fix the comment, but either way something is needed here.

Mike-Benoit: Actually, I see, the old code was sending the $data argument in the first $connection argument position. PHP 8.2 changed to requiring a PgSQL\Connection instance for that argument, and thus it would fail. I have no idea how it worked at all before to be honest, but it was clearly incorrect.

dregad: I'm not really sure what *old code* and *$data argument* you are referring to.

Before PHP 8.1, it was possible not to specify the connection ID as first parameter to pg_escape_string(), in which case the default connection (i.e. last used one) was used; this is what was deprecated.

The function's signature also [changed the 1st parameter's type from *resource* to *PgSQL\Connection* object](https://github.com/php/doc-en/commit/c2eca73ef79ebe78cebb34053e41b565af504c4f#diff-3193d27b92fb4a075689186b96ff25aa97d2163286d5c45e24cc387bfbe01a72L16-R16) in 8.1, but the parameter is still marked as optional so other than the deprecation warning there should not be any errors.

Anyway, as mentioned before I'm fine with this change. I've opened #956 to track this.

dregad (changes requested):
OK with the csvlib and postgres driver changes, I'll go ahead commit those separately.

As for bogus queryID, more discussion is needed

adodb.inc.php
dregad: Opened #957 to track this

dregad:
You might want to update your branch to fix the conflict I introduced by modifying the comment in the postgres64 driver.

drivers/adodb-postgres64.inc.php
Mike-Benoit: Sorry, I was referring to the argument names that pg_escape_string() accept in the PHP documentation.

 ADODB was only passing a string to escape as the first argument in this case, instead of the expected connection resource or connection object. How it ever functioned is beyond me. But either way it's fixed now.

adodb.inc.php
Mike-Benoit: Our internal ADODB branch has been out-of-sync with upstream for over a decade now. It's possible we missed some upstream changes that resolved some of our issues in other ways. 

These are some of the last bits to bring it back in sync though, so that will be nice. We can leave this out for now and we will run further tests and let you know if there still a problem.

Thanks for your prompt response with these PR's.

dregad: You're welcome. So should I close this PR then ?

Mike-Benoit: Sure

dregad:
Sorry Mike, I just realized that when I cherry-picked the changes (059ffc2eeac134f293ac7e729672cc452137914a and b881dea297a93496fbd6d1d8a3c2fbe277c9231c) I forgot to update the commits' author to credit you for them. My apologies.

Mike-Benoit:
No worries at all.

dregad:
Closing this, as 3 of the changes have been covered separately (see #954, #956, #957).
The remaining point will be remain as-is for now (see discussion https://github.com/ADOdb/ADOdb/pull/952#discussion_r1167928783)

PR #959, opened 2023-04-16
Initialize $fetchMode in ADORecordSet's constructor

### LINKED ISSUE #958: Initialize ADORecordSet::$fetchMode in parent class
There is a lot of duplicated code in the RecordSet class hierarchy, as the logic to initialize the $fetchMode property from $ADODB_FETCH_MODE global is repeated in most drivers. It should be set in the parent class.

It would also make sense to initialize the recently-added $adodbFetchMode property (see #911) at the same time, as it is currently not consistently set except in those drivers that actually use it.

dregad:
Fixes #958


dregad:
The 2 extra commits are just code cleanup.

dregad:
@mnewnham if you can spare the time I would appreciate your review and testing on this. If you're busy just let me know and I'll merge as-is.

Issue #968, opened 2023-05-08
AdoDbPager - htmlSpecialChars setting are ignored on rendering


bblemm:
### Description
Currently on deactivating "htmlSpecialChars" the html code is always escaped.
In the fix #850 was the function parameter $htmlspecialchars check removed before execute htmlspecialchars() on variable.

### Environment
- ADOdb version: 5.22.5
- Driver or Module: mysqli
- Database type and version: MariaDB 10.6 
- PHP version: 8.1
- Platform: debian

* [x] I have tested that the problem is reproducible in the [latest release](https://github.com/ADOdb/ADOdb/releases/latest)

### Steps to reproduce
```
$sql = 'Select "this<br>is<br>html<br>code" as htmlCode';
$pager = new AdoDbPager($conn, $sql);
$pager->htmlSpecialChars = false;
$pager->Render();
```

### Expected behavior
Add $htmlspecialchars check befor use  htmlspecialchars() on variable.

dregad:
Confirmed regression introduced by commit bda6e1ea4271d0818dc22213e6a37269eea427b4.

Thanks for the pull request @bblemm .

2 remarks, independent from the bug
- *AdoDbPager*, I guess you actually meant *ADODB_Pager* ? Or is that a custom class of yours ?
- I don't know if you just used the Pager for demonstrating the problem, but if not I would **strongly** advise against setting `$pager->htmlSpecialChars = false;` unless you have full confidence about the underlying data, as this leaves you vulnerable to cross-site-scripting (XSS) attacks. Try it:

      $sql = "select 'XSS <script>alert(\"p4wned\")</script>' as htmlCode";

PR #969, opened 2023-05-08
#968 add checking htmlspecialchars setting

bblemm:


dregad:
Fixed in 93fa6785ad0c4f51f44d0e5a9055575e4337742b

PR #984, opened 2023-06-11
Fixes #981 for all error handlers in ADOdb 5.22x

### LINKED ISSUE #981: adodb_throw() does not respect @ operator on PHP 8
### Description
When using Exceptions, ADOdb is throwing an error even when the function is called with error suppression operator.

### Environment
- ADOdb version: 5.22.5
- Driver or Module: exceptions
- Database type and version: any
- PHP version: 8.0+
- Platform: any

* [x] I have tested that the problem is reproducible in the **master** branch / **hotfix** branch  

### Steps to reproduce
```php
include 'adodb.inc.php';
include 'adodb-exceptions.inc.php';
$db = ADONewConnection('pgsql');
$db->connect(...);
var_dump(@$db->insert_id());
```

PHP Fatal error:  Uncaught ADODB_Exception: postgres9 error: [-1: ERROR:  lastval is not yet defined in this session] in EXECUTE("SELECT lastval()") in /Users/dregad/dev/adodb/adodb-exceptions.inc.php:89

### Expected behavior
Output `bool(false)`, no exception thrown.

### Additional context 
Problem noticed while investigating #979.

NathanGibbs3:


dregad:
Thanks 😃

PR #992, opened 2023-06-27
oci8: deprecation warning in selecLimit() on PHP 8.1

wesyah234:
This is to prevent a PHP 8.2 warning

### DIFF
--- a/drivers/adodb-oci8.inc.php
+++ b/drivers/adodb-oci8.inc.php
@@ -805,7 +805,10 @@ END;
 		} else {
 			$hint = '';
 		}
-
+		// If non-bound statement, $inputarr is false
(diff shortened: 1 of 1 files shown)


dregad:
Thanks for your contribution.

I assume that by *null $inputArr* you actually mean that the parameter has not been provided and the variable is therefore set to its default value of `false` (not `null`), triggering the *Automatic conversion of false to array is deprecated* warning at line 892, which does not occur when $inputarr === null.

I'm fine with the proposed change, even though it may be cleaner to change the parameter's default value to `[]` - but I'd rather not change the function's signature in a hotfix release.

Also, this deprecation was introduced in PHP 8.1, not 8.2.

Issue #970, opened 2023-05-18
Remove obsolete Date/Time library


dregad:
It does not make much sense anymore in the age of 64-bit computing, where timestamps will not overflow for 292 billion years, and it relies on use of strftime()/gmstrftime(), which have been deprecated in PHP 8.1.

Follows discussion in https://github.com/ADOdb/ADOdb/issues/963

dregad:
[Documentation has been updated](https://adodb.org/dokuwiki/doku.php?id=v5:datetime:datetime_index) with warning boxes indicating that the library is deprecated in 5.22.6 and removed in 5.23.0.

ptdesign:
adodb_mktime removed unfinished. still adodb_mktime present in adodb.inc.php and adodb-sybase.inc.php

dregad:
> adodb_mktime removed unfinished. still adodb_mktime present in adodb.inc.php and adodb-sybase.inc.php

@ptdesign yes, I noticed that a couple days ago as well. Will fix.

ptdesign:
Hello, any updates?

dregad:
Sorry no, I've been busy and have not been able to work on ADOdb lately... But it's not forgotten and I'll get to it eventually.

joebordes:
Can I try to clean that up?  Make a PR?

dregad:
@joebordes contributions are always welcome, feel free to submit a pull request. Thanks !

joebordes:
Got it! PR on it's way :-)

joebordes:
Can you please add the `hacktoberfest` label to the project?

dregad:
> Can you please add the `hacktoberfest` label to the project?

Done

joebordes:
Looking at the code and the commit where the date/time library was eliminated I found that there were 4 functions being used and not defined:

- adodb_gmdate
- adodb_date2
- adodb_date
- adodb_mktime

I studied a little about the history and the functionality and decided (hopefully correctly) to just use the standard PHP functions.

Then I validated my assumptions by creating an SQLite database and trying to access it. I received the error as expected:

```shell
PHP Fatal error:  Uncaught TypeError: SQLite3::createFunction():
 Argument #2 ($callback) must be a valid callback, function "adodb_date" not found or invalid function name
 in /home/joe/reps/ADOdb/drivers/adodb-sqlite3.inc.php:312
```

[...]

PR #1002, opened 2023-09-30
Fix issue #970

joebordes:


dregad (changes requested):
I'd rather not have a test SQLite database stored in the repository.

If the DB cannot be created programmatically by the test script, then provide the necessary instructions for the developer to do it manually in a comment.

Otherwise, everything looks good to me.

joebordes:
deleted the database and added code to create it programmatically

dregad:
Many thanks for your contribution !

dregad:
@joebordes Apologies for resurrecting this after nearly 2 years, but I just realized that the switch from _adodb_date()_ to sqlite3 internal _strftime()_ function introduced by this PR is not correct and causes a regression, because it alters the behavior of the [SQLDate() method](https://adodb.org/dokuwiki/doku.php?id=v5:reference:connection:sqldate) which expects PHP-style date formats, while the code introduced by this PR switches to strftime format (e.g. `Ymd` => `%Y%m%d`).

I'll open a new PR to fix this regression.

joebordes:
No problem at all. I'm not sure I can help, but reach out as needed. I will try

Issue #1000, opened 2023-09-06
mysqli: bulkBind reset after one call


eposjk:
setting `$db->bulkBind=true` and making multiple `$db->execute()`-calls produce the error message _2D Array of values sent to execute and 'ADOdb_mysqli::bulkBind' not set_. It seems that bulkBind is being set to false at https://github.com/ADOdb/ADOdb/blob/f86db42bded690064ac1b38e293b33c5a2d1c313/drivers/adodb-mysqli.inc.php#L1123

Proposed Solution:

\+ `$bulkBind = $this->bulkBind;`
  `$this->bulkBind = false;`
  `$ret = $this->_execute($sql, $bulkTypeArray);`
\+ `$this->bulkBind = $bulkBind;`

PR #1003, opened 2023-10-05
Bugfix: mysqli bulkBind reset after one call #1000

Yendissian:
Saved bulkbind to a new variable so $this->bulkBind can be set again after being set to false. This will prevent the errors during multiple executions of $db->execute()


dregad:
Merged in 4b5f56ee5193e0f0e64711dfba226eabd7260d55, thanks for your contribution !

Issue #999, opened 2023-07-26
oci8: Prevent str_replace NULL error in qstr() methods on PHP 8.1


wesyah234:
Description
adodb-oci8.inc.php:1585 str_replace(): Passing null to parameter #3 ($subject) of type array|string is deprecated

Environment
ADOdb version: 5.22.6
Driver or Module: oci8
PHP version: 8.2
Platform: linux

Solved by changing:

		return  "'" . str_replace("'", $this->replaceQuote, $s) . "'";

to:
		return  "'" . str_replace("'", $this->replaceQuote, $s ?? '') . "'";

Thanks!

wesyah234:
This method is still giving me trouble.  I'm using v5.22.7 now, and I get this error:

adodb-php/drivers/adodb-oci8.inc.php:1582 strlen(): Passing null to parameter #1 ($string) of type string is deprecated

the strlen check you put in has the same issue as the str_replace, in that it's not happy when $s is null, so now the issue is showing up in the: 
`		
else if (strlen($s) == 0) {
			return "''";
		}
`

and it's addressed by changing it to 
`		else if (strlen($s ?? '') == 0) {
			return "''";
		}`

PR #1005, opened 2023-10-29
Prevent str_replace NULL test error in PHP 8.1 qstr() methods

mnewnham:


dregad (approved):
LGTM.

You deserve extra credits for creating the longest branch name I have ever seen 😉

dregad:
That said, you really should avoid using special chars like `$` and `|` in a branch name, as it causes Git to choke when referencing it without quotes

```
~/dev/adodb (master)$ git co 999-adodb-oci8incphp1585-str_replace-passing-null-to-parameter-3-$subject-of-type-array|string-is-deprecated
error: pathspec '999-adodb-oci8incphp1585-str_replace-passing-null-to-parameter-3--of-type-array' did not match any file(s) known to git
string-is-deprecated: command not found
```

mnewnham:
> That said, you really should avoid using special chars like `$` and `|` in a branch name, as it causes Git to choke when referencing it without quotes
> 
> ```
> ~/dev/adodb (master)$ git co 999-adodb-oci8incphp1585-str_replace-passing-null-to-parameter-3-$subject-of-type-array|string-is-deprecated
> error: pathspec '999-adodb-oci8incphp1585-str_replace-passing-null-to-parameter-3--of-type-array' did not match any file(s) known to git
> string-is-deprecated: command not found
> ```

Interestingly, this is the branch name that git generated for me, and I tend to accept them unread. I've seen this issue with special characters in the name before - jira creates even longer names for branches than git.

dregad:
> this is the branch name that git generated for me

Oh well 🤷 

Removed milestone and tags, as I did not realize that this was fixing exising issue #999, which was already properly tagged.

Issue #1009, opened 2023-11-07
ldap_connect() host with port signature deprecated for PHP 8.3


andrewnicols:
### Description
The secondary signautre for `ldap_connect()` which explicitly provides a host and port has been deprecated from PHP 8.3 onwards.
A Uri should be used instad.

https://wiki.php.net/rfc/deprecations_php_8_3#deprecate_calling_ldap_connect_with_2_parameters

### Environment
- ADOdb version: master (478c9cfea84b94ed2e8f131b6d6eb453bfda5db0)
- Driver or Module: ldap
- Database type and version: any
- PHP version: 8.3
- Platform: any

* [ ] I have tested that the problem is reproducible in the [latest release](https://github.com/ADOdb/ADOdb/releases/latest) / **master** branch / **hotfix** branch  
  ⚠️ Keep only the applicable option(s), i.e. where you actually tested and remove the others)

### Steps to reproduce
See use of `@ldap_connect()` with multiple arguments here: https://github.com/ADOdb/ADOdb/blob/478c9cfea84b94ed2e8f131b6d6eb453bfda5db0/drivers/adodb-ldap.inc.php#L71
```
			$this->_connectionID = @ldap_connect( $conn_info[0], $conn_info[1] );
```

dregad:
@andrewnicols out of curiosity, are you actually using the LDAP driver, or is this coming out of some static analysis ?

andrewnicols:
No, we're not using it outselves, but I went through all of the RFC changes for the PHP 8.3 release and hunted for potential issues across our codebase. We do let our users configure a wide range of ADOdb database types for their use, but it turns out that ldap isn't one of them.

I should add that I misread the RFC and that one isn't targetted until PHP 8.4. The RFC was identified as PHP 8.3 but many of the changes are only intended for 8.4 and I missed that in my initial assessment.

dregad:
Thanks for the feedback.

I was just checking because since I took over ADOdb maintenance back in 2013 I think this is the first time anyone ever opened an issue about the LDAP driver so I was curious to find out whether someone was actually using it.

> I misread the RFC and that one isn't targetted until PHP 8.4. The RFC was identified as PHP 8.3 but many of the changes are only intended for 8.4 and I missed that in my initial assessment.

You're confusing me here... I see no mention of 8.4 in the RFC

andrewnicols:
Then I'm confused too!

[...]

PR #1014, opened 2023-11-18
Update of LDAP Driver

mnewnham:
- Now supports all ADOdb methods using AD Query syntax instead of SQL
- Honors **ADODB_ASSOC_CASE**
- Honors **ADODB_FETCH_MODE**
- Supports **_setConnectionParameter()_**

### DIFF
--- a/drivers/adodb-ldap.inc.php
+++ b/drivers/adodb-ldap.inc.php
@@ -32,52 +32,92 @@ if (!defined('LDAP_ASSOC')) {
 }
 
 class ADODB_ldap extends ADOConnection {
-	var $databaseType = 'ldap';
-	var $dataProvider = 'ldap';
+	public $databaseType = 'ldap';
+	public $dataProvider = 'ldap';
+		
+	/*
+	* Caches the version info
+	*/
+	protected $version;
+	
 
-	# Connection information
-	var $username = false;
-	var $password = false;
-
-	# Used during searches
-	var $filter;
-	var $dn;
-	var $version;
-	var $port = 389;
-
-	# Options configuration information
-	var $LDAP_CONNECT_OPTIONS;
+	/*
+	* Options configuration information
+	* @deprecated 5.23
+	*/
+	public $LDAP_CONNECT_OPTIONS;
 
 	# error on binding, eg. "Binding: invalid credentials"
-	var $_bind_errmsg = "Binding: %s";
+	protected $_bind_errmsg = "Binding: %s";
 
-	// returns true or false
+	protected $connectionParameters = array(
+		LDAP_OPT_PROTOCOL_VERSION=>3,
+		LDAP_OPT_REFERRALS=>0
+	);
 
-	function _connect( $host, $username, $password, $ldapbase)
+	
+	/**
+	 * Connect to a AD Server
+	 *
+	 * @param string|null $ldapServer The AD Server to connect to.
+	 * @param string|null $username The username to connect as.
+	 * @param string|null $password The password to connect with.
+	 * @param string|null $ldapbase The Base DN
+	 *
+	 * @return bool|null True if connected successfully, false if connection failed, or null if the mysqli extension
+	 * isn't currently loaded.
+	 */
+	public function _connect( $ldapServer, $username, $password, $ldapbase)
 	{
 		global $LDAP_CONNECT_OPTIONS;
 
-		if ( !function_exists( 'ldap_connect' ) ) return null;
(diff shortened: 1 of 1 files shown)


drivers/adodb-ldap.inc.php
```diff
@@ -32,54 +32,95 @@
...
+			return null;
 
-			$this->_connectionID = @ldap_connect( $conn_info[0], $conn_info[1] );
+		
+		if (strpos($ldapServer,'ldap://') !== 0 && strpos($ldapServer,'ldaps://') !== 0) 
+		{
+			/*
+			* If ldap SSL not explicitly specified, fall back to insecure
+			*/ 
+			$ldapServer = sprintf('ldap://%s',$ldapServer);
```
dregad, line 83: I don't think that block is necessary. I would just pass $ldapServer as-is to ldap_connect(), which will accept a plain host name (without :port) and return false if it's not a valid URI.

drivers/adodb-ldap.inc.php
```diff
@@ -32,54 +32,95 @@
...
+			if ($this->debug) 
+				ADOConnection::outp($e);
+			
 			return false;
 		}
-		if( count( $LDAP_CONNECT_OPTIONS ) > 0 ) {
+
+		if( is_array($LDAP_CONNECT_OPTIONS) && count( $LDAP_CONNECT_OPTIONS ) > 0 ) 
+		{
+			/*
```
dregad, line 101: Should we output a deprecation warning here ? e.g. 
```php
trigger_error('$LDAP_CONNECT_OPTIONS is deprecated, use setConnectionParameter() instead', E_USER_DEPRECATED);
```

drivers/adodb-ldap.inc.php
```diff
@@ -150,48 +193,90 @@ function _connect( $host, $username, $password, $ldapbase)
...
 		$this->_connectionID = false;
 	}
 
-	function SelectDB($db) {
-		$this->database = $db;
+	/**
+	 * Switches the baseDN to a new value
+	 *
+	 * @param string $baseDN
+	 * @return void
```
dregad, line 259: Return type does not match, should be bool / true, or remove the `return true;` statement below.

dregad (changes requested):
Some preliminary comments after a quick look at the code.

I'll look in more details and run some actual tests from work (I don't have an LDAP server to test with at home).

I'll commit some whitespace, coding guidelines and PHPDoc fixes in a bit.

dregad:
Hi @mnewnham you never responded to my review comments.

I'll commit a targeted fix for #1009 in hotfix/5.22 branch for now, and set the milestone for this improvement to 5.23.0

drivers/adodb-ldap.inc.php
```diff
@@ -32,54 +32,95 @@
...
+	protected $version;
+	
 
-	# Options configuration information
-	var $LDAP_CONNECT_OPTIONS;
+	/*
+	* Options configuration information
+	* @deprecated 5.23
+	*/
+	public $LDAP_CONNECT_OPTIONS;
```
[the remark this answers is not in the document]
mnewnham: removed, per suggestion

drivers/adodb-ldap.inc.php
```diff
@@ -32,54 +32,95 @@
...
+	
+	/**
+	 * Connect to a AD Server
+	 *
+	 * @param string|null $ldapServer The AD Server to connect to.
+	 * @param string|null $username The username to connect as.
+	 * @param string|null $password The password to connect with.
+	 * @param string|null $ldapbase The Base DN
+	 *
+	 * @return bool|null True if connected successfully, false if connection failed, or null if the mysqli extension
```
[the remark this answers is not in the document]
mnewnham: fixed

drivers/adodb-ldap.inc.php
mnewnham: This is just backwards compatibility. You've always been able to pass the hostname without the jdap:// and have it fixed

drivers/adodb-ldap.inc.php
mnewnham: changed, per suggestion

drivers/adodb-ldap.inc.php
mnewnham: Fixed return definition, improved docblock

dregad (approved):
Thanks for the fixes. I'll resolve the merge conflict and merge the PR.

PR #1017, opened 2023-12-18
Closes #1016

### LINKED ISSUE #1016: PHP Fatal error on array_change_key_case, bool given instead of array
### Description
Once every few calls `MetaColumns` fails in `ADODB_mysqli`. This is when `$schemaArray = $this->getAssoc($SQL);` returns false instead of a result.

```
[18-Dec-2023 14:57:29 UTC] PHP Fatal error:  array_change_key_case in /REDACTED/classes/database/adodb/drivers/adodb-mysqli.inc.php:939
Stack trace:
#0 /REDACTED/classes/database/adodb/drivers/adodb-mysqli.inc.php(939): array_change_key_case()
#1 /REDACTED/classes/database/adodb/adodb-active-record.inc.php(438): ADODB_mysqli->MetaColumns()
#2 /REDACTED/classes/database/adodb/adodb-active-record.inc.php(156): ADODB_Active_Record->UpdateActiveTable()
#3 /REDACTED/api/cronjobs/cron.REDACTED.php(264): ADODB_Active_Record->__construct()
#4 /REDACTED/api/cronjobs/cron.REDACTED.php(234): nan_store_user_for_project()
```

### Environment
- ADOdb version: v5.22.7
- Driver or Module: mysqli
- Database type and version: 
- PHP version: 8.1
- Platform: Ubuntu 20.04.6 LTS

* [x] I have tested that the problem is reproducible in the **master** branch 
  ⚠️ Keep only the applicable option(s), i.e. where you actually tested and remove the others)

### Steps to reproduce
Detailed, step-by-step instructions to reproduce the behavior, including:
Starting a new active directory intance
`$activeDirectory = new ActiveDirectory();`

This does not happen every query but once in a few bunch of calls.

### Expected behavior
It is expected that no result from `getAssoc` is caught before executing `array_change_key_case` so any code executed after this does not halt.

I have implemented this temporary fix on my own environment for now:

```
		$schemaArray = $this->getAssoc($SQL);
		if (!$this->metaColumnsSQL)
			return $false;
```

This is in style with the return a few rows above it:

```
		$false = false;
		if (!$this->metaColumnsSQL)
			return $false;
```

### Additional context 
We are not sure what is causing the problem with not giving a result. If we log the query above, the last query logged when the fatal error occurs always returns a result when queried manually

edit: After catching the error in the described way and as in pull #1017, adodb reports `ADOdb error: MySQL server has gone away`. So it handles the fault itself after that. No fatal errors anymore

JessieDR:
Fixes Issue #1016 in the 5.22x branch.

It is expected that no result from getAssoc is caught before executing array_change_key_case so any code executed after this does not halt.

### DIFF
--- a/drivers/adodb-mysqli.inc.php
+++ b/drivers/adodb-mysqli.inc.php
@@ -931,6 +931,9 @@ class ADODB_mysqli extends ADOConnection {
 				   AND table_name='$table'";
 
 		$schemaArray = $this->getAssoc($SQL);
+		if (!$schemaArray)
+			return $false;
+		
 		$schemaArray = array_change_key_case($schemaArray,CASE_LOWER);
 
 		$rs = $this->Execute(sprintf($this->metaColumnsSQL,$table));


dregad:
Closed via 249b7fb08569ce576cd1bf26098e00701145c6ec

dregad:
This introduced a regression, because the Fetch Mode is not reset to its original value before returning from the function.

PR #1037, opened 2024-06-08
postgres: Issue when using schema+"" around table. Fix the name to "s…

valqk:
When using $autoQuote=true and having postrgesql schema, accessing table like "schema.table" will not work, because it is invalid name.
This pull request fixes that by making the name "schema"."table" which works.

### DIFF
--- a/adodb-lib.inc.php
+++ b/adodb-lib.inc.php
@@ -125,6 +125,13 @@ function _adodb_replace($zthis, $table, $fieldArray, $keyCol, $autoQuote, $has_a
 {
 	// Add Quote around table name to support use of spaces / reserved keywords
 	$table=sprintf('%s%s%s', $zthis->nameQuote,$table,$zthis->nameQuote);
+    if($zthis->dataProvider == "postgres" 
+      && substr($table,0,1) == '"' 
+      && substr($table,-1,1) == '"' 
+      && strstr($table,'.')) {
+        // handle postgresql "table" when using schema. "schema.table" is invalid, but "schema"."table" is OK.
+        $table = str_replace('.', '"."', $table);
+    }
 
 	if (count($fieldArray) == 0) return 0;
 


dregad:
Thanks for the report and pull request. I had a closer look.

The quoting behavior was introduced in #390 (5.21.0). 

Your patch fixes the schema use case, but would effectively prevent using replace() on a table containing a dot in its name. While I this is definitely not something I would do or recommend, it is technically allowed in many databases, [including PostgreSQL](https://www.postgresql.org/docs/current/sql-syntax-lexical.html#SQL-SYNTAX-IDENTIFIERS): *Quoted identifiers can contain any character, except the character with code zero*. Considering that using dots in table names is a bit of a corner case, I guess it could be acceptable to drop support for it.

I'm also wondering if the problem is not also present for other databases as well (don't have time to analyze and test). If so it would be better to make this generic.

@mnewnham thoughts ?

mnewnham:
You can use this kind of formatting, (i.e. special characters) with all of the databases Most of them allow a full unicode range _**if quoted**_, so yes, we should address the issue from a system-wide perspective. 

But are we trying to do cross-schema queries here? A usual approach would be to set the schema at connection time and use the table name without a specifier. Indeed all of our discussions about name quoting are about hardening the quoting, not interpreting and relaxing it. I believe I have answered a question about how to use a table name with a "." in it that was not a reference to the schema, just part of the table name.  

If we are talking about a general ability to use the schema name as part of a query table name then I would be very hesitant to  include dynamic testing of table names each time we construct a query. As far as I can see, it would have no value at all for 99.9% of users and would ultimately be misinterpreted in some usage. 

Maybe it would be better to have a standardized approach:

- A parameter _**tableIncludesSchema**_ default null, otherwise is the name of the default schema
- If null, table is quoted per normal
- If not null, The function associated with quoting _**nameQuote**_ would take responsibility for unpacking the table name. Any table that was not of the form "schema.table" would have the default schema prepended 
- The table would then be quoted appropriately

The user would then be left with a choice to make - does my table include schemas **OR** does my table contain special characters. I think expecting us to handle 2 out-of-bounds design methodologies simultaneously would be unreasonable

mnewnham:
>We currently only have a nameQuote() method in ADODB_DataDict class, and the private lib >function _adodb-quote_fieldname(). Are you saying we should provide a standard >nameQuote() method in ADOConnection, and call that instead of manually concatenating >$nameQuote property, as we do all over the place now ?

Yes, I think that would be an opportunity to standardize, but also yes, It's probably outside the bounds of this discussion

Neustradamus:
@dregad: The 5.22.10 has been released, it has been merged?

PR #1042, opened 2024-08-24
Localization: Add Georgian language

NorwayFun:
Hello,
Here's translation to Georgian language.

BR, Temuri


dregad:
Thanks for your contribution

Issue #919, opened 2023-01-15
Mysqli connection with SSL: Not working because of wrong socket/flags


Mecanik:
I've been wasting quite some time trying to make this work, after debugging everything and making sure that connection itself works from any client -> server.

So the following snippet works fine:

```
$connectionID = mysqli_init();

if (!$connectionID) {
	die('mysqli_init failed');
}

if (!mysqli_options($connectionID, MYSQLI_INIT_COMMAND, 'SET AUTOCOMMIT = 0')) {
	die('Setting MYSQLI_INIT_COMMAND failed');
}

if (!mysqli_options($connectionID, MYSQLI_OPT_CONNECT_TIMEOUT, 5)) {
	die('Setting MYSQLI_OPT_CONNECT_TIMEOUT failed');
}

mysqli_ssl_set($connectionID, '...key', '...cert', '...ca', null, null);

if (!mysqli_real_connect($connectionID, 'hostname', 'username', 'password', 'database', 3306, null, MYSQLI_CLIENT_SSL)) {
	die('Connect Error (' . mysqli_connect_errno() . ') ' . mysqli_connect_error());
}

echo 'Success... ' . mysqli_get_host_info($connectionID) . "\n";

mysqli_close($connectionID);
```

However the following does not work at all:

```
$db = newAdoConnection('mysqli');

$db->ssl_key = '...key';
$db->ssl_cert = '...cert';
$db->ssl_ca = '...ca';
$db->ssl_capath = null;
$db->ssl_cipher = null;

$db->debug = true;

$db->Connect( 'hostname', 'username', 'password', 'database');
```

The error description is nothing information. Just "could not connect to..." and stack trace. 

So after checking the following: https://github.com/ADOdb/ADOdb/blob/11d361bf7f19fb1709b4e0fdfba9ccd171adb409/drivers/adodb-mysqli.inc.php#L200

I noticed the socket and flags are in the wrong order here. As per the PHP [documentation](https://www.php.net/manual/en/mysqli.real-connect.php) it's in the wrong place: 

```
socket
Specifies the socket or named pipe that should be used.

Note: Specifying the socket parameter will not explicitly determine the type of connection to be used when connecting to the MySQL server. How the connection is made to the MySQL database is determined by the host parameter.

flags
With the parameter flags you can set different connection options:

Supported flags
Name	Description
MYSQLI_CLIENT_COMPRESS	Use compression protocol
[...]

PR #1043, opened 2024-08-27
MYSQLI SSL bug fixes

look-was-here:
I originally made the changes here https://github.com/ADOdb/ADOdb/pull/622 which caused some unfortunate issues that I didn't mean to happen, https://github.com/ADOdb/ADOdb/issues/919 I'd like to make changes to fix said issues I introduced. Removed inappropriate setting of socket, if a user wants to set a custom socket the user can set the class property, updates to only set the clientFlags to MYSQLI_CLIENT_SSL_DONT_VERIFY_SERVER_CERT if the clientFlags is set to the default 0 but the other ssl properties have been set specifying that the connection should be ssl.

### DIFF
--- a/drivers/adodb-mysqli.inc.php
+++ b/drivers/adodb-mysqli.inc.php
@@ -231,9 +231,18 @@ class ADODB_mysqli extends ADOConnection {
 
 		// SSL Connections for MySQLI
 		if ($this->ssl_key || $this->ssl_cert || $this->ssl_ca || $this->ssl_capath || $this->ssl_cipher) {
+
 			mysqli_ssl_set($this->_connectionID, $this->ssl_key, $this->ssl_cert, $this->ssl_ca, $this->ssl_capath, $this->ssl_cipher);
-			$this->socket = MYSQLI_CLIENT_SSL;
-			$this->clientFlags = MYSQLI_CLIENT_SSL_DONT_VERIFY_SERVER_CERT;
+
+			// Check for set SSL client flag, NOTE: bitwise operations.
+			$ssl_client = ($this->clientFlags & MYSQLI_CLIENT_SSL);
+			$ssl_verify_cert = ($this->clientFlags & MYSQLI_CLIENT_SSL_VERIFY_SERVER_CERT);
+			$ssl_dont_verify_cert = ($this->clientFlags & MYSQLI_CLIENT_SSL_DONT_VERIFY_SERVER_CERT);
+
+			// Add SSL client flag to client flag(s) if not set, NOTE: bitwise operations.
+			if (!$ssl_client && !$ssl_verify_cert && !$ssl_dont_verify_cert) {
+				$this->clientFlags |= MYSQLI_CLIENT_SSL_VERIFY_SERVER_CERT;
+			}
 		}
 
 		$ok = @mysqli_real_connect($this->_connectionID,
--- a/drivers/adodb-mysqli.inc.php
+++ b/drivers/adodb-mysqli.inc.php
@@ -234,14 +234,10 @@ class ADODB_mysqli extends ADOConnection {
 
 			mysqli_ssl_set($this->_connectionID, $this->ssl_key, $this->ssl_cert, $this->ssl_ca, $this->ssl_capath, $this->ssl_cipher);
 
-			// Check for set SSL client flag, NOTE: bitwise operations.
-			$ssl_client = ($this->clientFlags & MYSQLI_CLIENT_SSL);
-			$ssl_verify_cert = ($this->clientFlags & MYSQLI_CLIENT_SSL_VERIFY_SERVER_CERT);
-			$ssl_dont_verify_cert = ($this->clientFlags & MYSQLI_CLIENT_SSL_DONT_VERIFY_SERVER_CERT);
-
-			// Add SSL client flag to client flag(s) if not set, NOTE: bitwise operations.
-			if (!$ssl_client && !$ssl_verify_cert && !$ssl_dont_verify_cert) {
-				$this->clientFlags |= MYSQLI_CLIENT_SSL_VERIFY_SERVER_CERT;
+			// Check for any SSL client flag set, NOTE: bitwise operation.
+			if (!($this->clientFlags & MYSQLI_CLIENT_SSL)) {
+        			ADOConnection::outp('When using certificates, set the client flag MYSQLI_CLIENT_SSL_VERIFY_SERVER_CERT or MYSQLI_CLIENT_SSL_DONT_VERIFY_SERVER_CERT');
+				return false;
 			}
 		}
 


look-was-here:
Moved port and socket param changes into the base connect so any driver could set the properties on connect if they are specified.

adodb.inc.php
```diff
@@ -973,60 +974,91 @@ abstract protected function _connect($argHostname, $argUsername, $argPassword, $
...
+		if (!is_null($port)) {
+
+			// Use default port if cast yields zero.
+			$port = (int)$port != 0 ? (int)$port : 3306;
+		
+		} else {
+
+			// Use default port if cast yields zero.
+			$port = (int)$this->port != 0 ? (int)$this->port : 3306;
+		}
```
dregad, line 1005: This does not belong in the parent method - port 3306 is a MySQL-specific thing.

dregad (changes requested):
I am not so keen on changing ADOConnection::connect()'s signature, which IMO needs to remain simple and not contain any database-specific features that would not be supported across the board. @mnewnham what are your thoughts on this ?

Also, your commits contain quite a bit of unrelated whitespace changes, some of which add no value (e.g. adding blank lines). I am generally OK with anything that brings the code closer to PSR-12 coding style, but the rest should be removed; moreover such changes belong in a separate commit.

dregad:
Discussion following above review 
https://matrix.to/#/!yOADOjbRlsBmkCvHdi:gitter.im/$gM24sWPm7cB1NAyA20agjwlffAfH9VBnhQmIw-bksT8

mnewnham:
I'm just working through understanding this: I'll join on Element

adodb.inc.php
look-was-here: Reverted changes not related to fixing the inappropriate setting of the socket, adding a check to make sure something isn't already set for the clientFlags before setting a default, and updating the default to verify the server cert for better default security for SSL. Created a second pull request for updates to setConnectionParameter 
 from our discussion https://github.com/ADOdb/ADOdb/pull/1044

look-was-here:
Reverted changes not related to fixing the inappropriate setting of the socket, adding a check to make sure something isn't already set for the clientFlags before setting a default, and updating the default to verify the server cert for better default security for SSL. Created a second pull request for updates to setConnectionParameter
from our discussion https://github.com/ADOdb/ADOdb/pull/1044

drivers/adodb-mysqli.inc.php
```diff
@@ -231,9 +231,13 @@ function _connect($argHostname = null,
 
 		// SSL Connections for MySQLI
 		if ($this->ssl_key || $this->ssl_cert || $this->ssl_ca || $this->ssl_capath || $this->ssl_cipher) {
+
 			mysqli_ssl_set($this->_connectionID, $this->ssl_key, $this->ssl_cert, $this->ssl_ca, $this->ssl_capath, $this->ssl_cipher);
-			$this->socket = MYSQLI_CLIENT_SSL;
-			$this->clientFlags = MYSQLI_CLIENT_SSL_DONT_VERIFY_SERVER_CERT;
+
+			// Default client flag to ssl with verify if not set.
+			if (0 === $this->clientFlags) {
```
dregad, line 238: I don't think this check is correct. 

The client flags can be set to a number of different values independently (binary flag), so it can be != 0 without MYSQLI_CLIENT_SSL being set.

I believe you need to check for the appropriate value(s) with a binary and, e.g. `$this->clientFlags & MYSQLI_CLIENT_SSL` instead.

That being said, reading the [documentation](https://www.php.net/manual/en/mysqli.real-connect.php#mysqli.real-connect.flags), I must admit I'm a bit confused with these SSL flags, and which one(s) should be used for the above-mentioned check... 

 * `MYSQLI_CLIENT_SSL_DONT_VERIFY_SERVER_CERT` this one is clear
 *  `MYSQLI_CLIENT_SSL_VERIFY_SERVER_CERT` is only referenced in the predefined constants page, but not on mysqli_real_connect page.
 * `MYSQLI_CLIENT_SSL` means *Use SSL (encryption)* but [here](https://www.php.net/manual/en/mysqli.constants.php#constant.mysqli-client-ssl) it is stated that it *is set internally in the MySQL client library* (without specifying how or when) and that *should not be set by application programs*

drivers/adodb-mysqli.inc.php
```diff
@@ -231,9 +231,13 @@ function _connect($argHostname = null,
...
+				$this->clientFlags = MYSQLI_CLIENT_SSL_VERIFY_SERVER_CERT;
```
dregad, line 239: Considering that clientFlags is a binary flag, a straight assignment will potentially overwrite the user's flags.

You should OR the value in `$this->clientFlags |= MYSQLI_CLIENT_SSL_VERIFY_SERVER_CERT;` instead

dregad (changes requested):

drivers/adodb-mysqli.inc.php
look-was-here: Updated to check for set bitwise options MYSQLI_CLIENT_SSL_DONT_VERIFY_SERVER_CERT and MYSQLI_CLIENT_SSL_VERIFY_SERVER_CERT  but I am still unsure if I should add MYSQLI_CLIENT_SSL as an additional check...

drivers/adodb-mysqli.inc.php
```diff
@@ -231,9 +231,18 @@ function _connect($argHostname = null,
...
 		if ($this->ssl_key || $this->ssl_cert || $this->ssl_ca || $this->ssl_capath || $this->ssl_cipher) {
+
 			mysqli_ssl_set($this->_connectionID, $this->ssl_key, $this->ssl_cert, $this->ssl_ca, $this->ssl_capath, $this->ssl_cipher);
-			$this->socket = MYSQLI_CLIENT_SSL;
-			$this->clientFlags = MYSQLI_CLIENT_SSL_DONT_VERIFY_SERVER_CERT;
+
+			// Check for set SSL client flag, NOTE: bitwise operations.
+			$ssl_client = ($this->clientFlags & MYSQLI_CLIENT_SSL) == MYSQLI_CLIENT_SSL;
+			$ssl_verify_cert = ($this->clientFlags & MYSQLI_CLIENT_SSL_VERIFY_SERVER_CERT) == MYSQLI_CLIENT_SSL_VERIFY_SERVER_CERT;
+			$ssl_dont_verify_cert = ($this->clientFlags & MYSQLI_CLIENT_SSL_DONT_VERIFY_SERVER_CERT) == MYSQLI_CLIENT_SSL_DONT_VERIFY_SERVER_CERT;
```
dregad, line 240: The `== MYSQLI_CLIENT_xxxx` bit is not necessary, the bitwise AND will return either 0 or a non-zero value, which PHP will evaluate as TRUE.

drivers/adodb-mysqli.inc.php
```diff
@@ -231,9 +231,18 @@ function _connect($argHostname = null,
...
+
+			// Add SSL client flag to client flag(s) if not set, NOTE: bitwise operations.
+			if (false == $ssl_client && false == $ssl_verify_cert && false == $ssl_dont_verify_cert) {
```
dregad, line 243: `false == $x` would be better written as boolean logic `!$x`.

dregad (changes requested):
> I am still unsure if I should add MYSQLI_CLIENT_SSL as an additional check...

Yeah, I don't know either... The documentation is not clear, and I do not have an environment to test. Do you ?

If not, maybe ask the PHP doc team to clarify ?

dregad (approved):
Looks good to me now. Thanks a lot !

Let me know if you're happy with me merging it as-is, or if you're doing more tests/research with regards to the MYSQLI_CLIENT_SSL flag.

mnewnham:
I'm adding myself as a reviewer. I hope to get through it in the next couple of hours

mnewnham:
So looking at this, the recent changes that have been made look ok, but there appears to be some problems with the original code that have lain undetected for a while

#### MYSQLI_CLIENT_SSL Isn't a socket parameter, but a client flag
In our existing code, we have 

````php
// SSL Connections for MySQLI
if ($this->ssl_key || $this->ssl_cert || $this->ssl_ca || $this->ssl_capath || $this->ssl_cipher) {
    mysqli_ssl_set($this->_connectionID, $this->ssl_key, $this->ssl_cert, $this->ssl_ca, $this->ssl_capath, $this->ssl_cipher);
    $this->socket = MYSQLI_CLIENT_SSL;
    $this->clientFlags = MYSQLI_CLIENT_SSL_DONT_VERIFY_SERVER_CERT;
}
````
But this isn't correct, it should be

````php
// SSL Connections for MySQLI
if ($this->ssl_key || $this->ssl_cert || $this->ssl_ca || $this->ssl_capath || $this->ssl_cipher) {
    mysqli_ssl_set($this->_connectionID, $this->ssl_key, $this->ssl_cert, $this->ssl_ca, $this->ssl_capath, $this->ssl_cipher);
    $this->clientFlags = MYSQLI_CLIENT_SSL|MYSQLI_CLIENT_SSL_DONT_VERIFY_SERVER_CERT;
}
````
#### Adding a certificate forces server certificate validation (and all other client Flags) to switch off
If I define a set of Parameters before connection, e.g, 

```
$db->clientFlags = MYSQLI_CLIENT_COMPRESS|MYSQLI_CLIENT_SSL;
$db->setConnectionParameter(MYSQLI_CLIENT_SSL_VERIFY_SERVER_CERT,1);
$db->ssl_ca = '/cert/DigiCertGlobalRootCA.crt.pem';
```
Because I have included a certificate name, then
1. The MYSQLI_CLIENT_SSL_VERIFY_SERVER_CERT gets overloaded by the MYSQLI_CLIENT_SSL_DONT_VERIFY_SERVER_CERT parameter
2. both the passed clientFlags are zapped and replaced by MYSQLI_CLIENT_SSL_DONT_VERIFY_SERVER_CERT

The real-world consequence of this is that if I am using a secure transport, I get an error,**Could not connect : Connections using insecure transport are prohibited while --require_secure_transport=ON.**. It's impossible to bypass this without hacking the function. But if I disable the certificate, the connection will work. One point I don't understand is why we force the `MYSQLI_CLIENT_SSL_DONT_VERIFY_SERVER_CERT` parameter in the first place.

#### How to fix this
1. I suggest pulling the `clientFlags` into `setConnectionParameter()` for cleanness
2. Fix the socket issue. This is generally **null**, but if set would be a string, e.g. **/home/mysql/mysql.sock**
3. Untangle the default setting of `clientFlags` , ensuring we don't overwrite any client flags already set

**Comment by dregad** [comment]
Hi Mark, thanks for the review. 

> #### MYSQLI_CLIENT_SSL Isn't a socket parameter, but a client flag
> #### Adding a certificate forces server certificate validation (and all other client Flags) to switch off

> 2. Fix the socket issue. This is generally **null**, but if set would be a string, e.g. **/home/mysql/mysql.sock**

> 3. Untangle the default setting of `clientFlags` , ensuring we don't overwrite any client flags already set

I spotted the same issues, and I believe are all covered by this PR now following my earlier review, see  https://github.com/ADOdb/ADOdb/pull/1043#pullrequestreview-2274257427.


> 1. I suggest pulling the `clientFlags` into `setConnectionParameter()` for cleanness

This is covered by PR #1044.

**Comment by mnewnham** [comment]
I'm wondering if these operations inside the if(certificate) branch adds too much unneeded complexity, 

1. If you have a certificate and no clientFlags, It auto adds `MYSQLI_CLIENT_SSL_VERIFY_SERVER_CERT `. But it doesn't tell you it is doing it. 
     - If the required connection actually requires `MYSQLI_CLIENT_SSL_DONT_VERIFY_SERVER_CERT` then you don't actually get any help with the error, it just fails to connect.
     - It also silently changes the passed in parameters, so that if you retrieve them for re-use, they are different from what is originally set.
2. It seems that when you have either of the above set, it automatically activates `MYSQLI_CLIENT_SSL` anyway, so for example
`MYSQLI_CLIENT_SSL_VERIFY_SERVER_CERT|MYSQLI_CLIENT_SSL` is functionally the same as `MYSQLI_CLIENT_SSL_VERIFY_SERVER_CERT `
3. Would it be better then to just do something like this?

````php
if ($this->ssl_key || $this->ssl_cert || $this->ssl_ca || $this->ssl_capath || $this->ssl_cipher) {
  mysqli_ssl_set($this->_connectionID, $this->ssl_key, $this->ssl_cert, $this->ssl_ca, $this->ssl_capath, $this->ssl_cipher);

  if ($this->clientFlags == 0 || $this->clientFlags == MYSQLI_CLIENT_SSL)
  {
       ADOConnection::outp('When using certificates, set one of MYSQLI_CLIENT_SSL_VERIFY_SERVER_CERT or MYSQLI_CLIENT_SSL_DONT_VERIFY_SERVER_CERT in ADOConnection::clientFlags');
      return false;
}
````

**Comment by dregad** [comment]
> 3. Would it be better then to just do something like this?

It's a different approach, forcing the user to set the client flags appropriately, instead of trying to second guess and preset a default. I am OK with your suggestion. 

That being said, I don't think the condition
> ```php
>   if ($this->clientFlags == 0 || $this->clientFlags == MYSQLI_CLIENT_SSL)
> ```

is correct, due to clientFlags being a binary flag. Please read https://github.com/ADOdb/ADOdb/pull/1043#discussion_r1740018914.  I did not actually test this, but based on what you wrote:

> when you have either of the above set, it automatically activates MYSQLI_CLIENT_SSL

I believe it should be 
```php
if (!($this->clientFlags & MYSQLI_CLIENT_SSL)) {
        ADOConnection::outp('When using certificates, set one of MYSQLI_CLIENT_SSL_VERIFY_SERVER_CERT or MYSQLI_CLIENT_SSL_DONT_VERIFY_SERVER_CERT in ADOConnection::clientFlags');
```

And the error message probably needs to be reworded, considering that we are moving towards having the user set the client flags via setConnectionParameter() in #1044.

mnewnham:
> > 3. Would it be better then to just do something like this?
> 
> It's a different approach, forcing the user to set the client flags appropriately, instead of trying to second guess and preset a default. I am OK with your suggestion.
> 
> That being said, I don't think the condition
> 
> > ```
> >   if ($this->clientFlags == 0 || $this->clientFlags == MYSQLI_CLIENT_SSL)
> > ```
> 
> is correct, due to clientFlags being a binary flag. Please read [#1043 (comment)]

I understand, my suggestion was really meant as a design concept although I think in this situation it might work.

Neustradamus:
To follow this PR :)

look-was-here:
I've made additional changes moving away from implicitly setting the SSL client flag and instead adding a friendly error letting the user know to set the client flag. I appreciate your time and efforts looking into this @mnewnham and @dregad, many thanks!

dregad (approved):
Thank you @look-was-here. 

@mnewnham I think this addresses [your comments](https://github.com/ADOdb/ADOdb/pull/1043#issuecomment-2325243176)
Let me know if you're happy with it

Neustradamus:
@mnewnham: Have you seen @look-was-here changes?

Neustradamus:
It will be merged?
We are in 2025 now.

dregad:
Merged in be251d3ff570110b13f6b9577a372c5898980e19. Thanks @look-was-here for your contribution

PR #1044, opened 2024-09-01
Improve MySQLi setConnectionParameter

look-was-here:
Updates to allow setting of ssl cert property values, clientFlags, port, and socket using the standard setConnectionParameter functionality.

### DIFF
--- a/drivers/adodb-mysqli.inc.php
+++ b/drivers/adodb-mysqli.inc.php
@@ -152,22 +152,74 @@ class ADODB_mysqli extends ADOConnection {
 	}
 
 	/**
-	 * Adds a parameter to the connection string.
+	 * Adds a parameter to the connection string, can also set connection property values.
 	 *
 	 * Parameter must be one of the constants listed in mysqli_options().
 	 * @see https://www.php.net/manual/en/mysqli.options.php
-	 *
-	 * @param int    $parameter The parameter to set
-	 * @param string $value     The value of the parameter
+	 * 
+	 * OR 
+	 * 
+	 * Parameter must be a string matching one of the following special cases.
+	 * 'ssl' - SSL values e.g. ('ssl' => ['ca' => '/path/to/ca.crt.pem'])
+	 * 'clientflags' - Client flags of type 'MYSQLI_CLIENT_'
+	 * @see https://www.php.net/manual/en/mysqli.real-connect.php
+	 * @see https://www.php.net/manual/en/mysqli.constants.php
+	 * 'socket' - The socket or named pipe that should be used
+	 * 'port' - The port number to attempt to connect to the MySQL server
+	 * 
+	 * @param string|int $parameter The parameter to set
+	 * @param string|int|array $value The value of the parameter
 	 *
 	 * @return bool
 	 */
 	public function setConnectionParameter($parameter, $value) {
-		if(!is_numeric($parameter)) {
-			$this->outp_throw("Invalid connection parameter '$parameter'", __METHOD__);
-			return false;
+
+		// Special case for setting SSL values.
+		if ("ssl" === $parameter && is_array($value)) {
+			if (isset($value["key"])) {
+				$this->ssl_key = $value["key"];
+			}
+			if (isset($value["cert"])) {
+				$this->ssl_cert = $value["cert"];
+			}
+			if (isset($value["ca"])) {
+				$this->ssl_ca = $value["ca"];
+			}
+			if (isset($value["capath"])) {
+				$this->ssl_capath = $value["capath"];
+			}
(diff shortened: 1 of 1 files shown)


drivers/adodb-mysqli.inc.php
```diff
@@ -157,16 +157,64 @@ function SetTransactionMode($transaction_mode)
...
 	 *
 	 * @return bool
 	 */
 	public function setConnectionParameter($parameter, $value) {
-		if(!is_numeric($parameter)) {
+
+		// Special case for setting SSL cert values.
+		if ("sslcert" === $parameter && is_array($value)) {
+
+			if (true === isset($value["ssl_key"])) {
```
dregad, line 170: isset() returns a bool, so `true ===` is not necessary

I also suggest naming the parameter just `ssl`, instead of `sslcert` as some of the values are not about the certificate (e.g. cipher). 

Do we need the `ssl_` prefix in the array key ? It feels redundant.

drivers/adodb-mysqli.inc.php
```diff
@@ -157,16 +157,64 @@ function SetTransactionMode($transaction_mode)
...
+			return true;
+		}
+
+		// Special case for setting the socket.
+		if ("socket" === $parameter && is_string($value)) {
+			$this->socket = $value;
+			return true;
+		}
+
+		// Special case for setting the client flag.
```
dregad, line 206: Surely you mean *port number*...

drivers/adodb-mysqli.inc.php
```diff
@@ -157,16 +157,64 @@ function SetTransactionMode($transaction_mode)
...
+			}
+
+			return true;
+
+		}
+
+		// Special case for setting the client flag.
+		if ("clientflags" === $parameter && is_numeric($value)) {
+			$this->clientFlags = $value;
+			return true;
```
dregad, line 197: Shouldn't we return `false` if $value is not numeric ? 

With this implementation, ADOConnection::setConnectionParameter() will just blindly set $connectionParameters property to whatever has been given, probably resulting in an Exception to be thrown later in _connect().

https://github.com/ADOdb/ADOdb/blob/1e0e602ed72628f7a6665f5c925e93a80bdbd710/drivers/adodb-mysqli.inc.php#L219-L222

NOTE: This comment also applies to the other special cases with a type check that you added (sslsert, socket, port).

dregad (changes requested):

look-was-here:
Made feedback updates, thank you!

dregad (changes requested):
To reduce code duplication, I would suggest to invert the logic of the last test, so there is only one single outp_throw / return false at the end.

```php
if (is_numeric($parameter)) {
	return parent::setConnectionParameter($parameter, $value);
}

$this->outp_throw("Invalid connection parameter '$parameter'", __METHOD__);
return false;
```

drivers/adodb-mysqli.inc.php
```diff
@@ -157,17 +157,67 @@ function SetTransactionMode($transaction_mode)
 	 * Parameter must be one of the constants listed in mysqli_options().
```
dregad, line 157: This needs to be updated to reflect the newly added possible values.

dregad (changes requested):
Loking pretty good, almost there...

- PHPDoc update to reflect usage with non-numeric param
- nitpick: use a single if statement with ANDed expressions

PR is still marked as draft, don't forget to mark it as ready to review whenever you are.

Thanks !

drivers/adodb-mysqli.inc.php
look-was-here: Updated PHPdoc with new possible values.

look-was-here:
Made additional feedback updates, thank you!

Neustradamus:
To follow this PR :)

dregad:
> To follow this PR :)

@Neustradamus next time, instead of spamming the thread with a useless comment, just click the *Subscribe* button on the right... Thanks for your understanding.

image

Neustradamus:
I have a lot of subscriptions, I need to comment to receive an e-mail...

Neustradamus:
@dregad: I confirm you, I subscribe a lot, but I received only an e-mail if I comment. It is normal.

look-was-here:
I think this is ready for review again, thank you!

dregad (approved):
Looks good to me, thank you @look-was-here

Neustradamus:
@dregad: You have approved but it was not merged, a reason?

Neustradamus:
It will be merged?
We are in 2025 now.

dregad:
Merged in 8d97170e8016b969505a29b8cd39f001294653a2, thanks @look-was-here !

PR #1053, opened 2024-12-13
Enable DROP and RENAME COLUMN support for SQLite

raspopov:
SQlite now supports a subset of [ALTER TABLE](https://www.sqlite.org/lang_altertable.html) operations with some limitations. I think it's up to the user to decide how to use it, and the capability should be there anyway.

### DIFF
--- a/datadict/datadict-sqlite.inc.php
+++ b/datadict/datadict-sqlite.inc.php
@@ -91,16 +91,4 @@ class ADODB2_sqlite extends ADODB_DataDict {
 		return array();
 	}
 
-	function DropColumnSQL($tabname, $flds, $tableflds='', $tableoptions='')
-	{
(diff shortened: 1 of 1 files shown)


dregad:
Thanks, I rarely use SQLite myself, and was not aware they added support for these operations. After checking the change log:

- RENAME COLUMN added in [SQLite 3.25.0](https://www.sqlite.org/releaselog/3_25_0.html) (2018)
- DROP COLUMN added in [SQLite 3.35.0](https://www.sqlite.org/releaselog/3_35_0.html) (2021)

Even though it's somewhat unlikely that people would still be using 3, respectively 6 years old software, to ensure backwards compatibility I'll add a version check instead of just dropping the methods override.

PR #1055, opened 2024-12-17
Fulfill the SQLite PRIMARY KEY AUTOINCREMENT requirements.

raspopov:
This patch extends ADODB_DataDict::_CreateSuffix() with two additional parameters - `$fprimary` (PRIMARY KEY flag) and `& $pkey` (reference to array of primary keys). The ADODB2_sqlite::_CreateSuffix() has been modified to do a special processing of the PRIMARY KEY AUTOINCREMENT column to make SQLite happy and accept this syntax.

This required for MantisBT [PR 2055](https://github.com/mantisbt/mantisbt/pull/2055).

### DIFF
--- a/adodb-datadict.inc.php
+++ b/adodb-datadict.inc.php
@@ -843,7 +843,7 @@ class ADODB_DataDict {
 						$fdefault = $this->connection->qstr($fdefault);
 				}
 			}
-			$suffix = $this->_createSuffix($fname,$ftype,$fnotnull,$fdefault,$fautoinc,$fconstraint,$funsigned);
+			$suffix = $this->_CreateSuffix($fname,$ftype,$fnotnull,$fdefault,$fautoinc,$fconstraint,$funsigned,$fprimary,$pkey);
 
 			// add index creation
 			if ($widespacing) $fname = str_pad($fname,24);
@@ -897,7 +897,7 @@ class ADODB_DataDict {
 
 
 	// return string must begin with space
-	function _createSuffix($fname,&$ftype,$fnotnull,$fdefault,$fautoinc,$fconstraint,$funsigned)
+	function _CreateSuffix($fname,&$ftype,$fnotnull,$fdefault,$fautoinc,$fconstraint,$funsigned,$fprimary,&$pkey)
 	{
 		$suffix = '';
 		if (strlen($fdefault)) $suffix .= " DEFAULT $fdefault";
--- a/adodb-datadict.inc.php
+++ b/adodb-datadict.inc.php
@@ -843,7 +843,7 @@ class ADODB_DataDict {
 						$fdefault = $this->connection->qstr($fdefault);
 				}
 			}
-			$suffix = $this->_CreateSuffix($fname,$ftype,$fnotnull,$fdefault,$fautoinc,$fconstraint,$funsigned,$fprimary,$pkey);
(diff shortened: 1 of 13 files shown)


datadict/datadict-sqlite.inc.php
```diff
@@ -84,12 +84,16 @@ function actualType($meta)
 	}
 
 	// return string must begin with space
-	function _createSuffix($fname, &$ftype, $fnotnull, $fdefault, $fautoinc, $fconstraint, $funsigned)
+	function _createSuffix($fname,&$ftype,$fnotnull,$fdefault,$fautoinc,$fconstraint,$funsigned,$fprimary,&$pkey)
```
dregad, line 87: Please keep the space after comma (and if you don't mind, add it to the overrides in the inherited methods)

datadict/datadict-sqlite.inc.php
```diff
@@ -84,12 +84,16 @@ function actualType($meta)
...
 		$suffix = '';
-		if ($funsigned) $suffix .= ' UNSIGNED';
+		if ($funsigned && !($fprimary && $fautoinc)) $suffix .= ' UNSIGNED';
 		if ($fnotnull) $suffix .= ' NOT NULL';
 		if (strlen($fdefault)) $suffix .= " DEFAULT $fdefault";
+		if ($fprimary && $fautoinc) {
+			$suffix .= ' PRIMARY KEY';
+			array_pop($pkey);
+		}
 		if ($fautoinc) $suffix .= ' AUTOINCREMENT';
```
dregad, line 97: I don't think that's right, per https://www.sqlite.org/syntax/column-constraint.html `AUTOINCREMENT` is not valid on its own, must follow `PRIMARY KEY`.

Guess it should be 
```
if ($fprimary) {
	$suffix .= ' PRIMARY KEY';
	array_pop($pkey);
    if ($fautoinc) $suffix .= ' AUTOINCREMENT';
}
```

adodb-datadict.inc.php
```diff
@@ -907,7 +907,7 @@ function _getSize($ftype, $ty, $fsize, $fprec, $options=false)
 
 
 	// return string must begin with space
-	function _createSuffix($fname,&$ftype,$fnotnull,$fdefault,$fautoinc,$fconstraint,$funsigned)
```
dregad, line 910: Can you please add a PHPDoc block ?

And the `// return string must begin with space` comment can be removed from all the inherited methods.

dregad (changes requested):
Thanks for your contribution. 

I need to test this.

datadict/datadict-sqlite.inc.php
raspopov: I tried to mimic the existing format, but I also like spaces.

datadict/datadict-sqlite.inc.php
raspopov: Good point.

raspopov: It's OK to have a separate PRIMARY definition if AUTOINCREMENT is not specified.

adodb-datadict.inc.php
raspopov: Done.

Neustradamus:
@dregad: Nice to see that you have looked this PR!

I wish you a Happy New Year 2025!

Like I have said, we are in 2025, a lot of need to be merged or reviewed.
Can you look?

Thanks in advance.

dregad (approved):

PR #1089, opened 2025-08-13
Identifies the version type and uses appropriate rename syntax

mnewnham:
Improves the PSR standard code formatting


dregad:
Replaced by #1105

PR #1091, opened 2025-08-13
PSR cleanup

mnewnham:


dregad:
Hi Mark,

Please do not to mix bug fix code with code reformatting within a single commit, especially if - like here - nearly the whole file is affected (as the indent was changed from tab to space). It makes code review impossible as the significant change gets drowned in the rest, and tools like git blame difficult to use.

While I fully agree with moving code towards better PSR compliance, I believe it would be better to perform global reformatting as a coordinated, targeted effort, ideally after implementing a tool (e.g. PHP CS fixer) to automate the effort. 

I'll rework the branch to fix the above.

dregad:
Merged in 85f4495d & b6379253

PR #1094, opened 2025-08-16
Re-initialize the xmlschema sqlArray after use

mnewnham:


dregad (approved):
The only potential issue I can think of with this change, is if someone needs the SQL they would need to save it before calling executeSchema(). Not a big risk, and easy to fix so I think this is OK.

PR #1100, opened 2025-08-18
sqlite: Fix regression in SQLDate() method

dregad:
The ADOdb date/time library was removed in #970. Follow-up PR #1002 replaced legacy adodb_date()/adodb_date2() calls through custom functions, by native strftime().

That introduced a regression in SQLDate() behavior, as the date formats changed from PHP-style to strftime (e.g. `Ymd` -> `%Y%m%d`).

This PR partially reverts 4700171afa6c5f939e8d04b079331e207dea22ca and reintroduces the createFunction() call with a closure providing behavior compatible with the legacy adodb_date2() function. This makes the __createFunctions()_ internal method unnecessary.

It also removes the now-useless test script that was added as part of PR #1002.

@joebordes this is the follow-up on https://github.com/ADOdb/ADOdb/pull/1002#issuecomment-3096159895; I'll wait for a few days before merging, to give you a chance to review and comment.


joebordes:
LGTM

dregad:
Thanks for the feedback !

Issue #1103, opened 2025-08-20
[mysqli] offsetDate() does not return correctly formatted and correctly calculated dates


mnewnham:
### Description

- The method should return an SQL statement that generates an ISO date format string. The method actually returns the correct date but with a trailing '00:00:00' time portion
- The method does not correctly generate a statement if the hours (/24) format is used
- The method does not correctly generate a statement if the increment is negative, because the MySQL command does not accept a + INTERVAL -value construct
- The method does not correctly generate a statement if the base date is an ISO date string

### Environment
- ADOdb version: _master_
- Driver or Module: adodb-mysqli.inc
- Database type and version: MySQL 8
- PHP version: 8.0.30
- Platform: Windows

* [X] I have tested that the problem is reproducible in the [latest release](https://github.com/ADOdb/ADOdb/releases/latest) / [master](https://github.com/ADOdb/ADOdb/tree/master) branch / latest [hotfix](https://github.com/ADOdb/ADOdb/branches/all?query=hotfix) branch

dregad:
If we set backwards-compatibility aside, does it make sense to allow $dayFraction to be a string ? I mean, the caller could simply perform the calculation, pass in a float and the code could be greatly simplified (no need for parsing with preg_match, just a < 0 check for negative value, etc).

e.g. `$db->offsetDate($hours / 24)` instead of `$db->offsetDate($hours . '/24')`

What do you think ?

mnewnham:
> If we set backwards-compatibility aside, does it make sense to allow $dayFraction to be a string ? I mean, the caller could simply perform the calculation, pass in a float and the code could be greatly simplified (no need for parsing with preg_match, just a < 0 check for negative value, etc).
> 
> e.g. `$db->offsetDate($hours / 24)` instead of `$db->offsetDate($hours . '/24')`
> 
> What do you think ?

To be honest, I didn't really understand the point of doing the string thing, but then I realized that Oracle can handle it naturally.
It was purely the compatibility to the already defined documentation and oracle that made me put complex code in there. We could however grab that code and make it a common function for all the drivers. I do think its a good idea to make it go away though, its just too wierd

I feel that a better way would be to go with a new method like

````php
$date = offsetInterval($baseDate,$interval,$intervalType);

$date = offsetInterval('2025-08-01,5,'weeks'); // 2025-08-01 + 5 weeks = 2025-09-05
````

dregad:
> compatibility to the already defined documentation

[...]

PR #1104, opened 2025-08-20
Redo offsetDate() method to cover various scenarios

mnewnham:


dregad:
See https://github.com/ADOdb/ADOdb/issues/1103#issuecomment-3207108052

PR #1099, opened 2025-08-18
MySQL now uses standard SQL syntax for renaming columns

mnewnham:

PR #1105, opened 2025-08-21
Use old ALTER COLUMN syntax for MySQL < 8.0

### LINKED ISSUE #1088: [mysqli] MySQL supports rename column syntax for version 8.0 and higher
MySQL supports the _**RENAME COLUMN**_ syntax for renaming a column in a table if the database is version 8.0 or higher. The data dictionary _**renameColumnSql()**_ method now checks the version and uses that syntax if appropriate. It is now no longer necessary to provide the column format for column renames in MySQL.

dregad:
This is a follow-up on PR #1099, fixes #1088.

It replaces PR #1089, with the following modifications :

- Removed indentation changes
- Moved the PHPDoc blocks from ADODB2_mysql to ADODB_DataDict.
- Use version_compare() instead of splitting the version number to compare major version

### DIFF
--- a/adodb-datadict.inc.php
+++ b/adodb-datadict.inc.php
@@ -445,12 +445,11 @@ class ADODB_DataDict {
 	{
 		$meta = strtoupper($meta);
 
-		/*
-		* Add support for custom meta types. We do this
-		* first, that allows us to override existing types
-		*/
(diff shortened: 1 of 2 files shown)


mnewnham (approved):
This passes all the tests I have

dregad:
Thanks for the feedback

PR #1121, opened 2025-09-12
Changes to make mysqli defaults work correctly

mnewnham:


dregad (changes requested):
Change is looking good, but shouldn't it be applied to the PDO mysqli driver as well? 

https://github.com/ADOdb/ADOdb/blob/712d28c430f62135dccd0a4db5a828177af07b32/drivers/adodb-pdo_mysql.inc.php#L30-L31

mnewnham:
> Change is looking good, but shouldn't it be applied to the PDO mysqli driver as well?
> 
> https://github.com/ADOdb/ADOdb/blob/712d28c430f62135dccd0a4db5a828177af07b32/drivers/adodb-pdo_mysql.inc.php#L30-L31

Agreed, but there are so many additional problems in the PDO MySQL driver (and all of the other PDO drivers) that I'd prefer to wait until we agree on a release mechanism for the remodeled PDO driver so I can use the unittest to analyze the problems. The refactored version is, after all close to 3 years without any movement forward

dregad:
Fair point.

I do realize that the PDO mysql driver currently does not support Data Dictionary so changing those properties now would not make an actual difference with regards to this issue, but doing it while it's hot has the advantage of making sure we don't forget about it when comes the time to merge #826 (which I believe has the latest version of your PDO refactoring).

So if that's OK with you I'll add a commit with the change and merge this PR, then cherry-pick the on the PDO branch.

dregad:
Merged in eac4213bc2be00fedde38510c31b370f95b52498

dregad:
The fix for PDO mysql driver was added to the *pdo-preload-demo* branch (3f0fa0dfbedb25603eee0c290fcc489fd6e48704).

PR #1123, opened 2025-09-15
PDO Firebird driver extends ADODB_pdo_base

### LINKED ISSUE #1122: Firebird driver pdo_firebird missing _init() function
### Description
I tried to use 

### Environment
- ADOdb version: 5.22.10
- Driver or Module: pdo_firebird
- Database type and version: Firebird (any)
- PHP version: 8.4
- Platform: Rocky 8

* [X ] I have tested that the problem is reproducible in the 5.22.10 version downloaded from adodb.org website  

File:
ADOdb-5.22.10/drivers/adodb-pdo_firebird.inc.php

### Steps to reproduce
Always fail connection attemp with error:

Fatal error: Uncaught Error: Call to undefined method ADODB_pdo_firebird::_init()

### Expected behavior
Succesful connection

### Additional context 
There si no _init() function in file:

ADOdb-5.22.10/drivers/adodb-pdo_firebird.inc.php

If You add this in line 34 it works:

function _init($parentDriver)
{
}

Is it missing for any reason? Or  is it a bug?

dregad:
Previously, it inherited from ADODB_pdo, which caused Fatal error: Uncaught Error: Call to undefined method ADODB_pdo_firebird::_init().

Fixes #1122

### DIFF
--- a/drivers/adodb-pdo_firebird.inc.php
+++ b/drivers/adodb-pdo_firebird.inc.php
@@ -24,7 +24,7 @@
 /**
  * Class ADODB_pdo_firebird
  */
-class ADODB_pdo_firebird extends ADODB_pdo
+class ADODB_pdo_firebird extends ADODB_pdo_base
 {
 	public $dialect = 3;
 	public $metaTablesSQL = "select lower(rdb\$relation_name) from rdb\$relations where rdb\$relation_name not like 'RDB\$%'";


dregad:
I'm planning to apply this change to other PDO drivers as well for consistency, but to avoid risk of regression in stable branch that will be targeted at master (5.23.0).

PR #1135, opened 2025-10-07
Split BOF and EOF functionality because featues are incompatible

mnewnham:


mnewnham:
@dregad, could you take a look at this and give me your opinion. It's not quite cleaned up yet but it works properly. The problem seems to be that the same EOF flag and _currentRow values are being used to handle errors at both the beginning and end of files. This means that the values are trying to manage situations where changes to the row number trigger an error both going forward and going backwards. This causes situations where the move() return value is incorrect and the EOF flag is switched off even though the recordset is still in EOF state

There's on odd piece of code which rolls back the currentRow by 2 if the pointer goes beyond EOF. 

````php
if ($rowNumber >= $this->_numOfRows) {
	if ($this->_numOfRows != -1) {
		$rowNumber = $this->_numOfRows-2;
	}
````
I can't work out what this is supposed to be doing, possibly related to using move() with fetchRow() which advances the recordset pointer. We don't have an array based equivalent of fetchObj() which reads the record but doesn't advance the record pointer.

I've handled this by holding the state of the recordset when at the beginning of the file in a new BOF flag. The only problem going forward would seem to be an edge-case situation where a user is looking directly at the EOF flag to manage an action where the recordset is going negative. They would have to change their code to look at BOF instead.

There is now a fairly substantial test, MoveTest() in the unit testing repo if you need it

dregad:
Following up on https://github.com/ADOdb/ADOdb/pull/1135#discussion_r2748739058

> I guess I'm not understanding your approach here. All of the EOF, BOF and _currentRow values are not just set but we need to know the values of the parameters as the method starts up. They also need to be set properly whenever the method exits. So move() needs to know for example whether it is already at BOF as the method starts up.

My apologies, re-reading myself I realize that my review comments were not very clear.

What I meant is that there is no need to store the current values of BOF, EOF and _currentRow, change them and then reset them if we're already at BOF, EOF or did not move the pointer. We can perform that initial check on the recordset's current values, return if needed and only then change the values. This simplifies the code and avoids unnecessary assignments.

I pushed a branch with my proposed change, which should make it easier for you to understand. [Click here](https://github.com/ADOdb/ADOdb/compare/1131-core-move-method-incorrectly-sets-eof-flag-when-advanced-multiple-times-past-eof...dregad:1131-pr-review) to see the full difference (only commit 593b219d1d8da47ae200559e4e043c5f6d68842b is truly relevant, the other two are just code cleanup).

The unit tests pass on mysqli (assuming I ran them correctly - had some difficulty setting them up).

mnewnham:
I looked at yours and ran it through my unit test and its fine. I'm afraid I can't really express any opinions of one over the other. It's just a difference in coding styles to me. One thing we should clarify is what the value of `currentRow` is when either `BOF` or `EOF` is set. The doc says that it's 0 for both and I think that is wrong. I think it should specifically be `false`. I'm happy for you to pull your own version to get it into master.

dregad:
> One thing we should clarify is what the value of currentRow is when either BOF or EOF is set. The doc says that it's 0 for both and I think that is wrong. I think it should specifically be false.

I agree with you. 0 would be confusing and even incorrect since the offsets are 0-based.

That being said, the docs are correct AFAICT (assuming you mean https://adodb.org/dokuwiki/doku.php?id=v5:reference:recordset:move), stating
* If the move is unsuccessful, (-ve offset, passed EOF etc) the function returns false. *

I'll merge the fix shortly.

dregad:
Merged in 4ff1ec4ad6fba4681750d14a823834eda3f7f374

Issue #1132, opened 2025-10-05
Refactor ADOConnection::metaTables()


dregad:
Code currently relies on nested for loops to filter the results. Use of array_filter() would be more efficient and easier to read.

dregad:
Commit 555555be5029980446a64bc597dafc28ae447923 introduced a regression.

In several drivers the _$metaTablesSql_ statement only returns a single column with the table name. If the driver does not override the metaTables() method (e.g. sqlite, ibase), ADOdb throws a PHP warning

Warning: Undefined array key 1 in .../adodb.inc.php on line 3284

This is probably the reason why #1149 was submitted.

PR #1149, opened 2026-01-14
Fix metaTables() for SQLite3

raspopov:
The SELECT call used in the metaTables() function must return a dataset with two columns: the name of the table or view and the type 'T' or 'V'.

This PR is needed for https://github.com/mantisbt/mantisbt/pull/2055

### DIFF
--- a/drivers/adodb-pdo_sqlite.inc.php
+++ b/drivers/adodb-pdo_sqlite.inc.php
@@ -23,7 +23,7 @@
 
 class ADODB_pdo_sqlite extends ADODB_pdo_base {
 	/** @noinspection SqlResolve */
-	var $metaTablesSQL   = "SELECT name FROM sqlite_master WHERE type='table'";
+	var $metaTablesSQL   = "SELECT name,CASE WHEN type='table' THEN 'T' ELSE 'V' END FROM sqlite_master WHERE type IN('table','view') AND name NOT LIKE 'sqlite_%'";
 	var $sysDate         = 'current_date';
 	var $sysTimeStamp    = 'current_timestamp';
 	var $nameQuote       = '`';
--- a/drivers/adodb-sqlite.inc.php
+++ b/drivers/adodb-sqlite.inc.php
@@ -35,7 +35,7 @@ class ADODB_sqlite extends ADOConnection {
 	var $hasLimit = true;
 	var $hasInsertID = true; 		/// supports autoincrement ID?
 	var $hasAffectedRows = true; 	/// supports affected rows for update/delete?
-	var $metaTablesSQL = "SELECT name FROM sqlite_master WHERE type='table' ORDER BY name";
+	var $metaTablesSQL = "SELECT name,CASE WHEN type='table' THEN 'T' ELSE 'V' END FROM sqlite_master WHERE type IN('table','view') AND name NOT LIKE 'sqlite_%'";
 	var $sysDate = "date('Y-m-d')";
 	var $sysTimeStamp = "date('Y-m-d H:i:s')";
 	var $fmtTimeStamp = "'Y-m-d H:i:s'";
--- a/drivers/adodb-sqlite3.inc.php
+++ b/drivers/adodb-sqlite3.inc.php
@@ -38,7 +38,7 @@ class ADODB_sqlite3 extends ADOConnection {
 	var $hasLimit = true;
 	var $hasInsertID = true; 		/// supports autoincrement ID?
 	var $hasAffectedRows = true; 	/// supports affected rows for update/delete?
-	var $metaTablesSQL = "SELECT name FROM sqlite_master WHERE type='table' ORDER BY name";
+	var $metaTablesSQL = "SELECT name,CASE WHEN type='table' THEN 'T' ELSE 'V' END FROM sqlite_master WHERE type IN('table','view') AND name NOT LIKE 'sqlite_%'";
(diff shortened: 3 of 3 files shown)


dregad:
@raspopov thanks for your contribution.

I believe you were hit by the regression described at https://github.com/ADOdb/ADOdb/issues/1132#issuecomment-3806042942. I just pushed a fix for that to master, let me know if it fixes your problem.

That being said, your PR adds support for $ttype parameter which the sqlite3 driver does not have today, so it's still worth merging. 

@mnewnham I noticed you are planning to improve the sqlite driver (#1158). Since you mentioned adding support for views, this PR might conflict with your work - let me know if you want me to merge this or if you prefer including the change yourself.

raspopov:
> I just pushed a fix for that to master, let me know if it fixes your problem.

Yes this fixes the error from the other side.

dregad:
Follow up on [above note](https://github.com/ADOdb/ADOdb/pull/1149?notification_referrer_id=NT_kwHOAAbdY9oAJFJlcG9zaXRvcnk7MTE5ODc3MjE7SXNzdWU7MzgxNDY3NzAxNA#discussion_r2733234275)
>> > `AND (name NOT LIKE 'sqlite_%') `
>> 
>> Is this condition really needed ?
>
>Yes, I see two system tables `sqlite_master` and `sqlite_sequence` in 3.50.4.

Very weird. I tried on several machines with sqlite versions 3.37.2, 3.48.0 and 3.51.2 on windows, ubuntu and macos, but the *sqlite_master* table is **never** listed (which is as it should be according to the docs). No idea why you are seeing it...

The *sqlite_sequence* table is indeed visible, if it exists (i.e. when the schema contains a table with an autoincrement column), as could *sqlite_statX* and *sqlite_autoindex_XXX*.

So I guess excluding _internal schema objects_ does not hurt, considering that [SQLite prohibits applications from creating objects whose names begin with "sqlite_"](https://sqlite.org/fileformat.html#internal_schema_objects).

dregad (approved):
OK for me. Waiting for @mnewnham's feedback before merging.

raspopov:
> Very weird. I tried on several machines with sqlite versions 3.37.2, 3.48.0 and 3.51.2 on windows, ubuntu and macos, but the _sqlite_master_ table is **never** listed

Perhaps I didn't express myself clearly. `sqlite_master` does not appear in the query results, but `sqlite_sequence` does. However, I can see both tables when viewing the database with a third-party utility. In any case, we need to exclude system objects from the query.

mnewnham:
There are unfortunately some limitations when trying to use `ADOConnection::MetaTables()` + `driver::MetaTablesSQL`, which is giving you lots of problems here.

The biggest problem is that the core method doesn't support `$mask` which is giving false responses when passed a value.

In addition, because SQLite supports views, the `$ttype` parameter needs to be accessed to filter the results. The core method provides no mechanism for injecting that into the query.

per @dregad , I've also never seen sqlite* appearing in either physical or memory  based databases.

dregad:
Thanks for the feedback @mnewnham. 

Indeed this PR does not make ADODB_sqlite3::metaTables() full-featured, but it's a step in the right direction.

Not sure what you mean by this though
> per @dregad , I've also never seen sqlite* appearing in either physical or memory based databases.

And you did not answer my questions, i.e. is are you OK with me merging this ?

mnewnham:
> Thanks for the feedback @mnewnham.
> 
> Indeed this PR does not make ADODB_sqlite3::metaTables() full-featured, but it's a step in the right direction.
> 
> Not sure what you mean by this though
> 
> > per @dregad , I've also never seen sqlite* appearing in either physical or memory based databases.
> 
> And you did not answer my questions, i.e. is are you OK with me merging this ?

Sorry, I thought my comments would indicate my rejection. Here's one that does work, if the code is needed quickly https://github.com/ADOdb/ADOdb/commit/abfc0b38c91b6f9f1e0d2cd2f88b447b2ed87a0a

raspopov:
@mnewnham, The latest "ADOdb:master(https://github.com/ADOdb/ADOdb/commit/f391fb97f3cb6620d918cda5fb2c5262e548f07a)" with PHP 7.4 and mysqli or pgsql or sqlite3 produces:
```
PHPUnit 9.6.34 by Sebastian Bergmann and contributors.

...............................................................  63 / 493 ( 12%)
............................................................... 126 / 493 ( 25%)
............................................................... 189 / 493 ( 38%)
......................E........................................ 252 / 493 ( 51%)
............................................................... 315 / 493 ( 63%)
............................................................... 378 / 493 ( 76%)
............................................................... 441 / 493 ( 89%)
....................................................            493 / 493 (100%)

Time: 00:36.498, Memory: 54.00 MB

There was 1 error:

1) Mantis\tests\soap\AttachmentTest::testProjectAttachmentIsAdded
SoapFault: Return value of ADOConnection::MetaColumnNames() must be an instance of mixed, array returned

/home/runner/work/mantisbt/mantisbt/tests/soap/AttachmentTest.php:190
phpvfscomposer:///home/runner/work/mantisbt/mantisbt/vendor/phpunit/phpunit/phpunit:106

ERRORS!
Tests: 493, Assertions: 1888, Errors: 1.
```
PHP 8 runs fine.

dregad:
@raspopov thanks for the report, the regression was introduced in 1297d4badaef465c3219f9325ea83bfa3865c398, `mixed` type is only available on PHP 8.0+.

Tracking in #1210

PR #1163, opened 2026-02-01
Change structure of metadatabase record iteration to not match associative keys

mnewnham:


dregad:
@mnewnham looking at this, I'm wondering why we even need to override metaDatabases() in the mysqli driver...

Wouldn't it be simpler if we simply set this in the mysql driver
```php
	var $metaDatabasesSQL = "SELECT schema_name FROM information_schema.schemata
		WHERE schema_name NOT IN ('mysql', 'information_schema', 'performance_schema')
		ORDER BY 1";
```
or if you prefer to use SHOW DATABASES
```php
	var $metaDatabasesSQL = "SHOW DATABASES 
            WHERE `database` NOT IN ('mysql', 'information_schema', 'performance_schema');";
```

then we can just let ADOConnection::metaDatabases() do the job...

What do you think ?

mnewnham:
Looks good, no issue with your suggested change

PR #1165, opened 2026-02-02
Consolidate metaType methods to single reference method

mnewnham:


adodb.inc.php
```diff
@@ -538,6 +538,14 @@ abstract class ADOConnection {
 	 */
 	public $debug = false;
 
+	/**
+	 * A placeholder for a data dictionary object, created at
+	 * first use of newDataDictionary and available as needed.
+	 *
+	 * @var object|null
+	 */
+	public ?object $dataDictionary = null;
```
dregad: Avoid using generic *object* type, better use the actual class when possible. Here it should be `ADODB_DataDict` this allows more accurate static analysis and smart IDEs will work better.

datadict/datadict-mysql.inc.php
```diff
@@ -33,88 +33,165 @@ class ADODB2_mysql extends ADODB_DataDict {
...
+	/**
+	 * Returns the meta type for a given type and length.
+	 *
+	 * @param mixed  $t        The object to test.
+	 * @param int    $len      The length of the field, if applicable.
+	 * @param object $fieldobj The field object, if available.
+	 *
+	 * @return string
+	 */
+	public function metaType($t, $len = -1, $fieldobj = false)
```
dregad: When I read the problem description in the attached issue, I was under the impression that you wanted to change the method signature ?

$t (and $fieldobj) should be an *ADOFieldObject* type (at least in PHPDoc declaration).

dregad (commented):
I did not actually test the code.

I'm all in favour of such refactoring, as I always found the situation with metaTypes confusing.

Considering that the dictionary already has a connection object, wouldn't it be better to move the canonical metaType function to ADOConnection, instead of adding a datadict to it ?

mnewnham:
I have additional motives here. I have been prototyping additional feature to https://github.com/ADOdb/ADOdb/tree/pdo-preload-demo. What we are able to do there, with some simple code updates, is to use the data dictionary functions from the native mode drivers to serve the data dictionary functionality to PDO drivers, e.g. in ``pdo-mysql``,   a class property ``$dictionaryProvider='mysql'`` causes the newDataDictionary() to load `datadict-mysql` as the dictionary. We can then eliminate all of the low quality datadict-pdo-* includes

dregad:
Ah OK so there would be a single datadict to maintain for a given database type, common to both the native and the PDO driver ? That makes sense

mnewnham:
That was my plan. I did a POC with some weak namespacing here. https://github.com/ADOdb/ADOdb/tree/pdo-preload-plus-namespaced-datadict. It only works for pdo-mysql but if you connect with that you can use the datadict-mysqli as a dictionary provider

dregad:
Looks like you did not address my code review comments above.

mnewnham:
> Looks like you did not address my code review comments above.

I really wanted to know if you thought it was a concept worth exploring in which case I'll go a bit deeper (such as changing the signature).  An additional thing might be extracting meta Functions so that they can be shared with pdo drivers. I realize that some might have PHP driver specific code but many just use straight SQL to return meta data.

dregad:
Sorry, I though I gave my high-level opinion in https://github.com/ADOdb/ADOdb/pull/1165#pullrequestreview-3769994351

> I'm all in favour of such refactoring, as I always found the situation with metaTypes confusing.

and validating the approach to move the code in datadict (https://github.com/ADOdb/ADOdb/pull/1165#issuecomment-3875861270)

> there would be a single datadict to maintain for a given database type, common to both the native and the PDO driver ? That makes sense

So by all means do carry on 😄 

BTW, for those explorative PRs, you may want to mark them as *draft* when you create them, so it's obvious that it's a work-in-progress. It's also possible to change the status after the fact
https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/proposing-changes-to-your-work-with-pull-requests/changing-the-stage-of-a-pull-request

mnewnham:
> BTW, for those explorative PRs, you may want to mark them as _draft_ when you create them, so it's obvious that it's a work-in-progress. It's also possible to change the status after the fact https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/proposing-changes-to-your-work-with-pull-requests/changing-the-stage-of-a-pull-request

I will certainly do this in future. The "can't merge to master" thing isn't working as well as I hoped. I don't think the granularity I wanted is there because it means I can't even merge my own pull requests. There are certain trivialities that I don't really think I want a code review on and I just want to pull them into master to clear them. I think if you can't configure it so that I cant commit to master but can pull my own requests then I think I'll go back to the original config but try to take more care on merging

dregad:
Follow up on this in #1211

mnewnham:
I have done some further work on this and there are some problems with my plan for consolidation across drivers. The core problem is the way that the `ADOrecordset` is instantiated within the connection, for example calling a function from the PDO mysql driver to access the mysqli dictionary would cause the ADOrecordset to be generated as an `ADORecordset_mysqli` object instead of an `ADORecordset_pdo_mysql` object, which would mean that the methods inside would be wrong.

I have, however worked up a [branch ](https://github.com/ADOdb/ADOdb/tree/Full-autoload-refactor) that has a complete restructuring of the classes into an autoload style of delivery which 
 - Demonstrates how a reworking could appear
 - Demonstrates how the native mode driver could provide services to the PDO driver

This demonstration does actually work for MySQL and PDO MySQL using the code from the PDO Preload branch. The approximate tree looks like this. This form provides the resource sharing I am looking for.

````
- adodb.inc.php
+- Resources    +
                + MySQL ------------+
                |                   +--- ADOConnection.php
                |                   +--- ADORecordset.php
                |                   +--- DataDictionary.php
                |                   +--- MetaFunctions.php
                + SQLite ------> 
                + Oracle ------> 
                + SQLServer ---> 
                |                   
                + PDO --------------+
                |                   +--- ADOConnection.php
                |                   +--- ADORecordSet.php
                |                   +--- DataDictionary.php
                |                   +--- MetaFunctions.php  
                |                   |
                |                   +--- MySQL ----------------+
                |                                               + --- ADOConnection.php
                |                                               + --- ADORecordSet.php
                + ActiveRecord ------+
                |
                +--- ADOConnection.php
                +--- ADORecordSet.php
                +--- ADORecordSetArray.php
                +--- ADORecordSetEmpty.php`
                +--- DataDictionary.php
                +--- MetaFunctions.php 
````
 Using this evaluation, I think it would be better to retreat a little from this plan and approach it a little simpler.  
 - The recordset **metaType()** should be split into driver specific entries
 - Copy and paste those into the datadict() entries 
I think that should be the limit of the metaType() fix

PR #1168, opened 2026-02-08
mysqli: use ADOConnection::metaDatabases()

### LINKED ISSUE #1162: [mysqli] Metadatabases() returns empty result set if fetch mode is ADODB_FETCH_NUM
If the current result set option is `ADODB_FETCH_NUM` and a call is made to `metaDatabases()`, an empty result set is returned because the method attempts to match values using associative keys.

The method also returns system schema because the filtering is outdated and does not recognize `information_schema` and `performance_schema` as system databases.

dregad:
As explained in https://github.com/ADOdb/ADOdb/pull/1163#issuecomment-3867039767.

Define $metaDatabasesSQL in the driver-specific class and remove the ADODB_mysqli::metaDatabases() method override, let the parent do the job.

Also define the meta query for the PDO mysql driver.

Fixes #1162

### DIFF
--- a/drivers/adodb-mysqli.inc.php
+++ b/drivers/adodb-mysqli.inc.php
@@ -47,6 +47,8 @@ class ADODB_mysqli extends ADOConnection {
 	var $dataProvider = 'mysql';
 	var $hasInsertID = true;
 	var $hasAffectedRows = true;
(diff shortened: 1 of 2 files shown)


dregad:
Approved by Mark, see https://github.com/ADOdb/ADOdb/pull/1163#issuecomment-3873806339

PR #1170, opened 2026-02-09
Fixes invalid array key error

mnewnham:


mnewnham (commented):
Fixes an invalid array key adding an new column with an array

datadict/datadict-postgres.inc.php
```diff
@@ -232,7 +232,11 @@ function alterColumnSQL($tabname, $flds, $tableflds='', $tableoptions='')
 					list(,$colname,$default) = $matches;
 					$alter .= $colname;
 					if ($this->connection) {
-						$old_coltype = $this->connection->metaType($existing[strtoupper($colname)]);
+						if (array_key_exists(strtoupper($colname), $existing)) {
+							$old_coltype = $this->connection->metaType($existing[strtoupper($colname)]);
+						} else {
+							$old_coltype = $t;
```
dregad: `$t` variable is undefined.

This is a pre-existing problem (see line 241); it's only initialized later in the method (line 244), so $old_coltype will effectively be null. I did not analyze the consequences.

PR #1175, opened 2026-02-14
Change the output of MetaForeignKeys

mnewnham:
- Remove the PHP8.5 deprecation warning
 - Add support for forceAssociative and Uppercase options
 - Ensure that the returned format conforms to the documented standard

### DIFF
--- a/drivers/adodb-oci8.inc.php
+++ b/drivers/adodb-oci8.inc.php
@@ -1549,7 +1549,7 @@ SELECT /*+ RULE */ distinct b.column_name
 	 * @param bool   $upper       Return keys in uppercase (true)
 	 * @param bool   $associative Force associative mode
 	 *
-	 * @return array|false An array where keys are tables, and values are foreign keys;
+	 * @return string[]|false An array where keys are tables, and values are foreign keys;
 	 *                     false if no foreign keys could be found.
 	 */
 	public function metaForeignKeys($table, $owner = '', $upper = false, $associative = false)


drivers/adodb-oci8.inc.php
```diff
@@ -1536,50 +1536,102 @@ function MetaPrimaryKeys($table, $owner=false,$internalKey=false)
...
-	 * @param bool   $upper       discarded
-	 * @param bool   $associative discarded
+	 * @param string $table	      Name of the table
+	 * @param string $owner		  Owner of the table
+	 * @param bool   $upper       Return keys in uppercase (true)
+	 * @param bool   $associative Force associative mode
 	 *
-	 * @return string[]|false An array where keys are tables, and values are foreign keys;
-	 *                        false if no foreign keys could be found.
+	 * @return array|false An array where keys are tables, and values are foreign keys;
```
dregad: Why change this ? *string[]* is perfectly [valid PHPDoc syntax](https://docs.phpdoc.org/guide/references/phpdoc/types.html)... using *array* 
- makes static analysis less accurate 
- diverges from the parent method

dregad (commented):
I do not have a working Oracle setup to actually test the code, but it looks good to me except IMO unnecessary change in PHPDoc return type.

drivers/adodb-oci8.inc.php
dregad: > as string[] isn't (I think) a valid return type 

While it's true that PHP only supports plain *array*, as mentioned above the *<type>[]* syntax is actually a valid PHPDoc thing that supersedes and enhances the basic *array* type (follow the link in my earlier post). 

Using it provides more accurate documentation which is the whole point of PHPDoc, and IDEs will pick up on that and flag type mismatch more accurately, for example in PHPStorm: 

image

dregad:
In the interest of keeping the history as linear as possible (which facilitates future uses of git blame, bisect, reverts, etc), unless it's a long-running branch with complex conflicts to resolve, it would be preferable if you could *rebase* the feature branch on top of master before merging it, instead of merging master into the branch before merging it back.

mnewnham:
no problem, just a habit I'd got into

dregad:
BTW in case you did not know, if there are no merge conflicts, the GitHub UI lets you easily rebase the branch with just 2 clicks, from the box at the bottom of the PR's page: 
image

PR #1186, opened 2026-02-16
Correctly identify foreign keys by reading information_schema database.

mnewnham:
- Replace regexp parsing of the table "CREATE TABLE" statement.
 - Fix casing issues
 - Fix transposed column names fro target to source tables

### DIFF
--- a/drivers/adodb-mysqli.inc.php
+++ b/drivers/adodb-mysqli.inc.php
@@ -940,6 +940,10 @@ class ADODB_mysqli extends ADOConnection {
 			*/
 			$alternateDb = array_map('strtolower', $this->MetaDatabases());
 			if (!in_array(strtolower($owner), $alternateDb)) {
+				$ADODB_FETCH_MODE = $saveModes[0];
+				if ($saveModes[1] !== false) {
+					$this->SetFetchMode($saveModes[1]);
+				}
 				return false;
 			}
 			$referenceSchema = $owner;


drivers/adodb-mysqli.inc.php
```diff
@@ -918,71 +918,110 @@ public function metaForeignKeys($table, $owner = '', $upper = false, $associativ
...
 
 		if ($ADODB_FETCH_MODE == ADODB_FETCH_ASSOC
-		|| $this->fetchMode == ADODB_FETCH_ASSOC)
+		|| $this->fetchMode == ADODB_FETCH_ASSOC) {
 			$associative = true;
+		}
+		$saveModes = [
+			$ADODB_FETCH_MODE,
+			$this->fetchMode
+		];
```
dregad: Unless I'm missing something, I don't believe we need to save both the global and the fetchMode property, since setFetchMode() will return the property's old value and take the global into consideration, it should be sufficient to do 

```php
$savem = $this->setFetchMode(ADODB_FETCH_ASSOC);
```

drivers/adodb-mysqli.inc.php
```diff
@@ -918,71 +918,110 @@ public function metaForeignKeys($table, $owner = '', $upper = false, $associativ
...
-		$foreign_keys = array();
-		$num_keys = count($matches[0]);
-		for ( $i = 0; $i < $num_keys; $i ++ ) {
-			$my_field  = explode('`, `', $matches[1][$i]);
-			$ref_table = $matches[2][$i];
-			$ref_field = explode('`, `', $matches[3][$i]);
+		$ADODB_FETCH_MODE = $saveModes[0];
+        if ($saveModes[1] !== false) {
+            $this->SetFetchMode($saveModes[1]);
+        }
```
dregad: $this->setFetchMode($savem);

dregad:
Other than that, the proposed change looks good.

I'm wondering if the logic to set $associative variable is correct, shouldn't it first check fetchMode (if not false) and then fallback to $ADODB_FETCH_MODE ?

Or just compare with the return value from setFetchMode() per my review comment ?

mnewnham:
I think we do have a problem here with ADODB_FETCH_MODE, I think that if a user **_never_** uses setFetchMode() then  introducing its use deep inside the code will cause a cascading change which means that setting $ADODB_FETCH_MODE will no longer work in the program because setFetchMode() captures it. This is, I think, how it appears in my tests

drivers/adodb-mysqli.inc.php
```diff
@@ -918,71 +918,110 @@ public function metaForeignKeys($table, $owner = '', $upper = false, $associativ
...
 			/*
-			* Invalid table or owner provided
+			* MySQL is using this for schema, so call metaDatabases to see if
+			* the schema exists
 			*/
-			$this->setFetchMode($savem);
-			return false;
+			$alternateDb = array_map('strtolower', $this->MetaDatabases());
+			if (!in_array(strtolower($owner), $alternateDb)) {
+				return false;
```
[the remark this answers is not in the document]
mnewnham: fixed, but I've left the double restore in there pending deeper analysis

dregad:
I'm not sure what kind of cascading effect you are worried about... This is the current situation (several variants of this code exist throughout the code base):

```php
$savem = $ADODB_FETCH_MODE;
$ADODB_FETCH_MODE = ADODB_FETCH_NUM;
$this->execute($sql);
$ADODB_FETCH_MODE = $savem;

# or the belt and braces version

$save = $ADODB_FETCH_MODE;
$ADODB_FETCH_MODE = ADODB_FETCH_NUM;
if ($this->fetchMode !== false) {
	$savem = $this->SetFetchMode(false);
}
$this->execute($sql);
if (isset($savem)) $this->SetFetchMode($savem);
$ADODB_FETCH_MODE = $save;
```

which is not really different from

```php
$savem = $this->setFetchMode(ADODB_FETCH_NUM);
$this->execute($sql);
$this->setFetchMode($savem);
```

In both cases, fetch mode is changed just for query execution and the initial state restored afterwards, so there should be no impact on the user's environment.. 

Looking at setFetchMode() method, I don't see it "capturing"  $ADODB_FETCH_MODE, it just returns it when fetchMode property is false.

```php
function SetFetchMode($mode) {
$old = $this->fetchMode;
$this->fetchMode = $mode;

if ($old === false) {
	global $ADODB_FETCH_MODE;
	return $ADODB_FETCH_MODE;
}
return $old;
}
```

mnewnham:
Here is a little sample program and the result. What it shows is that setting and using ADODB_FETCH_MODE as the global is possible, and it can continue to be used _**as long as the $db->setFetchMode() method is never used**_. As soon as it is, then the global value is no longer usable. Using database querying functions is fine because that checks each time to see if fetchMode has been set.

Our problem occurs when we use the metaFunctions. The ones that I am looking at arbitrarily call `setfetchMode()` without checking the initial status. this means that if you call for example metaColumns() in the middle of a procedure, subsequent changes to ADODB_FETCH_MODE will no longer have any effect (without warning).  This is what I am trying to address with these changes, basically if fetchMode is false, ensure that it is reset back to false separately from ADODB_FETCH_MODE. 

Many dictionary methods already do this already, but in a slightly different way
````php
$save = $ADODB_FETCH_MODE;
$ADODB_FETCH_MODE = ADODB_FETCH_NUM;
if ($this->fetchMode !== FALSE) {
	$savem = $this->setFetchMode(FALSE);
}

code ........

// restore fetchmode
if (isset($savem)) {
	$this->setFetchMode($savem);
}
$ADODB_FETCH_MODE = $save;
````
The minor problem I have with this is that

1. It relies on the **_existence_** of a variable as a test. This is just a coding style I've never cared for as the variables could easily be set in inline code
2. There is a possibility that a call to a sub-procedure changes the value of the class property `$fetchMode`, in which case it would never get reset if it was not set already when we entered the above function. This would then break the returned $ADODB_FETCH_MODE

My approach is a bit stabler
````php
$fetchModes = [
   $ADODB_FETCH_MODE,
   $this->fetchMode
];

code .....

$ADODB_FETCH_MODE = $fetchModes[0];
$this->fetchMode = $fetchModes[1];
````
This ensures that the modes are reset accurately every time

That is not to say that this is the correct approach. I don't particularly like use of a global to control this, and I have not used it in production code. I've always used setFetchMode(). I think that this should be controlled at a connection level so one connection could be NUM while the other is ASSOC.

The choice is therefore, either check and set the 2 settings throughout the dictionary wherever we force the fetch mode inside the dictionary and ensure that the 2 values are reset individually or capture the global at start up and ignore it from then on.

````php
include 'ADOdb/adodb.inc.php';
$db = newAdoConnection('sqlite3');
$db->connect(':memory:');
$db->execute('create table a (id integer primary key autoincrement, c varchar(10))');
$db->execute("insert into a (c) values ('test')");
print "\n --------------default ----------\n";
$saveFetchModes = [$ADODB_FETCH_MODE, $db->fetchMode ];
print_r($saveFetchModes);
print_r($db->getAll('select * from a'));
print "\n --------------set global to ASSOC ----------\n";
$ADODB_FETCH_MODE = ADODB_FETCH_ASSOC;
$saveFetchModes = [$ADODB_FETCH_MODE, $db->fetchMode ];		];
print_r($saveFetchModes);
print_r($db->getAll('select * from a'));
print "\n --------------set GLOBAL TO BOTH ----------\n";
$ADODB_FETCH_MODE = ADODB_FETCH_BOTH;
$saveFetchModes = [$ADODB_FETCH_MODE, $db->fetchMode ];
print_r($saveFetchModes);
print_r($db->getAll('select * from a'));
print "\n --------------set fetchMode To Num ----------\n";
$db->setFetchMode(ADODB_FETCH_NUM);
$saveFetchModes = [ $ADODB_FETCH_MODE, $db->fetchMode ];
print_r($saveFetchModes);
print_r($db->getAll('select * from a'));
print "\n --------------set GLOBAL TO BOTH ----------\n";
$ADODB_FETCH_MODE = ADODB_FETCH_BOTH;
$saveFetchModes = [$ADODB_FETCH_MODE, $db->fetchMode ];
print_r($saveFetchModes);
print_r($db->getAll('select * from a'));
print "\n --------------set fetchMode TO BOTH ----------\n";
$db->SetFetchMode(ADODB_FETCH_BOTH);
$saveFetchModes = [$ADODB_FETCH_MODE, $db->fetchMode];
print_r($saveFetchModes);
print_r($db->getAll('select * from a'));
````

````

 --------------default ----------
Array
(
    [0] => 0
    [1] => 
)
Array
(
    [0] => Array
        (
            [0] => 1
            [id] => 1
            [1] => test
            [c] => test
        )

)

 --------------set global to ASSOC ----------
Array
(
    [0] => 2
    [1] => 
)
Array
(
    [0] => Array
        (
            [id] => 1
            [c] => test
        )

)

 --------------set GLOBAL TO BOTH ----------
Array
(
    [0] => 3
    [1] => 
)
Array
(
    [0] => Array
        (
            [0] => 1
            [id] => 1
            [1] => test
            [c] => test
        )

)

 --------------set fetchMode To Num ----------
Array
(
    [0] => 3
    [1] => 1
)
Array
(
    [0] => Array
        (
            [0] => 1
            [1] => test
        )

)

 --------------set GLOBAL TO BOTH ----------
Array
(
    [0] => 3
    [1] => 1
)
Array
(
    [0] => Array
        (
            [0] => 1
            [1] => test
        )

)

 --------------set fetchMode TO BOTH ----------
Array
(
    [0] => 3
    [1] => 3
)
Array
(
    [0] => Array
        (
            [0] => 1
            [id] => 1
            [1] => test
            [c] => test
        )

)
````

dregad:
Many thanks for taking the time to explain. I think I get it now.

Even though this discussion was triggered by a change in this PR, it is somewhat off-topic so I have opened #1189 to track the issue, so we can follow-up there and merge this.

drivers/adodb-mysqli.inc.php
dregad: Follow-up in #1189

PR #1193, opened 2026-02-19
Create driver-specific metaTables method to add support for masking, make table and view filtering part of initial statement instead of post-processing returned records

mnewnham:


drivers/adodb-mssqlnative.inc.php
```diff
@@ -960,20 +960,75 @@ function MetaPrimaryKeys($table, $owner=false)
...
+				'TABLE_PRIVILEGES','COLUMNS','COLUMN_DOMAIN_USAGE',
+				'COLUMN_PRIVILEGES','DOMAINS',
+				'DOMAIN_CONSTRAINTS','KEY_COLUMN_USAGE',
+				'DTPROPERTIES')
+				%s";
+				
+		
+		if (!$ttype) {
+			$tSql = "sysobjects.type IN('U','V')";
+		} else if (substr(strtoupper($ttype), 0, 1) == 'T') {
```
dregad: It would be more efficient to first get the substring, and only then call strtoupper.
And since we're only interested in the first character, the substr() call is not really needed, better to just do `strtoupper($ttype[0]);`

adodb.inc.php
```diff
@@ -3300,6 +3300,7 @@ function (array $table) use ($ttype): bool {
 
 
 	function _findschema(&$table,&$schema) {
+		print "FS $table $schema\n";
```
dregad: Leftover debug statement ?

dregad (changes requested):

PR #1197, opened 2026-02-20
Does not return correctly if multiple constraints exist on same table

mnewnham:
Does not reset correct fetch mode if $db->setFetchMode mode not yet set Does not return correct value if invalid table name passed to method


dregad:
Follow-up on #1175

PR #1199, opened 2026-02-20
Reverse behavior of $primary flag

### LINKED ISSUE #1198: [oci8] MetaIndexes() $primary flag produces incorrect behavior with primary keys
The [MetaIndexes](https://adodb.org/dokuwiki/doku.php?id=v5:dictionary:metaindexes) indicates that primary keys flag causes the primary key to be added the list of keys returned. The oci8 has a reversed behavior in that it only returns the primary key if the flag is selected.

mnewnham:


dregad (changes requested):
OK with the change, just a couple minor comments.

See my comment in linked issue https://github.com/ADOdb/ADOdb/issues/1198#issuecomment-3936057523

dregad:
As a side note for optimization, if I remember correctly from my Oracle DBA days, using parameterized queries instead of injecting the table name in the SQL with sprintf() would reduce the DB server's load as it will not need to rebuild the explain plan with each query execution.

dregad:
I do not see any new commits since my earlier review, did you forget to push your changes ?

dregad:
Sorry I did not mean to force-rebase this one, I had multiple browser tabs open and thought I was on PR 1193.

drivers/adodb-oci8.inc.php
```diff
@@ -495,68 +495,79 @@ function MetaTables($ttype=false,$showSchema=false,$mask=false)
 		return $ret;
 	}
 
-	// Mark Newnham
+	/**
+      * Return a list of indexes for a specified table
+      *
+      * We don't use db2_statistics as the function does not seem to play
+      * well with mixed case table names
```
dregad, line 502: *db2_statistics*, is this a copy/paste error ?

drivers/adodb-oci8.inc.php
```diff
@@ -495,68 +495,79 @@ function MetaTables($ttype=false,$showSchema=false,$mask=false)
...
+			if (!$primary && $row[0] == $primary_key) {
 				continue;
 			}
 			if (!isset($indexes[$row[0]])) {
 				$indexes[$row[0]] = array(
-					'unique' => ($row[1] == 'UNIQUE'),
-					'columns' => array()
+					'unique' => ($row[1] == 'UNIQUE') ? 1 : 0,
+					'columns' => [],
+					'primary' => ($primary_key == $row[0] ? 1 : 0)
```
dregad, line 570: It would be more efficient for the unique and primary assignments to simply perform an integer cast to transform true/false to 1/0 here instead of using ternary operator, i.e.

```php
'unique' => (int)($row[1] == 'UNIQUE'),
'columns' => [],
'primary' => (int)($primary_key == $row[0]),
```

dregad (changes requested):

mnewnham:
Transferred to #1228

drivers/adodb-oci8.inc.php
mnewnham: If you want to change it, that's fine. I've always considered casting (especially with (float) to be a danger zone for junior programmers, especially in financial applications so I effectively banned its use for the last 15 years. As PHP gets stricter and stricter with this this kind of thing I still feel it's better to be explicit with the end result.

dregad: I would agree in general, but in this specific case we are 100% sure that the expression being cast is a boolean, so there is no risk.

PR #1214, opened 2026-03-05
Fixes #1213 Incorrect return value on success from AutoExecute()

### LINKED ISSUE #1213: [core] Standardize return value format of AutoExecute to other ADOdb Insert/Updates
All ADOdb INSERT/UPDATE functions return the following values

- Success returns an `ADORecordSet_empty` object
- Failure returns false

Excepting `AutoExecute()` which retuns true/false.

This enhancement standardizes the return value. This does not change the normal evaluation of return value because `ADORecordSet_empty == true`

mnewnham:


dregad (approved):
I updated the documentation to reflect this change
https://adodb.org/dokuwiki/doku.php?id=v5:reference:connection:autoexecute

PR #1237, opened 2026-04-05
Fixes #1236 Invalid use of driver attributes instead of metaPrimaryKeys

### LINKED ISSUE #1236: [activerecord] activerecord::updateActiveTable should not rely on driver specific attributes to locate primary keys
If an `ADOdb_Active_Record` class is instantiated without passing an array of columns that represents the primary key of the table, the `updateActiveTable()` method attempts to determine the primary key of the table by looking at the attributes returned by `metaColumns()`.

The attribute in question, **_primary_key_** is however not one of the standard required attributes returned by `metaColumns()` and is only present in the `mysqli` and `postgres` drivers. All other drivers therefore fail when instantiated without a primary key list.

The `updateActiveTable()` method should instead use the `metaPrimaryKeys()` method to obtain the primary key list.

mnewnham:


dregad (approved):

ziegenberg:
Any chance of getting this PR (and #1238) merged soon so we can get a [v5.22.12](https://github.com/ADOdb/ADOdb/milestone/44) release?

PR #1238, opened 2026-04-07
fixes #1233 | Restored previous return value semantics of AutoExecute

### LINKED ISSUE #1233: Possible breaking change regarding return value of AutoExecute in 5.2.11
### Description

The changes to the return semantics of `AutoExecute` in version `5.2.11` are a breaking change and should probably be rolled back.

They were introduced in #1148. With the motivation stated as:

>In UPDATE mode, if the specified where clause matches no rows, autoUpdate() still generates and executes the SQL UPDATE statement, then returns **true** even though no rows were updated.
>
> There is no point building and executing an UPDATE statement if we know it won't do anything, so we should abort early and return **false**.

I would argue, that this changed the semantics of the return value.
Previously the return value `false` was reserved for _technical_ failures or _invalid input_:
* table not existing
* `UPDATE` mode without `WHERE` statement
* empty `fields` list

As you mentioned, a successful `UPDATE` from a technical point of view (_SQL query was built and successfully executed against the DB_) returned `true`.

Now, the same call returns `false`, but this tries to communicate a semantic issue (_the update would not update anything_).

I would propose, to change the semantics of the return value back, while _also_ respecting the original motivation:

> There is no point building and executing an UPDATE statement if we know it won't do anything, so we should abort early

Something like so:

```php
		if (!$rs) {
			// Table does not exist
			return false;
		} elseif ($mode == DB_AUTOQUERY_UPDATE && $rs->EOF) {
			//  udpate where clause matches no rows, we can abort early
			return true;
		}
```

What do you think?

### Environment
- ADOdb version: `5.2.11`
- Driver or Module: `mysqli` (irrelevant for this report)
- Database type and version: MariaDB 10.6 (irrelevant for this report)
- PHP version: `8.3` (irrelevant for this report)
- Platform: linux (irrelevant for this report)

* [x] I have tested that the problem is reproducible in the [latest release](https://github.com/ADOdb/ADOdb/releases/latest)

steffendietz:
retaining the early return when no rows would be updated.

See #1233 for discussions.

fixes #1233

### DIFF
--- a/adodb.inc.php
+++ b/adodb.inc.php
@@ -2685,9 +2685,12 @@ if (!defined('_ADODB_LAYER')) {
 		}
 
 		$rs = $this->SelectLimit($sql, 1);
-		if (!$rs || $mode == DB_AUTOQUERY_UPDATE && $rs->EOF) {
-			// Table does not exist or udpate where clause matches no rows
+		if (!$rs) {
+			// Table does not exist
 			return false;
+		} elseif ($mode == DB_AUTOQUERY_UPDATE && $rs->EOF) {
+			// where clause for update matches no rows, we can abort early
+			return true;
 		}
 
 		$rs->tableName = $table;


dregad (changes requested):
With `return true;`, this PR should be targeted at *hotfix/5.22* branch.

If you want to target _master_, then you must return `ADORecordSet_empty` as discussed in #1233, to be consistent with the code base.

steffendietz:
Okay. I rebased my PR onto the `hotfix/5.22` branch and updated the target. Thanks for bringing it up.

dregad (approved):
Thanks, looking good now. I'll merge it later.

steffendietz:
Hi. A gentle reminder, that this issue is still unresolved. :sweat_smile:

ziegenberg:
Any chance of getting this PR (and #1237) merged soon so we can get a [v5.22.12](https://github.com/ADOdb/ADOdb/milestone/44) release?

